PR #4, opened 2017-07-27
Refactors sonobuoy into build

chuckha:
Sonobuoy gets refactorted into the build directory.

The top level Makefile now defers build commands to the Makefiles found in the build/ directory.


kensimon:
run tests

ncdc:
What's the motivation for this?

chuckha:
@ncdc The motivation behind this is to modify the sonobuoy image and have the image run a script exactly like the conformance test and systemd logs plugin do. It would take the results of a sonobuoy run and put a reference to the results directory in the `/done` file. In this way, we could have another container running in the same pod as the master that watches for the `/done` directory and say, uploads results somewhere more permanent.

ncdc:
What about having this in a `build` directory enables that / what about the current structure makes that difficult or impossible?

chuckha:
@ncdc this felt more organized to me as I was planning on adding some more files (like a runner script). It would be nice to keep them all in one directory and not in the top level (or have the build be split across the top level and a subdirectory). Happy to hear alternatives though.

.gitignore
```diff
@@ -1,10 +1,39 @@
...
+*.o
+*.a
+*.so
+
+# Folders
+_obj
+_test
+
+# Architecture specific extensions/prefixes
+*.[568vq]
```
timothysc: ?

chuckha: Oh, I haven't the foggiest. This is legacy and can instead be merged with what @kensimon added in #1

kensimon: Needs a rebase, and I'd say strip this down to what I merged in #1 unless you know you need to gitignore each of these lines.

timothysc:
@ncdc Thoughts from @chuckha 's last comment?

ncdc:
I think that reasoning is reasonable 😄. I was mostly wondering if other projects are going to be required to use this build subdirectory structure, as simpler projects might not need it.

timothysc:
> I was mostly wondering if other projects are going to be required to use this build subdirectory structure, as simpler projects might not need it. 

I think "it depends" on the complexity of the project, and we can rationalize over time.

kensimon:
run tests

timothysc:
@chuckha please update per comments then feel free to self-merge.

PR #7, opened 2017-07-28
Always write a done file

chuckha:
The master node will now write a done file when it has finished writing the results.
The contents of the done file will be the name of the results file that sonobuoy has written.

The idea here is that you can react to the results being generated using a second
container that knows to look for a `done` file in some directory.

An example of this would be uploading the results to s3 after sonobuoy master finishes.

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


pkg/discovery/discovery.go
```diff
@@ -80,6 +81,17 @@ func Run(kubeClient kubernetes.Interface, cfg *config.Config) []error {
 		errlst = append(errlst, err)
 	}
 
+	// 8. Always write the done file
+	doneFile := cfg.ResultsDir + "/done"
```
timothysc: Hmm I'd almost prefer to have this in the script wrapper, b/c it makes the contract super light.  

What benefits are there to adding this to code vs. script wrapper?

chuckha: There are a couple of benefits I saw:

1. The Dockerfile doesn't have to change

chuckha: This resulted in the least amount of changes to the CLI. If we make the assumption we always want to write a done file (which we can choose to react to or not) then this becomes the easiest way forward.

It works well with the `--no-exit flag`. This change allows us to get both the blocking behavior as well as a `done` file. It's not trivial to get this behavior by only adding a script. If sonobuoy never exits, we'd need a second process watching sonobuoy waiting for it to get to the blocking state and *then* write a done file.

timothysc: > This change allows us to get both the blocking behavior as well as a done file. 

Under what conditions do we want blocking + done?  
done indicates hand-off to another process in the pod.  

I'm not opposed to the logic of adding to the process, just wanted to make sure we vetted.  But I do think that done && --no-exit are an mutually exclusive.  If you specify done you are stating that you don't want to block, but instead you want to exit.  

Perhaps we should call it something other then RESULTS_DIR, to make it more explicit.

chuckha: The only reason for `--no-exit` really is manual inspection of the container or manual copying of the results. I think you're right, this would feel much lighter coming from a script rather than the sonobuoy code. Let me revisit and see what I can do with that assumption.

PR #8, opened 2017-07-29
Fix the build

chuckha:
When we build sonobuoy with the docker container we need the output binary to
end up in the Dockerfile context for sonobuoy. In otherwords, we need to put
the built sonobuoy binary in the build/sonobuoy directory.


timothysc:
@chuckha sign-off checker needs the PR to be committed with ```--signoff```

Issue #5, opened 2017-07-27
Auto generate examples with ksonnet.


timothysc:
Currently we have plugins.d + examples and the set of potential examples could grow quite large.  I would like to leverage ksonnet here to template our examples to generate a submission.

PR #9, opened 2017-07-30
Use ksonnet to generate examples

chuckha:
Closes issue #5.

- Adds jsonnet to generate the quickstart examples
- Removes the YAML
- Adds the generated JSON
- Docs are updated to reflect JSON files vs YAML
- Adds a Makefile target, `generate-examples`


chuckha:
Here is the process I took to ensure the generated json is equivalent to yaml:

1. Generate json with new make command
2. Convert generated json to yaml (https://github.com/redsymbol/json2yaml)
3. YAML diff (https://www.npmjs.com/package/koimo)

```
cha$ node_modules/.bin/koimo gen_00-rbac.yaml 00-rbac.yaml 
  
  ===================
   Koimo - YAML Diff 
  ===================
  
  [ gen_00-rbac.yaml ] vs. [ 00-rbac.yaml ]

  ✓ All good!

cha$ node_modules/.bin/koimo gen_10-configmaps.yaml 10-configmaps.yaml 
  
  ===================
   Koimo - YAML Diff 
  ===================
  
  [ gen_10-configmaps.yaml ] vs. [ 10-configmaps.yaml ]

  ✓ All good!

cha$ node_modules/.bin/koimo gen_20-pod.yaml 20-pod.yaml 
  
  ===================
   Koimo - YAML Diff 
  ===================
  
  [ gen_20-pod.yaml ] vs. [ 20-pod.yaml ]

  ✓ All good!
```

chuckha:
I would love input on what to do with the YAML. I'm not convinced deleting it is the best path forward because in my experience people would prefer reading YAML. If they are looking for a way to get started, it would be nice to have the YAML available.

Makefile
```diff
@@ -33,9 +36,15 @@ push:
...
+examples/quickstart/%.json: examples/quickstart/%.jsonnet
+	jsonnet -J /usr/local/lib/jsonnet/ksonnet-lib -o $@ $<
 
 clean:
 	$(MAKE) -C build/sonobuoy clean
 	$(MAKE) -C build/kube-conformance clean
-	$(MAKE) -C build/systemd-logs clean
\ No newline at end of file
+	$(MAKE) -C build/systemd-logs clean
+	rm -f /examples/quickstart/*.json
```
kensimon: Probably don't want to `rm -r /examples`, but `./examples`?

docs/building-examples.md
```diff
@@ -0,0 +1,9 @@
+# Examples
+
+## Quickstart
+
+The quickstart example is built with [ksonnet][0].
+
+The JSON files are generated with the `make generate-examples` command. To get this to work you must have ksonnet-lib installed in `/usr/local/lib/jsonnet/ksonnet-lib`. After that, `make generate-examples` will build the JSON files that are checked into this repository.
+
+0: https://ksonnet.heptio.com
```
kensimon: HTTPS doesn't work on this site, link needs fixing

examples/quickstart/00-rbac.json
```diff
@@ -0,0 +1,70 @@
...
+         "metadata": {
+            "labels": {
+               "component": "sonobuoy"
+            },
+            "name": "sonobuoy-serviceaccount",
+            "namespace": "heptio-sonobuoy"
+         }
+      },
+      {
+         "apiVersion": "rbac.authorization.k8s.io/v1beta1",
```
timothysc: Can we somehow option wrap rbac with a default of true but allow for disabled generation per other issue filed?

examples/quickstart/10-configmaps.json
```diff
@@ -0,0 +1,35 @@
+{
+   "apiVersion": "v1",
+   "items": [
+      {
+         "apiVersion": "v1",
+         "data": {
+            "config.json": "{\"Description\": \"EXAMPLE\", \"Filters\": {\"LabelSelector\": \"\", \"Namespaces\": \".*\"}, \"PluginNamespace\": \"heptio-sonobuoy\", \"Plugins\": [{\"name\": \"systemd_logs\"}, {\"name\": \"e2e\"}], \"Resources\": [\"CertificateSigningRequests\", \"ClusterRoleBindings\", \"ClusterRoles\", \"ComponentStatuses\", \"Nodes\", \"PersistentVolumes\", \"PodSecurityPolicies\", \"ServerVersion\", \"StorageClasses\", \"ThirdPartyResources\", \"ConfigMaps\", \"DaemonSets\", \"Deployments\", \"Endpoints\", \"Events\", \"HorizontalPodAutoscalers\", \"Ingresses\", \"Jobs\", \"LimitRanges\", \"PersistentVolumeClaims\", \"Pods\", \"PodLogs\", \"PodDisruptionBudgets\", \"PodPresets\", \"PodTemplates\", \"ReplicaSets\", \"ReplicationControllers\", \"ResourceQuotas\", \"RoleBindings\", \"Roles\", \"ServiceAccounts\", \"Services\", \"StatefulSets\"], \"ResultsDir\": \"/tmp/sonobuoy\", \"Server\": {\"advertiseaddress\": \"sonobuoy-master:8080\", \"bindaddress\": \"0.0.0.0\", \"bindport\": 8080, \"timeoutseconds\": 3600}, \"Version\": \"v0.8.0\"}"
```
timothysc: This almost makes it unreadable...

hausdorff: Couple of notes here.

* Jsonnet supports multi-line strings (denoted with, `|||`, [_c.f._][ml])
* [`kubecfg`][kf] exposes native extensions [parseYaml][py] and [manifestYaml][my] allows you to marshal back and forth from YAML.

[ml]: http://jsonnet.org/docs/tutorial.html#arith_cond
[kf]: https://github.com/ksonnet/kubecfg
[py]: https://github.com/ksonnet/kubecfg/blob/master/lib/kubecfg.libsonnet#L31
[my]: https://github.com/ksonnet/kubecfg/blob/master/lib/kubecfg.libsonnet#L43

examples/quickstart/00-rbac.json
hausdorff: It's not easy right now, but it's certainly on the roadmap for the CLI.

One way to do it right now is to use [ext vars][ev], which are passed in via command line. Your build process would involve interrogating the API server, getting capabilities, and passing a flag `--ext-str` with a value to indicate as such.

[ev]: http://jsonnet.org/docs/stdlib.html#ext_vars

examples/quickstart/00-rbac.jsonnet
```diff
@@ -0,0 +1,71 @@
...
+
+    # shared metadata
+    metadata: {
+        name: $.serviceAccount.name,
+        labels: $.labels,
+    },
+
+};
+
+local namespace = local ns = k.core.v1.namespace;
```
hausdorff: Super tiny micro nit, when `jsonnet fmt` matures, this local will get indented below. AFAIK that's actively being worked on by Dave's intern.

examples/quickstart/10-configmaps.json
hausdorff: I would use [`kubecfg`][kf] here, and generate YAML instead of JSON. `kubecfg` natively emits multi-line strings all prettified.

[kf]: https://github.com/ksonnet/kubecfg

examples/quickstart/10-configmaps.jsonnet
```diff
@@ -0,0 +1,233 @@
...
+    },
+    PluginNamespace: "heptio-sonobuoy",
+    Plugins: [
+          {name: "systemd_logs"},
+          {name: "e2e"}
+    ]
+};
+
+local plugins = {
+  "systemdlogs.yaml": 'name: systemd_logs
```
hausdorff: Hmm, it seems to me you could write this as a normal JSON object and then call `std.toString(systemdLogs)` on it to get a string? Am I missing something crucial?

examples/quickstart/10-configmaps.jsonnet
```diff
@@ -0,0 +1,233 @@
...
+    configMap.mixin.metadata.namespace(conf.namespace) +
+    configMap.mixin.metadata.labels(conf.labels) +
+    configMap.data(
+        plugins
+    );
+
+k.core.v1.list.new([
+    sonobuoyConfig,
+    pluginConfigs,
+])
```
hausdorff: Tiny, tiny microest of nits, you might consider putting newlines at the end so that diffs are cleaner.

examples/quickstart/10-configmaps.json
hausdorff: The top comment was supposed to be deleted. Hmm. Ignore it. I didn't know what I was looking at yet

timothysc:
So I'm inclined to hold off on this unless it is strictly cleaner and more concise, that is the bar we should be holding here.

hausdorff:
@timothysc Your comment here makes me wonder if there is anything beyond the comments listed you'd like to see changed? Changing to YAML will make it cleaner and more concise, but it's not a change to the Jsonnet code itself, which is where I'd expect the bulk of cleanliness changes to go.

timothysc:
@hausdorff Right now I think the biggest detractor is the readability.  I'd also like to get to the point where we can generate the plugin definitions, and perhaps use that as a mixin to the quickstart .yaml.

hausdorff:
@timothysc Strongly agree about the plugin definitions, which I think are doable with `std.toString`, I think I mentioned this before.

I think my main question, though, is that it's not clear (to me, at least) what else needs to change to meet the readability bar. Just the string pretty printing?

timothysc:
> I think my main question, though, is that it's not clear (to me, at least) what else needs to change to meet the readability bar. Just the string pretty printing? 

Eliminate the nested structure string escaping.

examples/quickstart/10-configmaps.jsonnet
chuckha: [Unfortunately the plugin loader only knows how to read YAML.](https://github.com/heptio/sonobuoy/blob/master/pkg/plugin/loader/loader.go#L133)

I'll make an issue and this issue will depend on that issue.

hausdorff: @chuckha Is YAML not a superset of JSON, though? So it seems like we should be able to use JSON here and be ok.

If not, `kubecfg` allows you to import YAML files, as well `manifestYAML`, which pretty-prints an object as YAML.

Or perhaps I am missing something?

chuckha:
I believe all issues have been addressed in #52

PR #10, opened 2017-07-31
Customizable resultsDir

chuckha:
Allow the environment variable RESULTS_DIR to take precedence over the configuration file.

This is an infrastructure change to move along the `done` file work.


kensimon:
Build is failing (`make all clean` ends up running `docker rmi` on an image that doesn't exist) but jenkins wasn't able to call into the github API for some reason (possibly due to the github outage today.)

PR #20, opened 2017-08-06
Implement the done file interface

chuckha:
Once Sonobuoy is finished running it writes the path of the results tarball
into a done file located in the results directory (as configured via
the config file or an env var).

The goal of this is to be able to react to a finished sonobuoy run and
do something interesting with the results. For instance, one could
upload the results to S3 for permanent storage and need not create a long
lived volume for sonobuoy specifically.

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -59,7 +59,7 @@ func LoadConfig() (*Config, error) {
 
 	// 3 - figure out what address we will tell pods to dial for aggregation
 	if cfg.Aggregation.AdvertiseAddress == "" {
-		if ip, ok := os.LookupEnv("SONOBUOY_ADVERTISE_IP"); ok {
+		if ip := os.Getenv("SONOBUOY_ADVERTISE_IP"); ip != "" {
 			cfg.Aggregation.AdvertiseAddress = fmt.Sprintf("%v:%d", ip, cfg.Aggregation.BindPort)
 		} else {
 			hostname, _ := os.Hostname()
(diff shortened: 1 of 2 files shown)


pkg/config/loader.go
```diff
@@ -59,7 +59,7 @@ func LoadConfig() (*Config, error) {
 
 	// 3 - figure out what address we will tell pods to dial for aggregation
 	if cfg.Aggregation.AdvertiseAddress == "" {
-		if ip, ok := os.LookupEnv("SONOBUOY_ADVERTISE_IP"); ok {
+		if ip := os.Getenv("SONOBUOY_ADVERTISE_IP"); ip != "" {
```
timothysc: This change seems orthogonal, what's the motivation?

chuckha: This is because the `SONOBUOY_ADVERTISE_IP` env (as well as all expected environment variables) will always be passed from the [`run_master.sh`](https://github.com/heptio/sonobuoy/pull/20/files#diff-9160fb798e31c8d887cdb95a8f0051a1R19) script into the `sonobuoy` binary. Therefore `LookupEnv` will always return `true` for `ok` since we are always setting it, even if it's empty.

If there is a case of setting `SONOBUOY_ADVERTISE_IP` to `''` where we don't want the `else` block to execute then I'll have to rethink this strategy.

timothysc: Did you verify all was well on a clustered run?

chuckha: Yes, but I wouldn't mind going over how I verified to ensure I didn't miss anything.

timothysc:
@luxas with this mechanism you could now put whatever tosser or web-service in front of sonobuoy that you wanted.  We strive to make it composable so that service is separate from sonobuoy itself.  

Fixes: https://github.com/heptio/sonobuoy/issues/24

PR #41, opened 2017-08-08
Don't mount /tmp as host volume in example pod

kensimon:
We're encouraging using `kubectl cp` in the meantime, using host mounts
was a temporary measure.

This fixes an issue where the documented `kubectl cp` command is
bringing in tarballs from previous runs.

### DIFF
--- a/examples/quickstart/20-pod.yaml
+++ b/examples/quickstart/20-pod.yaml
@@ -53,11 +53,11 @@ spec:
     - name: sonobuoy-plugins-volume
       configMap:
         name: sonobuoy-plugins-cm
-    # NOTE: This will drop the results on the hostpath of the machine as an example.
-    # We recommend using a persistent volume for more permanent recording.
+    # NOTE: This will drop the results in a temporary directory that only lasts
+    # while the pod is still running. We recommend using a persistent volume
+    # for more permanent recording.
     - name: output-volume
-      hostPath:
-        path: /tmp/sonobuoy
+      emptyDir: {}
 
 ---
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


examples/quickstart/20-pod.yaml
```diff
@@ -55,9 +53,6 @@ spec:
         name: sonobuoy-plugins-cm
     # NOTE: This will drop the results on the hostpath of the machine as an example.
     # We recommend using a persistent volume for more permanent recording.
-    - name: output-volume
-      hostPath:
```
chuckha: Would we want this to be an emptyDir instead of deleting it?

timothysc: During some incantation it was.

kensimon: Sure, that's probably a decent idea, even though it's roughly equivalent to no volume at all, since /tmp is writable inside the container.

With emptyDir it's easier to change the YAML to a different type of persistent volume for your purposes, without having to remember what path to use, etc... I'll change it to emptyDir.

abiogenesis-now: @kensimon it's a remnant from before you and Tim added `no-exit`. Previously, when the pod finished it disappeared I think (so the `emptyDir` would be removed with it?), so I changed the example to `hostPath` so that I could SSH in afterwards and poke at it.

TLDR should be fine to change to emptyDir. If you do so, could you also take out one of the sub-bullet points in https://github.com/heptio/sonobuoy/blob/master/docs/configuration.md#pod (the part under "Output Volume Mount")?

kensimon: Changed it to `emptyDir`, @abiogenesis-now please take a look at docs/configuration.md and let me know if you have any changes.

abiogenesis-now:
oh I don't have permissions to stamp 🤷‍♀️

abiogenesis-now:
oh wait nvm, there just isn't a bot to recognize /lgtm I guess

timothysc:
b/c we don't have turnkey validation in place I'll say ok-to-self merge on test verification.

PR #44, opened 2017-08-09
[docs] README clarifications for error handling and sonobuoy no-exit

abiogenesis-now:
@kensimon @timothysc PTAL This is meant to address the confusion raised in https://github.com/heptio/sonobuoy/issues/42.

Please see the [formatted markdown](https://github.com/abiogenesis-now/sonobuoy/blob/20870e11b922b0f6d07beb55f1c7b825d4303a93/README.md#2-run)! I'm not 100% sure that the quote block works----I wanted to visually highlight some important points and also show that they weren't "part of the steps". Let me know your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,25 @@ To verify that Sonobuoy has completed successfully, check the logs:
 ```
 kubectl logs -f sonobuoy --namespace=heptio-sonobuoy
 ```
-If you see the log line `no-exit was specified, sonobuoy is now blocking`, the Sonobuoy pod is done running.
+If you see the log line `no-exit was specified, sonobuoy is now blocking`, the Sonobuoy pod has completed its data collection.
 
-*Note*: If you see the error `plugin <name> does not exist`, delete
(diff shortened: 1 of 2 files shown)


kensimon:
@abiogenesis-now could you amend with a signed-off-by?

abiogenesis-now:
ah right, forgot..should prob just include it in my commit alias

README.md
```diff
@@ -56,10 +56,23 @@ To verify that Sonobuoy has completed successfully, check the logs:
...
 kubectl logs -f sonobuoy --namespace=heptio-sonobuoy
 ```
-If you see the log line `no-exit was specified, sonobuoy is now blocking`, the Sonobuoy pod is done running.
+If you see the log line `no-exit was specified, sonobuoy is now blocking`, the Sonobuoy pod has completed its data collection.
 
-*Note*: If you see the error `plugin <name> does not exist`, delete
-the `sonobuoy` pod via `kubectl delete pod sonobuoy --namespace heptio-sonobuoy` and rerun the `kubectl apply` command above. This error occurs when Sonobuoy's ConfigMap creations are not completed before its pod creation. While the YAML files in `examples/quickstart` use numeric prefixes ("00-", "10-", etc.) to specify the order of resource creation, `kubectl apply` is asynchronous and can result in dependency issues.
+> *Notes*:
+>
+> * **The Sonobuoy pod in this example continues to run after it finishes data collection**, due to the `--no-exit` flag in its YAML manifest. This allows you to easily grab the results tarball.
```
timothysc: I'd like to add the preferred to use persistent volume clause here if we're making a separate note about no-exit.

abiogenesis-now:
@timothysc PTAL (1) Added an additional note for persistent volumes (2) Added links in the plugin doc to the source code repos.

Issue #28, opened 2017-08-07
Why is ./plugins/e2e/results.tar.gz tarred twice?


luxas:
I'd expect sonobuoy to produce the tar file for me to untar once and not have multiple layers of tarred files.
I guess this is just the output from the plugin, but could sonobuoy handle untarring possible .tar.gz files produced by plugins (like e2e)?

timothysc:
So the plugins allow for one file to be transferred and it was simple.  I suppose we could modify aggregation to explode transferred file.

kensimon:
This is fixed in v0.8.2

PR #47, opened 2017-08-10
Implicitly extract tar.gz files from plugins that upload them

kensimon:
This makes it so plugins upload to a path like:

```
/api/v1/results/global/e2e.tar.gz
```

And it will show up in:

```
/results/plugins/e2e/results/<contents>
```

(For single-node plugins, this would map to:)

```
/results/plugins/systemd_logs/node1/<contents>
```

This needs to be documented along with the snapshot documentation that
is already under review.

Since the upload path is now expected to have a file extension in it, In
case for some reason the sonobuoy master launches with an old version
of the sonobuoy worker (or for plugins not using `sonobuoy worker` that
upload their own contents), if the path has no file extension, it is
guessed to be `.json` or `.tar.gz`, same as before.

### DIFF
--- a/pkg/plugin/aggregation/aggregator.go
+++ b/pkg/plugin/aggregation/aggregator.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/heptio/sonobuoy/pkg/plugin"
+	"github.com/viniciuschiele/tarx"
 )
 
 // Aggregator is responsible for taking results from an HTTP server (configured
@@ -196,26 +197,55 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
 	// Create the output directory for the result.  Will be of the
 	// form .../plugins/:results_type/:node.json (for DaemonSet plugins) or
 	// .../plugins/:results_type.json (for Job plugins)
-	resultsFile := path.Join(a.OutputDir, result.Path()+result.Extension())
+	resultsFile := path.Join(a.OutputDir, result.Path()+result.Extension)
 	resultsDir := path.Dir(resultsFile)
 	glog.Infof("Creating directory %v", resultsDir)
 	if err := os.MkdirAll(resultsDir, 0755); err != nil {
 		glog.Errorf("Could not make directory %v: %v", resultsDir, err)
-
 		return err
 	}
 
-	// Open the results file for writing
-	f, err := os.Create(resultsFile)
+	// Write the results file out and close it
+	err := func() error {
+		f, err := os.Create(resultsFile)
+		if err != nil {
+			glog.Errorf("Could not open output file %v for writing: %v", resultsFile, err)
+			return err
+		}
+		defer f.Close()
+
+		// Copy the request body into the file
+		_, err = io.Copy(f, result.Body)
+		if err != nil {
+			glog.Errorf("Error writing plugin result: %v", err)
+			return err
+		}
+
+		return nil
+	}()
 	if err != nil {
-		glog.Errorf("Could not open output file %v for writing: %v", resultsFile, err)
 		return err
 	}
-	defer f.Close()
 
-	// Copy the request body into the file
-	io.Copy(f, result.Body)
-	glog.Infof("wrote results to %v\n", resultsFile)
+	// If it's a tarball, extract it
+	if result.Extension == ".tar.gz" {
+		resultsDir := path.Join(a.OutputDir, result.Path())
+
+		err = tarx.Extract(resultsFile, resultsDir, &tarx.ExtractOptions{})
+		if err != nil {
+			glog.Errorf("Could not extract tar file %v: %v", resultsFile, err)
+			return err
+		}
+
+		err = os.Remove(resultsFile)
+		if err != nil {
+			return err
+		}
+
+		glog.Infof("extracted results tarball into %v", resultsDir)
+	} else {
+		glog.Infof("wrote results to %v", resultsFile)
+	}
 
 	return nil
 }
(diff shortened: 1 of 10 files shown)


kensimon:
Fixes #28

pkg/plugin/aggregation/aggregator.go
```diff
@@ -193,15 +194,22 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
 		a.resultEvents <- result
 	}()
 
+	// Get the file extension to save the result as, falling back on guessing it.
+	var extension string
+	if result.Extension != "" {
+		extension = result.Extension
+	} else {
+		extension = result.GuessExtension()
```
timothysc: Guess?  Aren't we always going to be explicit w/defaults?

pkg/plugin/aggregation/aggregator.go
```diff
@@ -217,5 +225,20 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
 	io.Copy(f, result.Body)
 	glog.Infof("wrote results to %v\n", resultsFile)
 
+	// If it's a tarball, extract it
+	if extension == ".tar.gz" {
+		f.Close()
```
timothysc: shouldn't we close above and remove the defer?

pkg/plugin/aggregation/aggregator.go
```diff
@@ -217,5 +225,20 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
...
+		err = tarx.Extract(resultsFile, path.Join(a.OutputDir, result.Path()), &tarx.ExtractOptions{})
+		if err != nil {
+			glog.Errorf("Could not extract tar file %v: %v", resultsFile, err)
+			return err
```
timothysc: I'd have to pop up, but what happens if there is an error in the aggregation?  Is it just reported and tar.gz the whole results?

pkg/plugin/aggregation/aggregator_test.go
```diff
@@ -44,7 +47,59 @@ func TestAggregation(t *testing.T) {
 				t.Errorf("results for node1 incorrect (got %v): %v", string(bytes), err)
 			}
 		} else {
-			t.Errorf("AggregationServer didn't record a result for node1")
+			t.Errorf("AggregationServer didn't record a result for node1. Got: %+v", agg.Results)
+		}
+	})
+}
+
+func TestAggregation_guessExtension(t *testing.T) {
```
timothysc: same comment as above.

pkg/plugin/aggregation/server.go
```diff
@@ -212,3 +215,16 @@ func (s *Server) globalResultsHandler(w http.ResponseWriter, r *http.Request) {
 	s.ResultsCallback(result, w)
 	r.Body.Close()
 }
+
+// given an uploaded filename, parse it into its base name and extension.  If
```
timothysc: Could use https://golang.org/pkg/path/filepath/

pkg/plugin/aggregation/aggregator.go
kensimon: Trying to be defensive in case later code introduces things that may panic... Close() is safe to call repeatedly on a filehandle in go, at least from what I've been able to dig up.

I'll make another pass at this and see if I can make it a bit cleaner, I'm not all that happy with it yet.

pkg/plugin/aggregation/aggregator.go
kensimon: We only know the extension if one was provided in the upload URI (ie. /api/v1/results/global/foo.json), I'm not introducing any concept of per-plugin defaults or anything.

In case a plugin doesn't include a `.` in the upload URI, we're using the old behavior of guessing `json` unless it's the e2e plugin.  This may be a bit over-defensive though.  Any suggestions on a better approach?

pkg/plugin/aggregation/aggregator.go
kensimon: Good point, currently errors are channeled from `(*Driver).Monitor()` as `Result` objects and get stored the same way, but nothing's specifying that their extension should be json.  I'll add `Extension: ".json"` to the MakeErrorResult function [here](https://github.com/heptio/sonobuoy/blob/master/pkg/plugin/driver/utils/utils.go#L77), and make sure we have tests for this.

pkg/plugin/aggregation/server.go
kensimon: TIL.  But looks like `filepath.Ext` grabs the last dot, I want the first dot (eg. I want ".tar.gz" from "e2e.tar.gz", not ".gz".)

kensimon:
@timothysc PTAL

Made some changes per review:

- If there's no file extension in the upload URI, we just save the file without an extension instead of guessing one (so no untarring will happen in that case)
- Plugin errors channeled from driver.Monitor() now carry a ".json" extension to make things consistent, and there's a test for this now
- Removed duplicate Close() calls by putting the file write in a closure and putting the `defer f.Close()` there

pkg/plugin/aggregation/aggregator.go
```diff
@@ -196,26 +197,55 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
...
 	if err != nil {
-		glog.Errorf("Could not open output file %v for writing: %v", resultsFile, err)
 		return err
 	}
-	defer f.Close()
 
-	// Copy the request body into the file
-	io.Copy(f, result.Body)
-	glog.Infof("wrote results to %v\n", resultsFile)
+	// If it's a tarball, extract it
```
timothysc: could switch case this as we'll likely have other file options to add.

kensimon: I'm thinking we can cross that bridge when we get there... it's possible we'll want to something more intelligent like some full post-processing logic that lives somewhere else, etc.

PR #48, opened 2017-08-10
Supports JSON configs

chuckha:
Sonobuoy looks at the extensions of the files the configmap defines
and uses a loader implementation to Unmarshal the bytes into a struct.

Both YAML and JSON are supported with respective file extensions of .yaml and .json.

### DIFF
--- a/pkg/plugin/loader/loader.go
+++ b/pkg/plugin/loader/loader.go
@@ -17,11 +17,12 @@ limitations under the License.
 package loader
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


chuckha:
Motivation here is that if we use jsonnet to *nicely* define the plugin configs, we will end up with a ConfigMap that defines JSON plugin configs instead of the expected YAML.

We *COULD* wrap the YAML in a jsonnet string, but that was rejected in PR #9. This PR adds support for JSON plugin config definitions which unblocks #5.

timothysc:
@kensimon could you review please.

Issue #35, opened 2017-08-08
Refactor Error Handling


chuckha:
Original error handling was made around gathering the details and continuing on.  The problem is they are spewed at the end of the run which make it more difficult to debug.  

I no longer think we need to propagate errors back to the root, but simply indicate pass fail and log at the location of the error.  

I do not think we need to fix this for the OSS release, but can do it afterwards.

chuckha:
@kensimon did #50 close this?

timothysc:
Let's leave it open for now to eval the behavior.

kensimon:
I think we can re-open (or open new issues) if there's any issues with the new behavior... for now we've accomplished the stated goals for this issue

PR #50, opened 2017-08-14
Improve error handling with pkg/errors lib

kensimon:
Goals:

- All errors are either handled and logged, or returned (not both.)
- When we log the errors, the real source line of the actual error can be emitted with a stack trace if --debug is enabled

Whereas before, we would build up an array of errors for any time we could proceed, now we just log them and return nil.  For errors where we can't proceed, we return the error immediately.  Errors making resource queries are serialized into the tarball same as before.

Here's the strategy or how to construct/wrap/forward errors:

- Our own errors are constructed with `errors.Errorf()` to keep track of the stack trace if we need it
- Errors from non-sonobuoy code are wrapped with `errors.Wrap()` if we want to add a context message, or `errors.WithStack()` if we don't.  This guarantees that the stack trace is captured at the innermost point.
- A simple wrapper `errlog.LogError()` is used to log errors, so that errors can optionally include stack traces, depending on whether `--debug` is provided (using "%+v" versus "%v" in the format string)

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -23,6 +23,7 @@ import (
 	"github.com/heptio/sonobuoy/pkg/buildinfo"
 	"github.com/heptio/sonobuoy/pkg/plugin"
 	pluginloader "github.com/heptio/sonobuoy/pkg/plugin/loader"
+	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -49,12 +50,12 @@ func LoadConfig() (*Config, error) {
 
 	// 1 - Read in the config file.
 	if err = viper.ReadInConfig(); err != nil {
-		return nil, err
+		return nil, errors.WithStack(err)
 	}
 
 	// 2 - Unmarshal the Config struct
 	if err = viper.Unmarshal(cfg); err != nil {
-		return nil, err
+		return nil, errors.WithStack(err)
 	}
 
 	// 3 - figure out what address we will tell pods to dial for aggregation
@@ -106,7 +107,7 @@ func LoadClient(cfg *Config) (kubernetes.Interface, error) {
 		config, err = rest.InClusterConfig()
 	}
 	if err != nil {
-		return nil, err
+		return nil, errors.WithStack(err)
 	}
 
 	// 2 - creates the clientset from kubeconfig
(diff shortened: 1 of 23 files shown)


kensimon:
fixes #35

pkg/config/loader.go
```diff
@@ -49,12 +50,12 @@ func LoadConfig() (*Config, error) {
 
 	// 1 - Read in the config file.
 	if err = viper.ReadInConfig(); err != nil {
-		return nil, err
+		return nil, errors.WithStack(err)
```
timothysc: For most of the base level operations that are not too deep do we need the WithStack?  

Other areas make a lot of sense, but what's the motivation for stack traces at this level?

pkg/discovery/discovery.go
```diff
@@ -51,35 +61,33 @@ func Run(kubeClient kubernetes.Interface, cfg *config.Config) []error {
...
-	// closure used to collect and report errors.
-	rollup := func(err []error) {
-		if err != nil {
-			errlst = append(errlst, err...)
-		}
-	}
-
 	// 4. Run the plugin aggregator
-	errlst = append(errlst, pluginaggregation.Run(kubeClient, cfg.LoadedPlugins, cfg.Aggregation, outpath)...)
+	trackErrorsFor("running plugins")(
```
timothysc: Should the plugins.Run just log errors inline and not return?  This was my conundrum for *.  Reporting up to this layer didn't make as much sense as we had originally thought.

pkg/discovery/discovery.go
```diff
@@ -51,35 +61,33 @@ func Run(kubeClient kubernetes.Interface, cfg *config.Config) []error {
...
+		pluginaggregation.Run(kubeClient, cfg.LoadedPlugins, cfg.Aggregation, outpath),
+	)
 
 	// 5. Run the queries
-	rollup(QueryClusterResources(kubeClient, cfg))
+	trackErrorsFor("querying cluster resources")(
```
timothysc: same.

pkg/config/loader.go
kensimon: Actually showing the stack trace only happens if you run sonobuoy with `--debug`, otherwise you still just get the error message, annotated with the source line of wherever you chose to log it (same behavior as now.)

So to make `--debug` work in general, I'm using WithStack so that the error has the source file:line "tagged" onto it at the place where it happened.  In this case, since we called WithStack on loader.go:53, the stack trace will show loader.go:53 in it.  If we instead passed the error up untouched until it was logged in (say) main.go, the stack trace would just show main.go in it, which is less helpful.

pkg/discovery/discovery.go
kensimon: Same logic as above, the querying of resources is only returning fatal errors now... non-fatal errors are being logged and not returned.

PR #52, opened 2017-08-17
Use ksonnet to generate examples

chuckha:
Closes issue #5.

- Adds jsonnet to generate the quickstart examples
- Removes the YAML
- Adds the generated JSON
- Docs are updated to reflect JSON files vs YAML
- Adds a Makefile target, `generate-examples`

More details:

The only dependency for building examples is docker. We use the official ksonnet image.


chuckha:
To add a little more detail:

This does not support an RBAC flag. That is fairly trivial to add, but I need more time to test it or a volunteer to test the example on a non-rbac enabled cluster.

I believe this PR addresses concerns about config map readability.

timothysc:
@kensimon could you take this one.

PR #57, opened 2017-08-18
Fix snapshot docs in README

kensimon:
- There's always one tarball now that we switched to emptyDir for the
  example
- Link to the snapshot docs


abiogenesis-now:
LGTM

PR #60, opened 2017-08-18
Makefile: tag with git describe output

kensimon:


kensimon:
Self-merging, this fixes the jenkins deployer to tag intelligently:

- using an 8-digit git sha1 for every deploy to master
- using a `:master` alias for whatever's in master
- using the tag name (without the "v" prefix) iff master's HEAD exactly matches a tag
- using `:latest`, iff master's HEAD explicitly matches a tag

Jenkins is now just calling `make container push GIT_BRANCH=master`

PR #43, opened 2017-08-09
Proposal and docs for changes to snapshot layout

kensimon:
This functions as both a proposal for how the snapshot tarballs should
be structured, as well as documentation for that structure.

Places with a "NOTE:" section are places where changes are proposed.

PR #61, opened 2017-08-18
Initial stab at changing snapshot layout

kensimon:
This shifts some of the layout based on the doc changes in #43.

Also adds some docs explaining the auto-extraction of plugin tar files

### DIFF
--- a/docs/snapshot.md
+++ b/docs/snapshot.md
@@ -1,7 +1,5 @@
 # Sonobuoy Snapshot Layout
 
-> NOTE: This documentation is a draft proposal for the structure of a Sonobuoy tarball and how it should be changed.  This documentation does not represent the current state of a Sonobuoy tarball.
-
 - [Filename](#filename)
 - [Contents](#contents)
 	- [/resources](#resources)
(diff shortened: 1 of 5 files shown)


timothysc:
So this looks like it solely shuffles the meta-data into a separate location.  

I debate if we want to create a total separate abstraction for the schema in the case where we need to shuffle again.  Thoughts?

kensimon:
There were a few things that got lost in the shuffle since the previous attempt at this (dropping `results.json` from each resources directory, mainly), I'm going to try and even some things out.

I like the abstraction idea, I think we're doing this a little bit with the QueryRecorder concept, we could do something similar for resources (keep them in an in-memory data structure and write them out in a separate layer), but I think we can wait for that.

I'll push an update that renames `resources/non-ns` to `resources/cluster` and drops the no-op `resources/**/results.json` which has no data in it.

kensimon:
(Oh and the other main thing that *is* already in this PR, other than query metadata, is pod logs, which have moved to `/podlogs` in the tarball to stay separate from normal resources.)

PR #62, opened 2017-08-24
Makefile: fix docker tagging when branches have slashes

kensimon:


kensimon:
Looks like jenkins already sets a `GIT_BRANCH` environment variable, which was overriding it.  So I just renamed the env var to IMAGE_BRANCH which reflects that it's the tag for the image that represents the current branch.

abiogenesis-now:
makes sense

PR #63, opened 2017-08-24
Drop old gcloud line from make push

kensimon:


kensimon:
Self merging because this is breaking the jenkins deployer.

Issue #55, opened 2017-08-18
Include run log as part of tarball results


kensimon:
The output of sonobuoy should be included in the tarball before it is packaged, so any errors can be clearly included in the tarball.

timothysc:
+ Update to logrus

timothysc:
Fixed by #64

PR #64, opened 2017-08-26
Use logrus for logs, write them into the snapshot

kensimon:
The logs will be stored in "meta/run.log" as JSON-formatted log lines

The kubernetes client's logs (which use glog) aren't included since
there's no good way to redirect those.  We can consider changing our
docs to drop the `-v 5 --logtostderr` from the command line, since it
only shows kube-client logs now.


kensimon:
fixes #55

Issue #56, opened 2017-08-18
Gather list of supported api groups


kensimon:
Similar to what `kubectl api-versions` does, we should gather what the api versions are that the server supports.

PR #66, opened 2017-08-28
Query ServerGroups to see supported API versions

kensimon:
We're storing them in /servergroups.json for now, may move this to
/server/groups.json later (and move /serverversion.json to
/server/version.json as well)


kensimon:
fixes #56

PR #70, opened 2017-09-18
Fix link for E2E_Focus so it points to the right line

ceridwen:
This just changes the documentation so the link points to the line that actually contains E2E_FOCUS in the quickstart example.

### DIFF
--- a/docs/conformance-testing.md
+++ b/docs/conformance-testing.md
@@ -46,6 +46,6 @@ To customize the set of tests that will be run as part of the report, the follow
 [5]: plugins.md
 [6]: https://github.com/heptio/sonobuoy/blob/master/build/Dockerfile
 [7]: https://github.com/heptio/sonobuoy/blob/master/plugins.d/e2e.yaml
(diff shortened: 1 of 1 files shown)


timothysc:
It's wrong in the PR too.  https://github.com/heptio/sonobuoy/blob/master/examples/quickstart/components/10-configmaps.yaml#L81 is the correct one.  

We will likely run through all the links again once we near the 0.9.0 release in a couple of weeks.

abiogenesis-now:
thanks for catching--I have ideas on better catching file-dependent changes like this in the future (e.g. via a script or something), but for now this works.

ceridwen:
@timothysc I fixed it here, no idea how that happened.  I was using Github rather than the command line, maybe that was the problem.

timothysc:
@ceridwen could you signoff your commit(s).  

git commit --signoff

ceridwen:
I put Signed-off-by: Ceridwen <ceridwen@users.noreply.github.com> in the commit message like the instructions said?

PR #87, opened 2017-09-29
[docs] Add links to Sonobuoy Scanner

abiogenesis-now:
@timothysc PTAL

See https://github.com/abiogenesis-now/sonobuoy/blob/39285b8ec5c78a0aba07a867476f2c3f5fd2c9b9/README.md to see what it looks like.

For better visibility, I decided to put the Scanner link in its own section before "Getting Started" (rather than under "Run"). Let me know if you think this is appropriate, or if I should change it. Thanks!


abiogenesis-now:
Addresses https://github.com/heptio/sonobuoy/issues/86 (and possibly https://github.com/heptio/sonobuoy/issues/19)?

abiogenesis-now:
And also maybe addresses https://github.com/heptio/sonobuoy/issues/29, since it looks like we get the relevant system logs for failed tests?

PR #92, opened 2017-10-02
Fixes #75 - Cleanup stale file handle

### LINKED ISSUE #75: worker.GatherResults doesn't make it possible to close the results file
It calls DoRequest and returns an `io.Reader`, so it never closes the file it opens.

timothysc:
Fixes #75 

### DIFF
--- a/pkg/worker/worker.go
+++ b/pkg/worker/worker.go
@@ -36,6 +36,7 @@ import (
 func GatherResults(waitfile string, url string) error {
 	var inputFileName []byte
 	var err error
+	var outfile *os.File
 
 	// just loop looking for a file.
 	done := false
@@ -60,9 +61,16 @@ func GatherResults(waitfile string, url string) error {
 		url += "." + filenameParts[1]
 	}
 
+	defer func() {
+		if outfile != nil {
+			outfile.Close()
+		}
+	}()
+
 	// transmit back the results file.
 	return DoRequest(url, func() (io.Reader, error) {
-		outfile, err := os.Open(s)
+		outfile, err = os.Open(s)
 		return outfile, errors.WithStack(err)
 	})
+
 }


pkg/worker/worker.go
```diff
@@ -36,6 +36,7 @@ import (
 func GatherResults(waitfile string, url string) error {
 	var inputFileName []byte
 	var err error
+	var outfile * os.File
```
chuckha: gofmt this bad boy

timothysc: It does, on save.. 

    "go.formatOnSave": true, 
    "go.formatTool": "gofmt",

pkg/worker/worker.go
```diff
@@ -61,8 +62,14 @@ func GatherResults(waitfile string, url string) error {
 	}
 
 	// transmit back the results file.
-	return DoRequest(url, func() (io.Reader, error) {
-		outfile, err := os.Open(s)
-		return outfile, errors.WithStack(err)
+	err = DoRequest(url, func() (io.Reader, error) {
+		outfile, e2 := os.Open(s)
+		return outfile, errors.WithStack(e2)
```
davecheney: I'm not sure about this; outfile on line 66 is in a different scope to line 39, so on line 70, outfile will always be nil.

I suggest

s/outfile/f here to make it clear that you're talking about a different variable.

pkg/worker/worker.go
```diff
@@ -36,6 +36,7 @@ import (
 func GatherResults(waitfile string, url string) error {
 	var inputFileName []byte
 	var err error
```
davecheney: delete this line, use err := on line 65

pkg/worker/worker.go
chuckha: weird!! When I save it it moves that * over to be `var outfile *os.File`

I literally don't care on this issue. Feel free to push.

timothysc:
Thx for the quick review.

pkg/worker/worker.go
```diff
@@ -36,6 +36,7 @@ import (
...
+	var outfile *os.File
 
 	// just loop looking for a file.
 	done := false
```
davecheney: I know this is out of scope, but what about

for {
 // do work
 if err == nil {
     // success!!
     break
 }
 // log and wait
}

pkg/worker/worker.go
```diff
@@ -60,9 +61,16 @@ func GatherResults(waitfile string, url string) error {
 		url += "." + filenameParts[1]
```
davecheney: line 45: inputFileName is misleading, it's not the name of a file, it's the contents of waitfile.

pkg/worker/worker.go
timothysc: Patches welcome ;-)  
fwiw I didn't write this part of the code, I'm just trying to throw some fixes in before release.

davecheney:
Noted, I'll be the change I seek.

PR #93, opened 2017-10-02
[docs] Fix plugin example

abiogenesis-now:
PTAL @timothysc Just a quick fix.

----

**Additional Notes**:

*This is not the ideal fix*---ideally we'd be using some sort of `include` functionality so that the docs would be pulling from actual code (rather than this manual copy-paste job which gets out of date).

I'm punting on that for now because there are two possible solutions:
1. Adding a build process and validation process with the `npm` package [`markdown-include`](https://www.npmjs.com/package/markdown-include)
2. Just porting docs to a static site generator like Jekyll/Hugo

Because we might end up doing (2) soon, I don't want to put in the effort for (1) yet. Let me know if you think otherwise.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -82,56 +82,56 @@ See [Parameter Reference][4] for descriptions of each of the important config ke
 
 ```
 spec:
+  serviceAccountName: sonobuoy-serviceaccount
   tolerations:
   - key: node-role.kubernetes.io/master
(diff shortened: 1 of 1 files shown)


timothysc:
Ideally I want to get out of the business of copying around thunks into doc but have a way to crib references.  Generating comes with it's own issues that we'd also have to deal with.

Issue #30, opened 2017-08-07
Configured plugin systemd_logs does not exist


evaldasou:
Hey Heptio Team.

When  I check logs of my just deployed pod it exists with this error: 

```
root:heptio# kubectl logs -f sonobuoy --namespace=heptio-sonobuoy
I0807 13:53:51.548619       1 loader.go:46] Scanning plugins in ./plugins.d (pwd: /)
I0807 13:53:51.548852       1 loader.go:46] Scanning plugins in /etc/sonobuoy/plugins.d (pwd: /)
I0807 13:53:51.548880       1 loader.go:46] Scanning plugins in ~/sonobuoy/plugins.d (pwd: /)
E0807 13:53:51.548907       1 master.go:49] Configured plugin systemd_logs does not exist
```

Any ideas , something is missing on the cluster itself?

kensimon:
I think this is a race condition with how the resources in `examples/quickstart` are created... we create 2 configmaps, one with `/etc/sonobuoy/config.json`, and one with `/etc/sonobuoy/plugins.d/*`... it's possible the pod launched after the config.json was available but before the plugins.d files were.

If you do `kubectl -n heptio-sonobuoy delete pod sonobuoy` and re-run the `kubectl apply -f ./examples/quickstart`, does it work for you then?

evaldasou:
thanks a lot @kensimon ! it worked this time after I have deleted pod!

jamesbuckett:
I am still getting this issue : `Configured plugin systemd_logs does not exist`

I tried the work around as described above but no joy after two attempts.

Environment is : 
```
heptio-sonobuoy Pods
No resources found, use --show-all to see completed objects.

heptio-sonobuoy Services
NAME              CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
sonobuoy-master   10.31.243.189   <none>        8080/TCP   57m

Nodes
NAME                                             STATUS    AGE       VERSION   EXTERNAL-IP      OS-IMAGE                             KERNEL-VERSION
gke-heptio-sonobuoy-default-pool-d1297397-7ztn   Ready     1h        v1.6.7    35.198.193.241   Container-Optimized OS from Google   4.4.52+
gke-heptio-sonobuoy-default-pool-d1297397-cwkf   Ready     1h        v1.6.7    35.198.210.119   Container-Optimized OS from Google   4.4.52+
[...]

PR #95, opened 2017-10-04
Adds more logging at different points of plugin loading

timothysc:
Adds more logging to debug #30.  Right now nothing in the control-flow stands out as a potential issue.  I'm hoping @gamename might have some cycles to repro once this patch has merged.  

I have not seen this issue on my cluster(s).  


timothysc:
This is pretty innocuous, self-merging for debugging.

gamename:
@timothysc I'll give it a try.  This only happens intermittently, so I'll do my best.

PR #99, opened 2017-10-04
[docs] Flatten quickstart directory into quickstart.yaml file

abiogenesis-now:
@timothysc PTAL

Since we removed the component YAML files for the Quickstart, it's probably more clear to have `examples/quickstart.yaml` rather than `examples/quickstart/aggregate.yaml` (where `aggregate.yaml` was the only file in the directory).

This PR makes that change and updates the `Makefile` and other documentation as necessary.

Signed-off-by: Jessica Yao <jessica@heptio.com>


abiogenesis-now:
Dunno why we had a link to the Dockerfile but it didn't make sense in that particular context anyways ("environmental variables") -- removed. This should be good to merge

PR #117, opened 2017-10-14
readme: use SSL vs. SSH for git clone instructions

v1k0d3n:
small thing, but it might be easier for people to use SSL vs. SSH for git clone instructions in the README.


timothysc:
@v1k0d3n could you please --signoff the commit and I'd be happy to merge.

Issue #124, opened 2017-10-30
Setup tools vet & lint tools on test target 


timothysc:
Add govet, golint, and other assorted tools on ```make test``` target 

Talk with @fabioy on creating a standard list.

liztio:
Will be closed by #126

PR #126, opened 2017-11-01
Add vet to tests

liztio:
Our tests should vet, lint, and otherwise verify the hygiene of our code before we allow it into master

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,15 +38,22 @@ BUILDCMD = go build -o $(TARGET) -v -ldflags "-X github.com/heptio/sonobuoy/pkg/
 BUILD = $(BUILDCMD) $(GOTARGET)/cmd/sonobuoy
 
 TESTARGS ?= -v -timeout 60s
-TEST = go test $(TEST_PKGS) $(TESTARGS)
 TEST_PKGS ?= $(GOTARGET)/cmd/... $(GOTARGET)/pkg/...
+TEST = go test $(TEST_PKGS) $(TESTARGS)
+
+VET = go vet $(TEST_PKGS)
+
+# Vendor this someday
+INSTALL_GOLINT = go get -u github.com/golang/lint/golint
+GOLINT_FLAGS ?= -set_exit_status
+LINT = $(INSTALL_GOLINT) && golint $(GOLINT_FLAGS) $(TEST_PKGS)
 
 WORKDIR ?= /sonobuoy
 RBAC_ENABLED ?= 1
 KUBECFG_CMD = $(DOCKER) run \
   -v $(DIR):$(WORKDIR) \
 	--workdir $(WORKDIR) \
-	--rm \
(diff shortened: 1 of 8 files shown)


liztio:
Fixes #124

pkg/config/loader_test.go
```diff
@@ -145,7 +145,7 @@ func TestLoadAllPlugins(t *testing.T) {
 	}
 
 	if len(dsplugin.GetPodSpec().Containers) != 2 {
-		t.Fatalf("JobPlugin should have 1 container, got 2", len(jobplugin.GetPodSpec().Containers))
+		t.Fatalf("JobPlugin should have 2 containers, got %d", len(jobplugin.GetPodSpec().Containers))
```
timothysc: 1 container, got %d

Makefile
```diff
@@ -38,15 +38,21 @@ BUILDCMD = go build -o $(TARGET) -v -ldflags "-X github.com/heptio/sonobuoy/pkg/
...
 TESTARGS ?= -v -timeout 60s
-TEST = go test $(TEST_PKGS) $(TESTARGS)
 TEST_PKGS ?= $(GOTARGET)/cmd/... $(GOTARGET)/pkg/...
+TEST = go test $(TEST_PKGS) $(TESTARGS)
+
+
+# Vendor this someday
+INSTALL_GOLINT = go get -u github.com/golang/lint/golint
+GOLINT_FLAGS ?= -set_exit_status
+VET = go vet $(TEST_PKGS) && $(INSTALL_GOLINT) && golint $(GOLINT_FLAGS) $(TEST_PKGS)
```
timothysc: You could probably break this into 2 var.. 
VET 
LINT

Makefile
```diff
@@ -56,10 +62,13 @@ DOCKER_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_
...
-test: cbuild
-	 $(DOCKER_BUILD) '$(TEST)'
+test: cbuild vet
+	$(DOCKER_BUILD) '$(TEST)'
+
+vet:
+	$(DOCKER_BUILD) '$(VET)'
 
-container: test
+container: test vet
```
timothysc: why is vet needed here?  

test: cbuild vet

Makefile
timothysc: Could also add a gfmt check but now were entering into the whole we should break up the target b/c we're chaining too many things.

pkg/config/loader_test.go
liztio: then why's it != 2?

timothysc: wait now I need to dblcheck

timothysc: my bad, ignore this comment.

PR #127, opened 2017-11-02
Add travis support

liztio:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+sudo: required
+language: go
+services:
+  - docker
+install: true
+
+jobs:
+  include:
+    - stage: test
+      script: make test
+    # - stage: push
+    #   script: echo "TODO"
--- a/Makefile
+++ b/Makefile
@@ -50,14 +50,16 @@ KUBECFG_CMD = $(DOCKER) run \
 	$(KSONNET_BUILD_IMAGE) \
 	kubecfg show -o yaml -V RBAC_ENABLED=$(RBAC_ENABLED) -J $(WORKDIR) -o yaml $< > $@
 
+DOCKER_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c
+
 .PHONY: all container push clean cbuild test local generate
 
 all: container
 
-test:
-	$(DOCKER) run --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c '$(TEST)'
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -50,22 +50,24 @@ KUBECFG_CMD = $(DOCKER) run \
 	$(KSONNET_BUILD_IMAGE) \
 	kubecfg show -o yaml -V RBAC_ENABLED=$(RBAC_ENABLED) -J $(WORKDIR) -o yaml $< > $@
 
+DOCKER_RUN ?= $(DOCKER) run --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c
```
timothysc: Might want to change the alias name.  DOCKER_BUILDER

Makefile
```diff
@@ -50,22 +50,24 @@ KUBECFG_CMD = $(DOCKER) run \
...
+
 .PHONY: all container push clean cbuild test local generate
 
 all: container
 
 test:
-	$(DOCKER) run --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c '$(TEST)'
+	 $(DOCKER_RUN) '$(TEST)'
```
timothysc: want to have the ```test: cbuild``` 

and you could get rid of the other stage above.

liztio: I don't quite understand this. They are running different commands?

.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+language: go
+services:
+  - docker
+install: true
+
+jobs:
+  include:
+    - stage: test
+      script: make test
+    - script: make cbuild
```
chuckha: Is this a separate stage with no name?

Makefile
timothysc: They are, but you probably shouldn't run the tests if the builds fail. 

So making it a dependent target makes sense to me.

.travis.yml
liztio: Unnamed stages inherit the previous stage's name, so you can have multiple simultaneous jobs.

Makefile
liztio: legit!

Makefile
```diff
@@ -50,22 +50,24 @@ KUBECFG_CMD = $(DOCKER) run \
...
 
-container: cbuild
+container: test cbuild
```
timothysc: if you do the ```test: cbuild``` change 

then this can become ```container: test```

timothysc:
@chuckha now to switch the default checks on the repo and setup the webhooks.

Also nix a jenkins job!  It will make @davecheney smile ;-)

PR #132, opened 2017-11-07
Update working directory to be / by default

timothysc:


chuckha:
I have tested this change locally and can confirm this is now needed.

This is the old master using 1.8:
```
$ docker run -it --rm gcr.io/heptio-images/sonobuoy:master /bin/bash
root@d869f3dd616f:/# 
```

You can see the root is `/`.

This is the new dockerfile using 1.9:
```
$ docker run -it --rm gcr.io/heptio-images/sonobuoy:master /bin/ash
/go #
```

You can see the root is `/go`.

LGTM.

Issue #96, opened 2017-10-04
Replace plugins with ksonnet files and lib.


timothysc:
The plugin definitions are a stop gap template that we had in place.  Ideally we could define the plugins in ksonnet and generate the .yaml properly.  

xref: https://github.com/heptio/sonobuoy/blob/master/plugins.d/systemdlogs.yaml#L73

/cc @hausdorff @chuckha

timothysc:
@hausdorff @jessicayuen - Where should I log and track progress of a golang interface library.  

https://github.com/ksonnet/ksonnet ?

timothysc:
stretch goal would be to replace the plugin semantics.

hausdorff:
@timothysc We've [just transitioned](https://github.com/ksonnet/ksonnet/pull/35) ksonnet/ksonent to use go-jsonnet (rather than the cgo) language bindings, so I think we're ready to talk about this whenever you are.

timothysc:
@chuckha this may flip in priority to overtake the interface library... sorry... I know, but there is good reason.

chuckha:
I had a couple of thoughts on this ticket over the weekend:

This ticket is really trying to get at two things and I think we should break it up as such.

## Formalize Plugin Spec

Right now a plugin is required to provide metadata (`driver`, `resultType`, etc) and provide a PodSpec. The driver implementation will use the appropriate Kubernetes resource (Pod/Daemonset) and use the provided PodSpec as well as create a ConfigMap with some metadata.

The plugin definition is not a valid Kubernetes resource, but it would be nice if it were. We can fit a plugin definition into the schema of Kubernetes resources, but we cannot know all the values until runtime. This suggest templating.

### Work Proposal

Extract the parts of the Go that generate ConfigMaps, Pods & Daemonsets (from a plugin definition) and turn it into a Go template that gets evaluated at runtime. We should move the metadata information from the YAML into labels or annotations on the plugin Resources.

This will also require a bit of reshuffling of the plugin logic.

## Ksonnet

The second part of this work is taking the template we create and using ksonnet to define the resources. We can replace the formalized templates we made above with ksonnet and evaluate the ksonnet to generate the Kubernetes resources at runtime.

timothysc:
[...]

PR #133, opened 2017-11-07
Change how plugins work

chuckha:
Instead of creating our own schema for a plugin we will use annotations
to get metadata to sonobuoy instead. We can then define full kubernetes
resources as plugins.

The resources we define are templates. The idea is to eventually move
these templates over to ksonnet and have a nice ksonnet library for the
shared pieces between plugins.

Remaining work:

* Whatever configuration can be passed into the template
should be allowed to be passed into the template (`RESULTS_DIR`).
* Add a test or two to the plugin.Interface implementations

Fixes: #96 


chuckha:
[201711071651_sonobuoy_d2e178db-0eeb-4c35-9001-f3d84be72fc2.tar.gz](https://github.com/heptio/sonobuoy/files/1450899/201711071651_sonobuoy_d2e178db-0eeb-4c35-9001-f3d84be72fc2.tar.gz)

Here is a sample run using this code.

pkg/plugin/loader/loader.go
```diff
@@ -97,107 +96,54 @@ func loadPlugin(namespace string, dfn plugin.Definition, masterAddress string) (
...
-		if err != nil {
-			return plugins, err
-		}
-
-		pluginDef, err := loaderFn(y)
+		pluginTemplate, err := ioutil.ReadFile(fullPath)
 		if err != nil {
-			logrus.Warningf("Error unmarshalling bytes at %v: %v", fullPath, err)
-			continue
+			return pluginDfns, err
```
timothysc: why not return nil,err on failure here?

timothysc: I realize it's nil, but it's less explicit when reading.

examples/quickstart.yaml
```diff
@@ -261,7 +290,7 @@ spec:
       valueFrom:
         fieldRef:
           fieldPath: status.podIP
-    image: gcr.io/heptio-images/sonobuoy:master
+    image: gcr.io/heptio-images/sonobuoy:latest
```
timothysc: should be master

pkg/worker/config.go
```diff
@@ -47,6 +47,7 @@ func LoadConfig() (*plugin.WorkerConfig, error) {
 	viper.BindEnv("masterurl", "MASTER_URL")
 	viper.BindEnv("nodename", "NODE_NAME")
 	viper.BindEnv("resultsdir", "RESULTS_DIR")
+	viper.BindEnv("resulttype", "RESULT_TYPE")
```
timothysc: Look into this one.

chuckha: We do need this. I am providing these values to the worker via Env Vars in the plugin template. This allows me to bypass the ConfigMap entirely.

PR #158, opened 2017-11-21
Make the toleration more flexible

chuckha:
By adding the Exists operator to the toleration we ensure
that we will schedule a systemd pod on the master node in cases
when the master label actually has a value. `...=true:NoSchedule`.

Using the quickstart on an azure cluster I get this output:

```
$ k get po -n heptio-sonobuoy
NAME                                READY     STATUS              RESTARTS   AGE
sonobuoy                            1/1       Running             0          8s
sonobuoy-e2e-job-73ed9d10e8e84d56   0/2       ContainerCreating   0          5s
systemd-logs-1l2rd                  0/2       ContainerCreating   0          5s
systemd-logs-41q5c                  0/2       ContainerCreating   0          5s
systemd-logs-sqspn                  0/2       ContainerCreating   0          5s

$ k get nodes
NAME                    STATUS         AGE       VERSION
k8s-agent-df59102b-0    Ready,agent    1d        v1.7.7
k8s-agent-df59102b-1    Ready,agent    1d        v1.7.7
k8s-agent-df59102b-2    Ready,agent    1d        v1.7.7
k8s-master-df59102b-0   Ready,master   1d        v1.7.7
```

You'll see there are 3 systemd-logs pods and 4 nodes.

With this patch, I reran the quickstart and got:

```
$ k get po -n heptio-sonobuoy
NAME                                READY     STATUS              RESTARTS   AGE
sonobuoy                            1/1       Running             0          4s
sonobuoy-e2e-job-e658a3698dd941ad   0/2       ContainerCreating   0          1s
systemd-logs-7w46d                  0/2       ContainerCreating   0          1s
systemd-logs-bsdqk                  0/2       ContainerCreating   0          1s
systemd-logs-ft569                  0/2       ContainerCreating   0          1s
systemd-logs-v48vm                  0/2       ContainerCreating   0          1s
```

Which is the expected result.

### DIFF
--- a/examples/ksonnet/components/10-configmaps.jsonnet
+++ b/examples/ksonnet/components/10-configmaps.jsonnet
@@ -91,6 +91,7 @@ local sonobuoyConfigData = {
 local tolerations = [
     {
         key: "node-role.kubernetes.io/master",
+        operator: "Exists",
         effect: "NoSchedule",
     },
     {
--- a/examples/quickstart.yaml
+++ b/examples/quickstart.yaml
@@ -165,6 +165,7 @@ data:
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
+        operator: Exists
       - key: CriticalAddonsOnly
         operator: Exists
       volumes:
@@ -250,6 +251,7 @@ data:
           tolerations:
           - effect: NoSchedule
             key: node-role.kubernetes.io/master
+            operator: Exists
           - key: CriticalAddonsOnly
             operator: Exists
           volumes:
--- a/plugins.d/e2e.jsonnet
+++ b/plugins.d/e2e.jsonnet
@@ -24,6 +24,7 @@ local sonobuoyLabels = {
 local tolerations = [
     {
         key: "node-role.kubernetes.io/master",
+        operator: "Exists",
         effect: "NoSchedule",
     },
     {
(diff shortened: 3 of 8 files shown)


examples/ksonnet/components/10-configmaps.jsonnet
```diff
@@ -91,6 +91,7 @@ local sonobuoyConfigData = {
 local tolerations = [
     {
         key: "node-role.kubernetes.io/master",
+        operator: "Exists",
```
timothysc: So do you know why the original reason was failing to schedule on your master nodes?

chuckha: I cannot for the life of me remember. I went through the slack history and nothing triggered for me.

Very likely it was an error on my part (something dumb like forgetting to add tolerations at all) and removing `Exists` still worked in this case since we don't specify any value on the master taint.

After re-reading the documentation and experimenting, I'm very confident this is the change we want. 

The key piece from [here](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) is 

> A toleration “matches” a taint if the keys are the same and the effects are the same, and:
> * the operator is Exists (in which case no value should be specified), or
> * the operator is Equal and the values are equal
> * Operator defaults to Equal if not specified.

chuckha:
Updated to keep plugins.d tolerations in sync with the actual library. We need to clean this stuff up soon.

This also has me wondering if we want the tolerations on the e2e plugins. Seems like no, but I'm wondering why we used to have the taints on the e2es? Was it for something like a single-node cluster/minikube?

PR #164, opened 2017-12-01
Adds a sonobuoy enhancement proposal

chuckha:
Looking forward to feedback. I think this is going to be pretty neat.


docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Goals
+
+ - Define and implement an API to access files in a Sonobuoy tarball.
+ - Support all tarball versions starting at v0.9.0.
+ - Have a familiar API.
+
+### Non-Goals
+
+- We will be unable to customize the tarball for every plugin as people can
```
liztio: Could we make an interface plugin authors could plug in to? Maybe some high-level metadata we expect plugin authors to provide?

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+    // QueryInfos shows how long each object took to query.
+    QueryInfos []*sonobuoy.QueryInfo
+}
+
+// Plugin is a single plugin that executed during a Sonobuoy run.
+type Plugin struct {
+    // Name is the name of the plugin.
+    Name string
+    // Results is a map of plugin result filenames to the file contents.
+    Results map[string][]byte
```
liztio: should these be io.Reader()s as well? These could be potentially pretty big, no?

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+to the core data generated by the [Sonobuoy queries][queries].
+- There will be no support for tarballs before Sonobuoy version 0.9.0.
+
+## Proposal
+
+```go
+type Tarball interface {
+    // Sonobuoy dumps stuff out that can be queried throught the k8s interface.
+    // The source will be a tarball instead of a k8s server.
+    Resources() kubernetes.Interface
```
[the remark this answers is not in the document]
chuckha: [We save quite a bit of the API][queries]. It's not everything and I'm starting to wonder if there is a better way to extract things we care about using a dynamic client instead of hardcoded a set of things we extract. But this may be a question for later.

What do you mean by support selectors? We do allow sonobuoy to be configured to [only extract things that match certain labels or in specific namespaces][filters]

[queries]: https://github.com/heptio/sonobuoy/blob/master/pkg/discovery/queries.go#L114
[filters]: https://github.com/heptio/sonobuoy/blob/master/pkg/config/config.go#L85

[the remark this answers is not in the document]
liztio: I meant that the Corev1 API lets you do things like query labels by selectors. So would something like `Resources().Nodes().List(metav1.ListOptions{LabelSelector:"k8s-app=kubernetes-dashboard"})` work?

docs/enhancements/sep-tarball-client.md
timothysc: Registering a callback for sub-parsing would be useful here.

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+- We will be unable to customize the tarball for every plugin as people can
+write custom plugins. However, there should be an easy way for folks to get
+their results using this client. The main feature we are supporting is access
+to the core data generated by the [Sonobuoy queries][queries].
+- There will be no support for tarballs before Sonobuoy version 0.9.0.
+
+## Proposal
+
+```go
+type Tarball interface {
```
timothysc: s/Tarball/Results

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+    // StatusCode is the result of the healthz endpoint for the node.
+    StatusCode int
+}
+
+// Metadata is all the stuff about the Sonobuoy run and how long it took to query the system.
+type Metadata struct {
+    // Config is the config used during this Sonobuoy run.
+    Config sonobuoy.Config
+    // QueryInfos shows how long each object took to query.
+    QueryInfos []*sonobuoy.QueryInfo
```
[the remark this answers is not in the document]
timothysc: +1

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+    Name string
+    // Logs are the logs in io.Reader form because they can be quite large.
+    Logs io.Reader
+}
+```
+
+Some sample code using this interface:
+
+```go
+data, _ := ioutil.ReadFile("sonobuoy.tar.gz")
```
timothysc: Do you need to read in the whole file?  This won't scale, should we explode it 1st?

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+tarballClient := sonobuoy.NewTarball(data)
```
timothysc: Yeah I think it would be better to walk the results directory vs. reading in the whole thing

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+
+// Getting resources (namespaced or not) should be really familiar.
+_ := tarballClient.Resources().Core().Nodes()
```
timothysc: I don't think we need to follow api conventions here.  There are cluster scoped and NS-scoped resources, it's much more akin to how RBAC's interface.

docs/enhancements/sep-tarball-client.md
chuckha: this entire section was an example of how the api could be used, it is not meant to be used in production.

docs/enhancements/sep-tarball-client.md
chuckha: yeah, that's how we're doing it elsewhere

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+#### https://scanner.heptio.com
+
+- Scanner provides a user interface to Sonobuoy Results. It has to read through
+tarballs generated by Sonobuoy and extract files and support multiple tarball
+versions. It should not have to implement this custom logic and should be able
+to leverage this API.
+
+#### Custom plugin user
+
+- A user has developed a custom plugin and wants an easy way to get the files
```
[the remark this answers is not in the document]
chuckha: This was unclear. I meant sonobuoy plugin. If someone develops a custom sonobuoy plugin there will (probably) be files they want to inspect in the tarball. Likely they will want that to be automated. 

I want to make sure we capture the idea that this client will be flexible enough to work with in a situation where we, the authors of the client, don't know anything about a sonobuoy plugin someone else has written.

docs/enhancements/sep-tarball-client.md
chuckha: yeah, totally. For example, this just won't work if a plugin captures pod logs.

maybe something like extracting the header data of the tar and providing a list of files and then providing an io.Reader to that file

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+  * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+  * [User Stories](#user-stories)
+* [Unresolved Questions](#unresolved-questions)
+
+## Summary
+
+Sonobuoy delivers results in a gzipped tar archive. The structure of the tar
+is [documented here][tar structure] but programmatic access is up to every
+consumer to implement. The structure of the tarball can change between
```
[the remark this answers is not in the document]
chuckha: I actually wonder if we can get away without having to version resources and just provide enough metadata to let the consumer deserialize the objects into the correct version themselves. 

So far, in all of the use cases I have outlined, we will get a ton of milage even without exposing *any* of the k8s resources we gather during a sonobuoy run.

docs/enhancements/sep-tarball-client.md
chuckha: Maybe, yeah. We might tackle that at a later time, I think there is plenty of work to do and that's not really a use case we need to support right off the bat.

docs/enhancements/sep-tarball-client.md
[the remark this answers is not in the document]
chuckha: That's up to the implementation of the interface. I was thinking "yes" when I wrote this doc, but I'm thinking of nixing the entire Resources bit at this point. It provides very little value and will be very hard to get right.

docs/enhancements/sep-tarball-client.md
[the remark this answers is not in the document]
chuckha: sure

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### Tarballs from various Sonobuy versions
+
+- A user has a collection of tarballs across multiple versions of Sonobuoy and
+wants to read the data without having to worry about the implementation of the
+tarball schema directly.
+
+## Unresolved Questions
+
+* None at this time.
```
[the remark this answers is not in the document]
chuckha: That's interesting. I think that could be a neat modification to sonobuoy itself. It could provide a checksum at write time, but it's very hard to know if a run is complete. We'd have to be pretty flexible on that.

Like, if resources don't exist on the cluster they won't exist in the tarball. Did we collect them or did the run fail on collecting a specific resource? 

Still, I think it's possible. It will require some modifications in sonobuoy core (at the very least to write a checksum to the archive) as well as thought into how exactly to know what a complete sonobuoy run looks like.

chuckha:
todo: add user story about what is running on the cluster

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// Host contains the metadata for a given host in a Kubernetes cluster.
+type Host struct {
+    // Name is the name of the host.
+    Name string
+    // Configz is the result of the configz endpoint.
+    KubeletConfig map[string]interface{}
+    // StatusCode is the result of the healthz endpoint for the node.
+    StatusCode int
```
mohitsoni: It will be useful to provide `Ready`/`NotReady` state, as well as whether the kubelet had scheduling enabled or disabled on it.

Although, the `Resources` field in the previous struct might already contain that information. Not sure what's the right thing to do here is.

chuckha:
@liz I think not managing k8s resource versions and using `runtime.Unknown` to push the decision off to the client is pretty future proof-y at the cost of being slightly harder to use.

Other ideas include: deciding to follow upstream versioning and have this library with multiple versions that track k8s resources and the apis that k8s version supports e.g. v1.8 of this library tracks v1.8 and 1.7, etc. I plan to wait for feedback (and also dogfood it once it's available) to see if this would be a better pattern.

As for the sonobuoy bits, we are still willing to change the interface before we go to v1.0.0 so this is kind of at the mercy of the sonobuoy project. If we end up needing to change that drastically (I am hoping not), this may need to change as well.

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,189 @@
...
+        reader, ok := info.Sys().(io.Reader)
+        if !ok {
+            return errors.New("info.Sys() did not return an io.Reader!")
+        }
+        data, _ := ioutil.ReadAll(reader)
+        fmt.println(data)
+    })
+}
+
+podLogs := tarballClient.PodLogs()
```
timothysc: How do you associate all the pod data that you would need to access in order to determine the logs you want?

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,189 @@
...
+- There will be no support for tarballs before Sonobuoy version 0.9.0.
+
+## Proposal
+
+```go
+type Result interface {
+    // NamespacedResources returns a list of resources (the entire collection
+    // can be seen in the sonobuoy config) as `*runtime.Unknown`s to have this
+    // client avoid kubernetes //resource versions.
+    NamespacedResources() []*runtime.Unknown
```
timothysc: You'll want to be able to walk anything that is a list.  

Perhaps instead of outlining the details in the doc which is fraught with perl, include w/ the PR a coded set of interfaces that are light, b/c it will force you to experience the UX.

docs/enhancements/sep-tarball-client.md
```diff
@@ -0,0 +1,189 @@
...
+ - Provide good access to Sonobuoy-generated data.
+
+### Non-Goals
+
+- Manage Kubernetes resource versioning.
+- Customize the client library for every plugin written.
+- There will be no support for tarballs before Sonobuoy version 0.9.0.
+
+## Proposal
+
```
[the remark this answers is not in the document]
chuckha: Yes, absolutely.

docs/enhancements/sep-tarball-client.md
chuckha: That was the idea behind this whole section...design it based on how I'd use it.

I don't really understand why I'd want to walk this list. I would expect a `for` loop to be sufficient, but I must be missing something. Can you explain why I need to walk every list?

timothysc: Folks may want to prefilter the data, and we need a means for them to iterate over the namespace-scoped resources for each resource kind, b/c you can't load them all from disk in 1 go without it hogging the machine, and they may want to skip sections based on 'kind'.

chuckha: Discussed offline. Big clusters are big.

client/client.go
```diff
@@ -0,0 +1,197 @@
...
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+	return out
+}
+
+func (r *results) ServerVersion() *version.Info {
```
chuckha: godoc

client/client.go
```diff
@@ -0,0 +1,197 @@
...
+			err := decoder.Decode(v)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+	return v
+}
+func (r *results) ServerGroups() *metav1.APIGroupList {
```
chuckha: godoc

client/client.go
```diff
@@ -0,0 +1,197 @@
...
+func (r *results) Metadata() *Metadata {
+	out := &Metadata{
+		Config:    config.Config{},
+		QueryData: make([]discovery.QueryData, 0),
+	}
+	r.IterateFilesWithPrefix(MetadataDir, func(path string, info os.FileInfo, err error) error {
+		if path == MetadataDir+"/config.json" {
+			reader, ok := info.Sys().(io.Reader)
+			if !ok {
+				return errors.New("info.Sys() is not a reader")
```
chuckha: can probably extract this error to a const

client/client.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// Metadata
+func (r *results) Metadata() *Metadata {
+	out := &Metadata{
+		Config:    config.Config{},
+		QueryData: make([]discovery.QueryData, 0),
+	}
+	r.IterateFilesWithPrefix(MetadataDir, func(path string, info os.FileInfo, err error) error {
+		if path == MetadataDir+"/config.json" {
```
chuckha: need to make this backwards compatible

client/client.go
```diff
@@ -0,0 +1,197 @@
...
+			if !ok {
+				return errors.New("info.Sys() is not a reader")
+			}
+			decoder := json.NewDecoder(reader)
+			err := decoder.Decode(&out.Config)
+			if err != nil {
+				return err
+			}
+		}
+		if path == MetadataDir+"/query-time.json" {
```
chuckha: also make this backwards compatible

client/client_test.go
```diff
@@ -0,0 +1,138 @@
...
+				})
+				if len(defaultServices) != len(tc.expectedServices) {
+					t.Fatalf("expected %v services found %v", len(tc.expectedServices), len(defaultServices))
+				}
+			}
+		})
+	}
+}
+
+// func TestHosts(t *testing.T) {
```
davecheney: if this test compiles, but doesn't pass, you can use `t.Skip` to return early from the test and avoid checking in commented code.

client/client.go
```diff
@@ -0,0 +1,208 @@
...
+	PodLogs                   = "podlogs"
+	MetadataDir               = "meta"
+	ServerVersionFile         = "serverversion.json"
+	ServerGroupsFile          = "servergroups.json"
+)
+
+// TODO(chuckha): export results
+
+// results holds a pointer to a file that contains Sonobuoy Results.
+type results struct {
```
davecheney: suggestion: make this an io.ReadCloser not an *os.file. It will mean that you won't be able to reopen or rewind the file pointer to create a second pass, but it would make it easier to test, and easier to support results that do not exists on a filesystem.

client/client_test.go
chuckha: ah yeah, it doesn't compile, I only finished converting the first function and have to come back to these.

client/client.go
```diff
@@ -0,0 +1,208 @@
...
+	f *os.File
+}
+
+// NewResults creates a results struct from a filename
+func NewResults(filename string) (*results, error) {
```
davecheney: suggestion

NewResiltsFromFile(filename string) that calls os.Open, then passes the file to NewResults(rc io.ReadCloser)

client/client.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	return &results{
+		f: f,
+	}, nil
+}
+
+// NewReadCloser returns a readCloser where the underlying file is the
+// file held by the results struct. This allows for concurrent reads.
+// The caller is responsible for closing the file.
+func (r *results) NewReadCloser() (io.ReadCloser, error) {
```
davecheney: Does this need to be public?

Makefile
```diff
@@ -45,7 +45,7 @@ BUILDCMD = go build -o $(TARGET) $(VERBOSE_FLAG) -ldflags "-X github.com/heptio/
 BUILD = $(BUILDCMD) $(GOTARGET)/cmd/sonobuoy
```
liztio: should we add a build target here too?

chuckha: I don't think so since we don't have anything to build for this lib

client/client.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// PodLogs walks each pod log file executing walkfn on each.
+func (r *Results) PodLogs(walkfn filepath.WalkFunc) error {
+	return r.IterateFilesWithPrefix(PodLogs, walkfn)
+}
+
+// Metadata extracts the metadata that Sonobuoy gathers from the results.
+// This function does trial and error to detect the version of the results file.
+func (r *Results) Metadata() *Metadata {
```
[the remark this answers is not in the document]
chuckha: I tried to address that in the comments -- maybe a rewording of something like

"This does not depend on `r.Version` because this is the method used to discover the version of the results."

client/client_test.go
```diff
@@ -0,0 +1,366 @@
...
+		expectedServices []string
+	}{
+		{
+			testName:         "no services in the default namespace",
+			namespace:        "default",
+			version:          &version{0, 8},
+			expectedServices: []string{},
+		},
+		{
+			testName:         "no services in the default namespace",
```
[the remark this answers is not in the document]
chuckha: right

client/client_test.go
```diff
@@ -0,0 +1,366 @@
...
+			namespace:        "kube-system",
+			version:          &version{0, 8},
+			expectedServices: []string{},
+		},
+		{
+			testName:         "no services in the default namespace",
```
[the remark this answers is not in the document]
chuckha: yep, next up on my todo

client/client_test.go
```diff
@@ -0,0 +1,366 @@
...
+				t.Fatalf("Got an error reading NamespacedResources: %v", err)
+			}
+			if len(svcs) != len(tc.expectedServices) {
+				t.Fatalf("expected %v services found %v", len(tc.expectedServices), len(svcs))
+			}
+		})
+	}
+}
+
+// func TestMetadata(t *testing.T) {
```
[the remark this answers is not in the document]
chuckha: it sure is!  I'll clean this file up.

PR #167, opened 2017-12-06
SEP: V2 Worker/Master communication protocol proposal

liztio:
SEP is Sonobuoy Enhancement Proposal.

I know I was all gung-ho about grpc for this, but having read the code I no longer think that makes sense.

No rush in reviewing this, ofc!

### DIFF
--- a/docs/enhancements/v2-worker-master.md
+++ b/docs/enhancements/v2-worker-master.md
@@ -1,0 +1,105 @@
+# v2 Worker-Master communications
+
+<!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-refresh-toc -->
+**Table of Contents**
+
+- [v2 Worker-Master communications](#v2-worker-master-communications)
+    - [How it works now](#how-it-works-now)
+    - [Summary](#summary)
+    - [Objectives](#objectives)
+        - [Goals](#goals)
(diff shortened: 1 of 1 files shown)


docs/enhancements/v2-worker-master.md
```diff
@@ -0,0 +1,105 @@
...
+* As an existing Sonobuoy plugin author, it should take little to no work to
+  update my plugin
+* As a new plugin author, it should be clear how I author a new plugin
+* As an author of a new Go-based plugin, I should be able to directly use the
+  client library if I need to
+
+## Unresolved Questions
+
+* Is it worth bumping to `/api/v2` over this change?
+* No Metadata other than node name and plug-in name is currently collected. If
```
timothysc: Yeah, we should banter on this one.  I'm not sold that URL paths are verbose enough.

docs/enhancements/v2-worker-master.md
```diff
@@ -0,0 +1,105 @@
...
+  at some point we wish to do this, there will not be any out-of-band way to
+  collect this other than URL parameters or headers.
+* Do we want to support sending partial results?
```
timothysc: I'm not sold on sending partial results so much as having some form of status pass back.  If the results are not complete they may not be parsable.

Issue #76, opened 2017-09-25
Add support for lite-medium-full tests


timothysc:
There will need to be default configs that support a lite-medium-full set of conformance tests to speed up testing on larger clusters.  

We should also investigate lite-parallelism on the suite to speed it up

/cc @chuckha

timothysc:
This is nice to have, but not a priority for 0.10.0

timothysc:
We may also want other knobs like (slow, medium, fast)

chuckha:
Not blocking but something we'll have to figure out: what do we put in each of these tests? If you have something in mind that's great, otherwise I'll probably need help bucketing tests into most critical, less critical, least critical. Again, not blocking.

timothysc:
@chuckha I'm going to take this one, b/c there are some broader UX issues I'm going to need to work through with @fabioy .  I'm also deep in the think of it atm.

fabriziopandini:
@timothysc what about adding also support for "smoke" tests?
This I will help in uses cases like e.g. local kubeadm development cycles 😉, where someone wants a test just if "it works", which is not compliance, but can be a hint if you are on the right track or not.
Ps. I'm interested in contributing as well, but didn't had time yet to look at the codebase yet ...

timothysc:
After a number of experiments and conversations, we're going to scrap the ksonnet approach and aim for a cli for sonobuoy.  

xref #27

chuckha:
This is closed by the mode option under `gen`, yeah?

timothysc:
We should talk about it tomorrow, we need to make this list complete.

timothysc:
So I'm going to close this, and we will need to define what "quick" means over time.

PR #172, opened 2017-12-13
Example generation updates

chuckha:
* Makefile updates
* Move some jsonnet to a libsonnet file

Related to work on #76


chuckha:
Regenerated yaml is identical.

PR #173, opened 2017-12-13
Rearrange the plugins.d directory for cleanliness.

timothysc:
Originally we had put the plugins.d directory under the main directory so that you could test sonobuoy from the command line without having to make containers etc. 

We've trivialized the `make dev` build and submission now so there is no need for it and as we add example plugins it would get confusing.


chuckha:
The Makefile will also need some updating as the plugins `.tmpl` files are generated from jsonnet.

timothysc:
@chuckha From the looks of it, it was buried in the examples/ksonnet/configmap file.  It didn't look like it was using the plugins.d directory unless I missed it.

chuckha:
@timothysc https://github.com/heptio/sonobuoy/blob/master/Makefile#L108-L111 and https://github.com/heptio/sonobuoy/blob/master/Makefile#L24-L25

chuckha:
Some of it is duplicated, yeah. Not ideal at all. But I'll work on combining it for #76.

liztio:
I dunno how I feel about `pkg` being used as the new home for these. It seems like `pkg` on k8s is usually go packages? Am I mistaken?

timothysc:
They are pretty much just test files now, but let me try to redux it a bit more and update the generation.

timothysc:
k, I gotta a plan for this now.  It'll be a bigger change but cleaner.

timothysc:
I'm revamping the build pretty heavily, so I'll close this PR and open a different one.

Issue #154, opened 2017-11-20
Secure master <> worker and formalize protocol a little better.


timothysc:
Currently the master <> worker is an unsecure rest endpoint that the workers report back results too.  We should clean this up to make it secure and formalize the protocol between the master and worker to be a bit cleaner.  

- investigate grpc vs. simple rest .

liztio:
Proposal in #167

PR #174, opened 2017-12-14
refactor-aggregator-server

liztio:
Factors responsibility for starting the HTTP server out of aggregater/server. This both gets us closer to #154 and allows the tests to be greatly simplified

### DIFF
--- a/pkg/plugin/aggregation/run.go
+++ b/pkg/plugin/aggregation/run.go
@@ -17,7 +17,6 @@ limitations under the License.
 package aggregation
 
 import (
-	"context"
 	"net/http"
 	"strconv"
 	"time"
@@ -101,18 +100,12 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 	go aggr.IngestResults(monitorCh)
 
 	// Ensure we only wait for results for a certain time
-	timeout := make(chan bool, 1)
-	go func() {
-		time.Sleep(time.Duration(cfg.TimeoutSeconds) * time.Second)
-		timeout <- true
-	}()
+	timeout := time.After(time.Duration(cfg.TimeoutSeconds) * time.Second)
 
 	// 5. Wait for aggr to show that all results are accounted for
 	select {
 	case <-timeout:
-		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-		srv.Shutdown(ctx)
-		defer cancel()
+		srv.Close()
 		stopWaitCh <- true
 		return errors.Errorf("timed out waiting for plugins, shutting down HTTP server")
 	case err := <-doneServ:
--- a/pkg/plugin/aggregation/run.go
+++ b/pkg/plugin/aggregation/run.go
@@ -17,6 +17,8 @@ limitations under the License.
 package aggregation
 
 import (
+	"context"
+	"net/http"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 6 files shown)


pkg/plugin/aggregation/server.go
```diff
@@ -18,37 +18,32 @@ package aggregation
...
+	"github.com/sirupsen/logrus"
 )
 
 // Server is a net/http server that can handle API requests for aggregation of
 // results from nodes, sending them back over the Results channel
-type Server struct {
-	// BindAddr is the address for the HTTP server to listen on, eg. 0.0.0.0:8080
-	BindAddr string
+type AggregationHandler struct {
+	mux *http.ServeMux
```
chuckha: If the struct gets embedded directly then the function online 66 can disappear. Also the calls to `handler.mux.Handle` would become `handler.Handle` which seems cleaner.

liztio: @chuckha `http.ServeMux` has an initalisation function that returns a pointer to the struct, the documentation doesn't make the usual guarantees about "a zero-value struct is valid." Looking at the code, there's a map in there which usually needs to get initialised. How does that usually work?

pkg/plugin/aggregation/run.go
```diff
@@ -105,7 +110,9 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 	// 5. Wait for aggr to show that all results are accounted for
 	select {
 	case <-timeout:
-		srv.Stop()
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
timothysc: Why the extra second shutdown cleaning on the timeout condition vs. the hard stop on the timeout?

pkg/plugin/aggregation/server.go
```diff
@@ -18,37 +18,32 @@ package aggregation
...
-	stopCh  chan bool
-	readyCh chan bool
 }
 
 // NewServer constructs a new aggregation server which will listen for results
 // on `bindAddr` and pass them to the given results callback.
-func NewServer(bindAddr string, resultsCallback func(*plugin.Result, http.ResponseWriter)) *Server {
-	return &Server{
-		BindAddr:        bindAddr,
+func NewHandler(resultsCallback func(*plugin.Result, http.ResponseWriter)) http.Handler {
```
timothysc: The comment above no longer makes sense here.

pkg/plugin/aggregation/server.go
chuckha: The ServeMux would still need to be initialized, but the initialization for Aggregation server would look like:

```
handler := &AggregationHandler{
ServeMux: http.NewServeMux(), 
ResultsCallback: resultsCallback,
}
```
instead of defining each field. Omitting field names is also an option.

pkg/worker/worker_test.go
```diff
@@ -41,9 +41,9 @@ func TestRun(t *testing.T) {
 		})
 	}
 
-	withAggregator(t, expectedResults, func(aggr *aggregation.Aggregator) {
+	withAggregator(t, expectedResults, func(aggr *aggregation.Aggregator, baseURL string) {
 		for _, h := range hosts {
-			url := "http://:" + strconv.Itoa(aggregatorPort) + "/api/v1/results/by-node/" + h + "/systemd_logs.json"
+			url := baseURL + "/api/v1/results/by-node/" + h + "/systemd_logs.json"
```
timothysc: We should move a lot of these strings to consts... 

I still don't like explicit URL composition that isn't hidden under a layer of abstraction.

pkg/plugin/aggregation/server.go
liztio: https://stackoverflow.com/questions/12537496/initialize-embedded-struct-in-go lol, just needed to google it 😅

pkg/plugin/aggregation/run.go
liztio: because I didn't `server.Close()` the first time I read the docs 😅

PR #179, opened 2017-12-19
Redefine Worker/Server contract

liztio:
Instead of being based on file extensions, the decision to untar an archive is based on a mime-type. The worker has been updated to use the `application/gzip` mimetype when it encounters a .tar.gz archive.

Additionally, the worker has been updated to take an `http.Client` instead of relying on the default one. This is plumbing for upcoming HTTPS releases.

Part of #154

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -110,6 +110,18 @@
   version = "v0.1.0"
 
 [[projects]]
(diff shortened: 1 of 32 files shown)


liztio:
wrong branch 😞

PR #180, opened 2017-12-19
Refactor aggregation server some more

liztio:
Instead of being based on file extensions, the decision to untar an archive is based on a mime-type. The worker has been updated to use the application/gzip mimetype when it encounters a .tar.gz archive.

Additionally, the worker has been updated to take an http.Client instead of relying on the default one. This is plumbing for upcoming HTTPS releases.

Part of #154


pkg/worker/request.go
```diff
@@ -47,12 +48,13 @@ func DoRequest(url string, callback func() (io.Reader, error)) error {
 			return errors.WithStack(err)
 		}
 		req, err := http.NewRequest(http.MethodPut, url, bytes.NewReader(errbody))
+		req.Header.Add("content-type", mimeType)
```
chuckha: TIL headers are case insensitive

pkg/plugin/aggregation/aggregator.go
```diff
@@ -196,61 +197,71 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
...
+		defer os.Remove(tempFile.Name())
+		if _, err = io.Copy(tempFile, result.Body); err != nil {
+			err = errors.Wrapf(err, "could not write gzip to file")
 			errlog.LogError(err)
 			return err
 		}
 
-		err = os.Remove(resultsFile)
+		// extract to the working directory
+		err = tarx.Extract(tempFile.Name(), resultsDir, &tarx.ExtractOptions{})
```
chuckha: I wish tarx took an io.Reader :/

pkg/plugin/aggregation/server_test.go
```diff
@@ -82,26 +87,56 @@ func TestStart(t *testing.T) {
 		t.Fatalf("Valid request for %v did not get recorded", expectedResult)
 		t.Fail()
 	}
+
+	URL, err = GlobalResultURL(srv.URL, "gztest")
+	if err != nil {
+		t.Fatalf("error getting global result URL %v", err)
+	}
+
+	mimeType := "application/gzip"
```
chuckha: is this worth making a const? i've seen it a fair few times throughout

pkg/plugin/aggregation/aggregator.go
liztio: dare to dream 😞

pkg/plugin/aggregation/aggregator.go
```diff
@@ -201,54 +201,76 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
...
 	}
 
-	if result.MimeType == gzipMimeType {
-		// extract to the working directory
-		err = tarx.Extract(outFile.Name(), resultsDir, &tarx.ExtractOptions{})
+	return nil
+
+}
+
+func (a *Aggregator) handleArchiveResult(result *plugin.Result) error {
```
chuckha: this could be a little more general and not be a method on the aggregator. What do you think about a function that looked like `Extract(readCloser io.ReadCloser, dir string) error`? I was thinking about that as a standalone library, but would be fine inside sonobuoy as well.

liztio: If I wanted to make this general, I'd need to handle a bunch more stuff. I'm not setting uid/guid, not handling symlinks/hardlinks, all that stuff. So I don't want people to see this library and think it's up for general use.

pkg/tarball/tarball.go
```diff
@@ -0,0 +1,61 @@
...
+	"archive/tar"
+	"compress/gzip"
+	"io"
+	"os"
+	"path"
+
+	"github.com/pkg/errors"
+)
+
+func DecodeTarball(reader io.Reader, baseDir string) error {
```
chuckha: just needs a line or two of godoc and it lgtm. super happy to see this :)

pkg/tarball/tarball.go
```diff
@@ -0,0 +1,61 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return errors.Wrap(err, "couldn't opening tarball from gzip")
+		}
+
+		switch header.Typeflag {
+		case tar.TypeDir:
+			if err := os.MkdirAll(path.Join(baseDir, header.Name), os.FileMode(header.Mode)); err != nil {
```
chuckha: oh, one thing i've run into is that sometimes the `header.Name` starts with `./` and sometimes it does not. It might be good to call `filepath.Clean` on `header.Name` when it is used

PR #185, opened 2018-01-02
An easier to use client library

chuckha:


results/client.go
```diff
@@ -41,50 +40,70 @@ const (
...
 	// Version is the version of the archive being inspected.
 	Version string
 }
 
-// OpenArchive will open the archive, detect the version and return a new
-// archive. The file is expected to be a file generated by Sonobuoy.
-// Caller is responsible for closing.
-func OpenArchive(filename string) (*Archive, error) {
-	r, err := OpenArchiveWithVersion(filename, UnknownVersion)
+func NewArchiveFromBytes(data []byte) (*Archive, error) {
```
chuckha: Document this function

results/client.go
```diff
@@ -41,50 +40,70 @@ const (
...
 	if err != nil {
 		return nil, err
 	}
-	return &Archive{zr, version}, nil
+	return &Archive{
+		Version: version,
+	}, nil
+}
+
+func NewArchiveWithVersion(reader io.Reader, version string) *Archive {
```
chuckha: Document this function

results/client.go
```diff
@@ -41,50 +40,70 @@ const (
...
+	return &Archive{
+		reader,
+		version,
+	}
+}
+
+func DiscoverVersion(reader io.Reader) (string, error) {
```
chuckha: Document this function

results/client.go
```diff
@@ -100,9 +119,8 @@ func (t *tarFileInfo) Sys() interface{} {
 	return t.Reader
 }
 
-// WalkFilesWithPrefix will go through every file in the archive and
-// run the walkfn on any file that matches the prefix passed in.
-func (r *Archive) WalkFilesWithPrefix(prefix string, walkfn filepath.WalkFunc) error {
+// WalkFiles walks all of the files in the archive.
+func (r *Archive) WalkFiles(walkfn filepath.WalkFunc) error {
```
chuckha: Improve documentation

results/client.go
```diff
@@ -114,159 +132,93 @@ func (r *Archive) WalkFilesWithPrefix(prefix string, walkfn filepath.WalkFunc) e
...
-	return r.WalkFilesWithPrefix(PluginsDir, walkfn)
-}
+// Functions to be used within a walkfn
 
-// PodLogs walks each pod log file executing walkfn on each.
-func (r *Archive) PodLogs(walkfn filepath.WalkFunc) error {
-	return r.WalkFilesWithPrefix(PodLogs, walkfn)
-}
+// ExtractFileIntoStruct takes data from the `file` and decodes the data into the object.
+func (a *Archive) ExtractFileIntoStruct(file, path string, info os.FileInfo, object interface{}) error {
```
chuckha: Naming things is hard. Feel free to bike shed here.

chuckha: Also considering changing the `file` parameter to a predicate function instead. Then `ExtractConfig` can go away.

timothysc:
I'll dig more in the a.m.  but why is travis wonking out again .?.?.?

chuckha:
travis isn't wonking, out build is broken because `go vet` failed.

results/client.go
```diff
@@ -41,50 +40,70 @@ const (
...
+	err := archive.WalkFiles(func(path string, info os.FileInfo, err error) error {
+		return ExtractConfig(path, info, conf)
+	})
+	if err != nil {
+		return "", err
+	}
+	var version string
+	// Get rid of any of the extra version information that doesn't affect archive layout.
+	// Example: v0.10.0-a2b3d4
+	if strings.HasPrefix(conf.Version, VersionEight) {
```
timothysc: Should use the semver utility and check minver<=xyz<=maxver

timothysc:
I'm digging through this, but we should move this directory under ./pkg to follow upstream practices.

results/client.go
```diff
@@ -27,12 +24,14 @@ const (
 	PodLogs                   = "podlogs/"
 	MetadataDir               = "meta/"
 
-	ServicesFile      = "Services.json"
-	NodesFile         = "Nodes.json"
-	ServerVersionFile = "serverversion.json"
-	ServerGroupsFile  = "servergroups.json"
+	DefaultServicesFile      = "Services.json"
```
timothysc: Do these need to be public?

results/client.go
```diff
@@ -41,50 +40,70 @@ const (
 	VersionTen     = "v0.10"
 )
 
-// Archive holds an io.ReadCloser that can read through the archive once.
+// Archive holds an io.Reader that can read through the archive once.
```
timothysc: It's more of a ResultsReader then an Archive isn't it?

results/client.go
chuckha: No, I think the desired use would be through the methods on the Archive struct. Will make them private.

results/client.go
chuckha: How about ArchiveReader? Mostly I want to avoid `results.Results`

Actually could do `results.Reader`, what do you think of that?

pkg/results/reader.go
```diff
@@ -0,0 +1,227 @@
...
+func ExtractIntoStruct(predicate func(string) bool, path string, info os.FileInfo, object interface{}) error {
+	if predicate(path) {
+		reader, ok := info.Sys().(io.Reader)
+		if !ok {
+			return errors.New("info.Sys() is not a reader")
+		}
+		decoder := json.NewDecoder(reader)
+		err := decoder.Decode(object)
+		if err != nil {
+			return err
```
liztio: might be worth wrapping this

pkg/results/reader.go
```diff
@@ -0,0 +1,227 @@
...
+		return path == ConfigFile(versionTen) || path == ConfigFile(versionEight)
+	}, path, info, conf)
+}
+
+// Functions for helping with backwards compatibility
+
+func (a *Reader) ServerVersionFile() string {
+	switch a.Version {
+	case versionEight:
+		return "serverversion/serverversion.json"
```
liztio: should maybe be a constant to?

pkg/results/reader.go
chuckha: I've got a todo at the top to go through and wrap them all, I'll be refining and refactoring after a merge

pkg/results/reader.go
chuckha: yeah, I'll revisit these in future versions

PR #187, opened 2018-01-03
Detect type to decode

chuckha:

### DIFF
--- a/pkg/results/reader.go
+++ b/pkg/results/reader.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/json"
+	"encoding/xml"
 	"errors"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/results/reader.go
```diff
@@ -155,8 +156,20 @@ func ExtractIntoStruct(predicate func(string) bool, path string, info os.FileInf
 		if !ok {
 			return errors.New("info.Sys() is not a reader")
 		}
+		// TODO(chuckha) there must be a better way
```
timothysc: upstream has the "universal decoder", I'll have to dig up a link.

timothysc: https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/util.go#L244 

legacyscheme.Codecs.UniversalDecoder()

chuckha: i looked at it briefly but it looked like it was pretty k8s specific. Maybe worth looking at deeper though.

For now, I went with the simplest approach possible.

ncdc: UniversalDecoder only decodes what kube's apimachinery can handle. That's usually json, yaml, and proto.

PR #190, opened 2018-01-04
Use TLS and client certs for master/worker communications

### LINKED ISSUE #154: Secure master <> worker and formalize protocol a little better.

liztio:
Closes #154


liztio:
This isn't ready to merge yet, because the infrastructure to wire up the workers with certs isn't done. But what's here now is unlikely to change, so feel free to give it a review @chuckha

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+// Authority represents a root certificate authority that can issues
+// certificates to be used for Client certs.
+// Sonobuoy issues every worker a client certificate
+type Authority struct {
+	privKey    *rsa.PrivateKey
+	cert       *x509.Certificate
+	lastSerial *big.Int
+}
+
+var randReader = rand.Reader
```
chuckha: nit: Combining this with line 23 into a `var` block would be nice

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+	auth := &Authority{
+		privKey: privKey,
+	}
+	cert, err := auth.makeCert(privKey.Public(), func(cert *x509.Certificate) {
+		cert.IsCA = true
+		cert.KeyUsage = x509.KeyUsageCertSign
+		cert.Subject.CommonName = caName
+	})
+	if err != nil {
+		return nil, err
```
chuckha: wrap this  error

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	mut(&tmpl)
+	parent := a.cert
+	// NewAuthority case
+	if a.cert == nil {
+		parent = &tmpl
+	}
+	der, err := x509.CreateCertificate(randReader, &tmpl, parent, pub, a.privKey)
+	if err != nil {
+		return nil, errors.Wrap(err, "coouldn't make authority certificate")
```
chuckha: nit: typo

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+		cert.Subject.CommonName = caName
+	})
+	if err != nil {
+		return nil, err
+	}
+	auth.cert = cert
+	return auth, nil
+}
+
+func (a *Authority) makeCert(pub crypto.PublicKey, mut func(*x509.Certificate)) (*x509.Certificate, error) {
```
chuckha: is `mut` short for mutate? I'd suggest either a comment or a more descriptive name (former if it's common lingo, latter if not)

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+		ExtKeyUsage:           []x509.ExtKeyUsage{},
+		BasicConstraintsValid: true,
+	}
+	mut(&tmpl)
+	parent := a.cert
+	// NewAuthority case
+	if a.cert == nil {
+		parent = &tmpl
+	}
+	der, err := x509.CreateCertificate(randReader, &tmpl, parent, pub, a.privKey)
```
chuckha: `der` is not very descriptive, maybe `newCert`?

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "coouldn't make authority certificate")
+	}
+	cert, err := x509.ParseCertificate(der)
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't re-parse created certificate")
```
chuckha: nit: where is the first parse if this is a re-parse? Does it get parsed during creation?

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+	return a.lastSerial.Add(a.lastSerial, big.NewInt(1))
+}
+
+// CACert is the root certificate of the CA.
+func (a *Authority) CACert() *x509.Certificate {
+	return a.cert
+}
+
+func (a *Authority) CACertPool() *x509.CertPool {
+	pool := x509.NewCertPool()
```
chuckha: Should this be cached so it always returns the same pool? I guess it only matters if we have multiple sonobuoy-servers -- I could imagine that if we had an HA sonobuoy run...but that's probably not very likely.

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+	// Make a copy
+	return a.lastSerial.Add(a.lastSerial, big.NewInt(1))
+}
+
+// CACert is the root certificate of the CA.
+func (a *Authority) CACert() *x509.Certificate {
+	return a.cert
+}
+
+func (a *Authority) CACertPool() *x509.CertPool {
```
chuckha: nit: documentation for this func

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+	pool := x509.NewCertPool()
+	pool.AddCert(a.CACert())
+	return pool
+}
+
+// ServerKey makes a client cert signed by out root CA. The returned certificate
+// has a chain including the root CA cert.
+func (a *Authority) ServerKey(name string) (*tls.Certificate, error) {
```
chuckha: nit: Could this be called `ServerCert`? I would expect `ServerKey` to just return the private key and not the whole cert, but if it makes sense as is, that's fine too.

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// ClientKey makes a client cert signed by out root CA. The returned certificate
+// has a chain including the root CA
+func (a *Authority) ClientKey(name string) (*tls.Certificate, error) {
+	privKey, err := rsa.GenerateKey(randReader, rsaBits)
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't generate private key")
+	}
+	cert, err := a.makeCert(privKey.Public(), func(cert *x509.Certificate) {
```
chuckha: the `mut` function seems to be the only difference in the `ServerKey` and `ClientKey` functions. Maybe there's another way to organize this so we can combine the common code and pass a different callback?

I suggest this because the wrapped errors are identical and it has the potential to be confusing if there is no stack to go with it

pkg/backplane/ca/ca.go
liztio: for reasons known only to the go library authors, create certificate returns DER bytes, and if you want an actual certificate object you have to re-encode it

pkg/backplane/ca/ca.go
liztio: it's called `der` because it's DER-encoded, and I get the certificate from it. but I can add better comments.

pkg/backplane/ca/ca.go
liztio: `tls.Certificate` is an object containing the x509 cert and the private key, so I don't know what to do with it

pkg/backplane/ca/ca.go
chuckha: ahhh gotcha `newDERCert` maybe? comments also fine, thank you for explaining.

liztio:
This PR doesn't add TLS support. Instead, it plumbs the certificates through to the pods (which currently don't use them).

pkg/plugin/aggregation/run.go
```diff
@@ -76,10 +82,18 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 		doneAggr <- true
 	}()
 
+	// TODO (EKF): Do Uncomment when HTTPS is plumbed through
+	//tlsCfg, err := auth.MakeServerConfig(cfg.AdvertiseAddress)
+	if err != nil {
```
chuckha: This stuff can be commented out if it's not used

pkg/plugin/aggregation/run.go
```diff
@@ -90,13 +104,16 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
...
 	for _, p := range plugins {
-		logrus.Infof("Running (%v) plugin", p.GetName())
-		err := p.Run(client, cfg.AdvertiseAddress)
-		// Have the plugin monitor for errors
-		go p.Monitor(client, nodes.Items, monitorCh)
+		cert, err := auth.ClientKeyPair(p.GetName())
 		if err != nil {
+			return errors.Wrapf(err, "couldn't make certificate for plugin %v", p.GetName())
+		}
+		logrus.Infof("Running (%v) plugin", p.GetName())
```
chuckha: We should use the Fields API of logrus to get better structured logging

pkg/plugin/driver/job/job.go
```diff
@@ -108,12 +122,12 @@ func (p *Plugin) FillTemplate(hostname string) ([]byte, error) {
...
 
 // Run dispatches worker pods according to the Job's configuration.
-func (p *Plugin) Run(kubeclient kubernetes.Interface, hostname string) error {
+func (p *Plugin) Run(kubeclient kubernetes.Interface, hostname string, cert *tls.Certificate) error {
 	var (
 		job v1.Pod
 	)
 
-	b, err := p.FillTemplate(hostname) // TODO EKF
+	b, err := p.FillTemplate(hostname, cert) // TODO EKF
```
chuckha: TODO accomplished?

pkg/plugin/driver/daemonset/daemonset.go
```diff
@@ -158,6 +189,30 @@ func (p *Plugin) listOptions() metav1.ListOptions {
...
+func (p *Plugin) createClientSecret(client kubernetes.Interface, key crypto.PrivateKey) error {
+	rsaKey, ok := key.(*rsa.PrivateKey)
+	if !ok {
+		return errors.New("private key not RSA")
+	}
+
+	bytes := x509.MarshalPKCS1PrivateKey(rsaKey)
+	secretName := p.getSecretName()
+
+	secret := &v1.Secret{
```
chuckha: nice good call here

pkg/plugin/driver/daemonset/daemonset.go
```diff
@@ -158,6 +189,30 @@ func (p *Plugin) listOptions() metav1.ListOptions {
 	}
 }
 
+func (p *Plugin) getSecretName() string {
+	return fmt.Sprintf("daemonset-%s-%s", p.GetName(), p.SessionID)
```
chuckha: p.GetSessionID()?

pkg/plugin/driver/job/job.go
```diff
@@ -196,6 +214,30 @@ func (p *Plugin) listOptions() metav1.ListOptions {
 	}
 }
 
+func (p *Plugin) getSecretName() string {
+	return fmt.Sprintf("job-%s-%s", p.GetName(), p.SessionID)
```
chuckha: p.GetSessionID()?

pkg/plugin/driver/job/job.go
```diff
@@ -196,6 +214,30 @@ func (p *Plugin) listOptions() metav1.ListOptions {
...
+}
+
+func (p *Plugin) createClientSecret(client kubernetes.Interface, key crypto.PrivateKey) error {
```
chuckha: This looks identical to the other plugin's implementation. Not necessarily for this PR but maybe we should think about common functions between plugins? plugin util? 😂

timothysc: I also noticed there is a fair amount of overlap that we could reduce over time.

pkg/plugin/driver/daemonset/daemonset.go
```diff
@@ -158,6 +189,30 @@ func (p *Plugin) listOptions() metav1.ListOptions {
...
+
+	bytes := x509.MarshalPKCS1PrivateKey(rsaKey)
+	secretName := p.getSecretName()
+
+	secret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: p.Namespace,
+		},
+		Data: map[string][]byte{clientKeyName: bytes},
```
mattmoyer: I don't really know if there's a compelling reason to use it, but there's a standard Kubernetes Secret `type` for TLS keypairs:
```yaml
apiVersion: v1
kind: Secret
metadata:
  creationTimestamp: null
  name: server-cert
  namespace: heptio-raven
type: kubernetes.io/tls
data:
  tls.crt: "<certificate + chain in PEM format>"
  tls.key: "<private key in PEM format>"
```

I think may be valuable to put the certificate alongside the private key in a Secret. Even though the certificate itself isn't secret, it's tightly bound to the private key so you almost always want them to change as an atomic unit.

pkg/plugin/driver/job/job_test.go
```diff
@@ -1,9 +1,12 @@
 package job
 
 import (
+	"crypto/sha1"
```
mattmoyer: I don't see where these fingerprints are being originally calculated, but we should use SHA-256 for these if possible.  SHA1 is generally deprecated for this use case: https://blog.qualys.com/ssllabs/2014/09/09/sha1-deprecation-what-you-need-to-know

cmd/sonobuoy/app/operations/gen_plugin.go
```diff
@@ -17,7 +17,14 @@ limitations under the License.
 package operations
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
```
mattmoyer: It doesn't really matter security-wise, but since we control both ends of this connection and we know plugins are going to be written in Go, we could safely use ECDSA public keys here, which should be quite a bit faster.

If you're interested I'd look at swapping out `crypto/rsa` for `crypto/ecdsa` with the [`P256`](https://golang.org/pkg/crypto/elliptic/#P256) curve. This is along the lines of Mozilla's ["modern compatibility"](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility) recommendation.

pkg/plugin/driver/job/job_test.go
liztio: @mattmoyer it's literally only used in the tests, so it being faster is an advantage

cmd/sonobuoy/app/operations/gen_plugin.go
liztio: oh sure, can do!

liztio:
Closes #154

cmd/sonobuoy/app/worker.go
```diff
@@ -91,11 +94,17 @@ func runGatherSingleNode(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	client, err := getHTTPClient(cfg)
```
timothysc: shouldn't we retry on the client due to the flakes we are seeing?  There is a separate issue on this but just a PSA.

integration/stress_test.go
```diff
@@ -59,6 +59,7 @@ func TestStress(t *testing.T) {
 	// Launch the aggregator and server
 	aggr := aggregation.NewAggregator(dir+"/results", expected)
 	handler := aggregation.NewHandler(aggr.HandleHTTPResult)
+	//srv := authtest.NewTLSServer(handler, t)
```
timothysc: ?

pkg/templates/manifest.go
```diff
@@ -98,7 +98,8 @@ data:
             "bindport": 8080,
             "timeoutseconds": 5400
         },
-        "Version": "v0.11.0"
+        "Version": "v0.11.0",
```
timothysc: b/c this is in the code now we can/should override with the build arg.

pkg/worker/request.go
```diff
@@ -34,7 +33,7 @@ import (
 // come.)
 func DoRequest(url string, client *http.Client, callback func() (io.Reader, string, error)) error {
 	input, mimeType, err := callback()
-	pesterClient := pester.NewExtendedClient(client)
+	//pesterClient := pester.NewExtendedClient(client)
```
timothysc: ?

pkg/worker/request.go
```diff
@@ -54,7 +53,7 @@ func DoRequest(url string, client *http.Client, callback func() (io.Reader, stri
 		}
 
 		// And if we can't even do that, log it.
-		resp, err := pesterClient.Do(req)
+		resp, err := client.Do(req) //pesterClient.Do(req)
```
timothysc: ?

cmd/sonobuoy/app/worker.go
liztio: the retry is handled by Pester in pkg/worker/request.go

pkg/worker/request.go
liztio: whoops. Testing stuff I forgot to uncomment!

PR #194, opened 2018-01-08
Adds ExtractBytes function for results

chuckha:


pkg/results/reader.go
```diff
@@ -181,6 +196,10 @@ func ExtractConfig(path string, info os.FileInfo, conf *config.Config) error {
 
 // Functions for helping with backwards compatibility
 
+func (a *Reader) MetadataDir() string {
```
timothysc: Nix the "Dir" bit.

pkg/results/reader.go
```diff
@@ -147,6 +147,21 @@ func (r *Reader) WalkFiles(walkfn filepath.WalkFunc) error {
 
 // Functions to be used within a walkfn
 
+// ExtractBytes pulls out bytes into a buffer for any path matching file.
+func ExtractBytes(file string, path string, info os.FileInfo, buf *bytes.Buffer) error {
```
davecheney: try this

if file != path {
  return nil
}
reader, ok := info.Sys().(io.Reader)
 if !ok {
 }
 _, err := buf.ReadFrom(reader)
return err

pkg/results/reader.go
```diff
@@ -194,6 +209,10 @@ func ExtractConfig(path string, info os.FileInfo, conf *config.Config) error {
 
 // Functions for helping with backwards compatibility
 
+func (a *Reader) Metadata() string {
```
davecheney: Please add a comment on this public function

timothysc: Hmm that should have been caught by the build jiggery... 
/cc @liztio

liztio: I think that's part of the linter, which we decided was too flaky to actually use to gate builds?

Issue #196, opened 2018-01-12
Update to 1.9 client and validate api-group fallback across versions. 


timothysc:
In qualifying the qs I only see one minor issue. 

{"level":"error","msg":"error querying PodPresets: the server could not find the requested resource (get podpresets.settings.k8s.io)","time":"2018-01-11T21:45:51Z"}

We'll likely do cascading checks by groups again, I know some elements were promoted to v1 and we should clean those up.

PR #201, opened 2018-01-19
Client go

chuckha:
Fixes: #196


Gopkg.lock
```diff
@@ -311,41 +245,35 @@
   branch = "v2"
   name = "gopkg.in/yaml.v2"
   packages = ["."]
-  revision = "eb3733d160e74a9c7e442f435eb3bea458e1d19f"
+  revision = "d670f9405373e636a5a2765eea47fac0c9bc91a4"
 
 [[projects]]
-  branch = "release-1.8"
+  branch = "master"
```
timothysc: You're going to want to pin these to the branch version.

timothysc: Then call dep-ensure.

chuckha: ahh ok, let me try agian

chuckha:
@timothysc Updated with pinned k8s versions, please take a look.

Gopkg.toml
```diff
@@ -62,15 +62,15 @@
   name = "github.com/viniciuschiele/tarx"
 
 [[constraint]]
-  branch = "master"
+  branch = "release-1.9"
```
sttts: tag kubernetes-1.9.0 if you pin to v6.0.0.

PR #202, opened 2018-01-22
Gate stress test on darwin

chuckha:


chuckha:
I would love to fix my environment!

timothysc:
@liz we could make a separate integration set of tests, but we'd still want them on CI.

liztio:
Than makes sense to me

liztio:
We can close this in favour of #203 probably?

timothysc:
@chuckha could you verify.

chuckha:
nice, thanks for the fix!

PR #203, opened 2018-01-22
seperate stress tests into integration tests

liztio:
Stress tests don't work very well on OSX, so run them
seperately.


.travis.yml
```diff
@@ -4,4 +4,4 @@ services:
   - docker
 install: true
 
-script: VERBOSE=true make test
+script: VERBOSE=true make int
```
timothysc: don't you want `test int`

PR #204, opened 2018-01-22
Sonobuoy Enhancement Proposal for new Plugin format

liztio:

### DIFF
--- a/docs/enhancements/streamlined-templates.md
+++ b/docs/enhancements/streamlined-templates.md
@@ -75,7 +75,8 @@ a plugin. It will also allow us to make changes behind the scenes to the
 Sonobuoy execution model without requiring updates to template files.
 
 All this boiler plate isn't just annoying. It paralyses changes to the Sonobuoy
-execution model. Adding
+execution model. Adding any new environment variables or metadata requires
+changing all existing plugin definitions.
 
 ## Objectives
 
@@ -176,7 +177,7 @@ spec:
 
 ### Proposal 2
 
-In this scenario, the user provides a valid Kubernetes API Document.
+In this scenario, the user provides a (templated) valid Kubernetes API Document.
 
 ```yaml
 apiVersion: v1
@@ -212,3 +213,13 @@ of Go structs.
 * Template for plugin drivers is harder write/understand
 * Extension points are fixed: No new Pod attributes could be added outsite of
   `spec` or Labels.
+
+# Results
+
+Proposal 1 accepted. 
+
+## User stories:
+
+1. As a power user, I would like to see the generated YAML for a plugin.yml file.
+2. As a plugin author, I can write a simplified plugin.yml file
+3. As an end user, I should see no changes as a result of this plugin work
--- a/docs/enhancements/streamlined-templates.md
+++ b/docs/enhancements/streamlined-templates.md
@@ -1,0 +1,214 @@
+# Streamlined Templates
+
+## How it works now
+
+Currently, Sonobuoy plugins are large, sprawling YAML files. This works, but it
+makes the boundary between Sonobuoy and the plugin very messy.
+
+By way of illustration, here is the current `heptio-e2e` plugin file. Every part
+of user configuration is in green, all boilerplate required by Sonobuoy is in
+red.
+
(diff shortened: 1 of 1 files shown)


docs/enhancements/streamlined-templates.md
```diff
@@ -0,0 +1,214 @@
...
+sonobuoy-config:
+  driver: Job
+  plugin-name: heptio-e2e
+  result-type: heptio-e2e
+spec:
+  image: gcr.io/heptio-images/heptio-e2e:master
+    imagePullPolicy: Always
+    name: heptio-e2e
+    volumeMounts:
+    - mountPath: /tmp/results
```
timothysc: I like it, we just need to make certain it's clear that the producer is expected to output to the location of the producer.

docs/enhancements/streamlined-templates.md
```diff
@@ -0,0 +1,214 @@
...
+  namespace: '{{. Namespace}}'
+spec:
+  containers:
+  - image: gcr.io/heptio-images/heptio-e2e:master
+    imagePullPolicy: Always
+    name: heptio-e2e
+    {{.SonobuoyMount}}
+  - {{.SonobuoyConsumer}}
+  {{.SonobuySpecExtras}} 
+```
```
timothysc: I kind of like this better, b/c the transform is explicit.  Implicit transforms could collide and be non-intuitive e.g. what if they had their own name provided that they used for debugging?

docs/enhancements/streamlined-templates.md
```diff
@@ -0,0 +1,214 @@
...
+* Simple substitution
+* Can add additional fields later
+
+#### Disadvantages
+* Not a Kubernetes API object
+* Not as clear how it'll be executed
+
+### Proposal 2
+
+In this scenario, the user provides a valid Kubernetes API Document.
```
chuckha: Might clarify this language a bit, something like the user provides a templatized version of a k8s resource (pod). I was expecting to see a valid yaml definition following, but it definitely looks more like a template than yaml.

docs/enhancements/streamlined-templates.md
```diff
@@ -0,0 +1,214 @@
...
+    sonobuoy-result-type: heptio-e2e
+  labels: {{.SonobuoyLabels}
+  name: sonobuoy-heptio-e2e-job-{{.SessionID}}
+  namespace: '{{. Namespace}}'
+spec:
+  containers:
+  - image: gcr.io/heptio-images/heptio-e2e:master
+    imagePullPolicy: Always
+    name: heptio-e2e
+    {{.SonobuoyMount}}
```
chuckha: I think we will need to add to the contract a bit and require the plugin container be configurable via the `RESULTS_DIR` environment variable. Otherwise we're just guessing where the `done` file gets written to when sonobuoy configures this plugin container.

docs/enhancements/streamlined-templates.md
```diff
@@ -0,0 +1,214 @@
...
+Replace the existing templates with a greatly streamlined one.
+
+### Goals
+* Conceal the execution mechanics from users
+* Allow seamless upgrading of Sonobuoy without user intervention
+
+### Non-goals
+* Do not modify the done contract betweens producer and consumer container
+
+## Proposals
```
chuckha: Couple of ideas that sparked out of this:

1) What if we did away with requiring the plugin author to write to a done file but instead told them to write to stdout. We could run their container in a dind container and wrap up the done-file semantics in that. That would be a nice way to hide those nasty details.

```yaml
spec:
- name: plugin
  image: heptio.com/sonobuoy-plugin-runner:latest
  args:
  - example.com/custom-sonobuoy-plugin:latest
  - any other
  - arguments to the custom plugin
```

2) Could we use a CRD? I don't have much experience here but what if we defined a CRD and some api aggregator that we cleaned up after a run. That might make sonobuoy a bit more flexible for us without needing any external facing templating.

chuckha:
Sometimes I almost feel like the original creation was the cleanest solution. A non k8s yaml document that embedded the pod definition. No templating involved! Just metadata & pod defn.

PR #220, opened 2018-01-31
Sonobuoy CA

liztio:
This PR has been spun out of #190, because that PR was wayyy to big.

@chuckha I tried to follow your requests in the previous PR.

Doesn't add any new functionality, but lays an important foundation for #154

### DIFF
--- a/pkg/backplane/ca/authtest/testserver.go
+++ b/pkg/backplane/ca/authtest/testserver.go
@@ -1,0 +1,65 @@
+package authtest
+
+import (
+	"crypto/tls"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/heptio/sonobuoy/pkg/backplane/ca"
+)
+
+// Server is an extension of httptest.Server that uses our own CA
+type Server struct {
+	*httptest.Server
+	auth *ca.Authority
+	t    *testing.T
+}
+
+// NewServer is a passthrough wrapper for httptest.NewServer. It does not
+// use the CA at all. This provided only for debugging purposes.
+func NewServer(handle http.Handler, t *testing.T) *Server {
+	return &Server{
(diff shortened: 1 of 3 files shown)


pkg/backplane/ca/authtest/testserver.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+}
+
+// NewTLSServer Wraps httptest.NewTLSServer, injecting our CA and TLS config
+func NewTLSServer(handle http.Handler, t *testing.T) *Server {
+	auth, err := ca.NewAuthority()
+	if err != nil {
+		t.Fatalf("Couldn't create certificate authority")
+	}
+	cfg, err := auth.MakeServerConfig("127.0.0.1")
```
timothysc: So is this just for the purposes of testing then?

liztio: Yep. It's a drop-in replacement for [net/http/httptest](https://golang.org/pkg/net/http/httptest/)

pkg/backplane/ca/ca.go
```diff
@@ -0,0 +1,176 @@
...
+	return &tls.Certificate{
+		Certificate: [][]byte{cert.Raw, a.cert.Raw},
+		PrivateKey:  privKey,
+		Leaf:        cert,
+	}, errors.Wrap(err, "couldn't make leaf cert")
+
+}
+
+func (a *Authority) nextSerial() *big.Int {
+	if a.lastSerial == nil {
```
mattmoyer: Not sure if it matters for the intended use case, but this isn't thread safe. It might be nice to add a `sync.Mutex` or add an explicit warning that `ServerKeyPair`/`MakeServerConfig ` cannot be called on the same `Authority` from multiple goroutines.

PR #223, opened 2018-02-01
Implements logs subcommand

### LINKED ISSUE #208: Implement logs subcommand

chuckha:
Closes #208

### DIFF
--- a/cmd/sonobuoy/app/logs.go
+++ b/cmd/sonobuoy/app/logs.go
@@ -20,6 +20,8 @@ import (
 	"os"
 
 	ops "github.com/heptio/sonobuoy/cmd/sonobuoy/app/operations"
+
+	"github.com/heptio/sonobuoy/pkg/config"
 	"github.com/heptio/sonobuoy/pkg/errlog"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -31,15 +33,20 @@ func init() {
 		Short: "Dumps the logs of the currently running sonobuoy containers for diagnostics",
 		Run:   getLogs,
 	}
-	// TODO: Determine options
+	cmd.Flags().BoolP("follow", "f", false, "Specify if the logs should be streamed.")
 	RootCmd.AddCommand(cmd)
 }
 
 func getLogs(cmd *cobra.Command, args []string) {
-	code := 0
-	if err := ops.GetLogs( /*opts*/ ); err != nil {
-		errlog.LogError(errors.Wrap(err, "error attempting to get sonobuoy logs"))
-		code = 1
+	follow, err := cmd.Flags().GetBool("follow")
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "error getting follow flag"))
+		os.Exit(1)
 	}
-	os.Exit(code)
+	cfg := config.NewWithDefaults()
+
+	if err := ops.GetLogs(cfg.PluginNamespace, follow); err != nil {
+		errlog.LogError(errors.Wrap(err, "error attempting to get sonobuoy logs"))
+		os.Exit(1)
+	}
 }
--- a/cmd/sonobuoy/app/operations/logs.go
+++ b/cmd/sonobuoy/app/operations/logs.go
@@ -17,11 +17,66 @@ limitations under the License.
 package operations
 
 import (
-	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/heptio/sonobuoy/cmd/sonobuoy/app/utils"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
-// Logs gathers the logs for the containers in the sonobuoy namespace and prints them out in a structured format
-func GetLogs( /*opts?*/ ) error {
-	// This should be straight forwards
-	return errors.New("not implemented")
+var (
+	podLogSeparator = strings.Repeat("-", 79)
+)
+
+// Logs gathers the logs for the containers in the sonobuoy namespace and prints them
+
+// GetLogs streams logs from the sonobuoy pod by default to stdout.
+func GetLogs(namespace string, follow bool) error {
+	client, err := utils.OutOfClusterClient()
+	if err != nil {
+		return errors.Wrap(err, "could not load kubernetes client")
+	}
+
+	// Follow the main logs if -f is provided
+	if follow {
+		return streamLogs(client, namespace, utils.SonobuoyPod, &v1.PodLogOptions{Follow: true})
+	}
+
+	// Otherwise print every log we can
+	pods, err := client.CoreV1().Pods(namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "could not list pods")
+	}
+	for _, pod := range pods.Items {
+		for _, container := range pod.Spec.Containers {
+			logrus.WithFields(logrus.Fields{"pod": pod.Name, "container": container.Name}).Info("Printing container logs")
+			err := streamLogs(client, namespace, pod.Name, &v1.PodLogOptions{
+				Container: container.Name,
+			})
+			if err != nil {
+				return errors.Wrap(err, "failed to stream logs")
+			}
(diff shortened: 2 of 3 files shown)


cmd/sonobuoy/app/logs.go
```diff
@@ -31,15 +35,55 @@ func init() {
...
+		Follow: follow,
+	}
+	if err := ops.GetLogs(clientset, logOptions); err != nil {
 		errlog.LogError(errors.Wrap(err, "error attempting to get sonobuoy logs"))
-		code = 1
+		os.Exit(1)
+	}
+}
+
+func getClientset() (*kubernetes.Clientset, error) {
```
timothysc: This probably belongs in config as a more general f(n) b/c we do similar things in multiple locations.

cmd/sonobuoy/app/logs.go
```diff
@@ -31,15 +35,55 @@ func init() {
...
+	}
+
+	follow, err := cmd.Flags().GetBool("follow")
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "error getting follow flag"))
+		os.Exit(1)
+	}
+
+	logOptions := &ops.LogOptions{
+		Follow: follow,
```
timothysc: If the number of options over time still = 1 then we just flatten this.

cmd/sonobuoy/app/operations/logs.go
```diff
@@ -17,11 +17,62 @@ limitations under the License.
...
 )
 
-// Logs gathers the logs for the containers in the sonobuoy namespace and prints them out in a structured format
-func GetLogs( /*opts?*/ ) error {
-	// This should be straight forwards
-	return errors.New("not implemented")
+// Logs gathers the logs for the containers in the sonobuoy namespace and prints them
+
+// TODO(chuckha) should this be an option?
+const sonobuoyNamespace = "heptio-sonobuoy"
```
timothysc: It is pulled in from config, we should leverage the init routine to grab that data.

cmd/sonobuoy/app/operations/logs.go
```diff
@@ -17,11 +17,62 @@ limitations under the License.
...
+	}
+
+	// Otherwise print every log we can
+	pods, err := client.CoreV1().Pods(sonobuoyNamespace).List(metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "could not list pods")
+	}
+	for _, pod := range pods.Items {
+		for _, container := range pod.Spec.Containers {
+			logrus.WithFields(logrus.Fields{"pod": pod.Name, "container": container.Name}).Info("Printing container logs")
```
timothysc: You will probably some larger separator for containers.

cmd/sonobuoy/app/operations/logs.go
```diff
@@ -17,11 +17,62 @@ limitations under the License.
...
+			}
+		}
+	}
+	return nil
+}
+
+// TODO(chuckha) the output is a little confusing because our containers already produce structured logs.
+
+func streamLogs(client kubernetes.Interface, podName string, logOptions *v1.PodLogOptions) error {
+	req := client.CoreV1().Pods(sonobuoyNamespace).GetLogs("sonobuoy", logOptions)
```
timothysc: const "sonobuoy" from Config.

cmd/sonobuoy/app/logs.go
```diff
@@ -31,15 +35,55 @@ func init() {
...
 	}
-	// TODO: Determine options
+	cmd.Flags().BoolP("follow", "f", false, "Specify if the logs should be streamed.")
 	RootCmd.AddCommand(cmd)
 }
 
 func getLogs(cmd *cobra.Command, args []string) {
-	code := 0
-	if err := ops.GetLogs( /*opts*/ ); err != nil {
+	clientset, err := getClientset()
```
timothysc: we could build a singleton utility function in config that does this.

cmd/sonobuoy/app/logs.go
chuckha: flattened because we can expand it later.

cmd/sonobuoy/app/logs.go
chuckha: I think I can make LoadClient do a little bit more work and we can use that most places.

cmd/sonobuoy/app/operations/logs.go
```diff
@@ -17,11 +17,66 @@ limitations under the License.
...
+	for _, pod := range pods.Items {
+		for _, container := range pod.Spec.Containers {
+			logrus.WithFields(logrus.Fields{"pod": pod.Name, "container": container.Name}).Info("Printing container logs")
+			err := streamLogs(client, cfg.PluginNamespace, pod.Name, &v1.PodLogOptions{
+				Container: container.Name,
+			})
+			if err != nil {
+				return errors.Wrap(err, "failed to stream logs")
+			}
+			fmt.Println(podLogSeparator)
```
chuckha: :/ not happy about this. Think it's alright?

chuckha: I was also thinking of disabling this line with -q but maybe we can implement that later if need be.

pkg/config/loader.go
```diff
@@ -147,6 +149,39 @@ func LoadClient(cfg *Config) (kubernetes.Interface, error) {
...
+		return nil, errors.Wrap(err, "could not build config from kubeconfig")
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not make a new clientset from config")
+	}
+	return clientset, nil
+}
+
+func locateKubeconfig() string {
```
liztio: Could we use Viper for this maybe? Might be nice to be able to pass in --config or something

cmd/sonobuoy/app/operations/logs.go
```diff
@@ -17,11 +17,66 @@ limitations under the License.
...
 )
 
-// Logs gathers the logs for the containers in the sonobuoy namespace and prints them out in a structured format
-func GetLogs( /*opts?*/ ) error {
-	// This should be straight forwards
-	return errors.New("not implemented")
+// Logs gathers the logs for the containers in the sonobuoy namespace and prints them
+
+// GetLogs streams logs from the sonobuoy pod by default to stdout.
+func GetLogs(cfg *config.Config, follow bool) error {
```
liztio: could `config.Config` be a bit narrower? Seems like a big thunk to pass in to a single log function

pkg/config/loader.go
```diff
@@ -147,6 +149,39 @@ func LoadClient(cfg *Config) (kubernetes.Interface, error) {
 	return clientset, nil
 }
 
+func OutOfClusterClient() (kubernetes.Interface, error) {
```
liztio: This should probably have a comment. What does `OutOfClusterClient` mean?

cmd/sonobuoy/app/operations/logs.go
chuckha: yeah it is, i only need the namespace from the config so no reason to pass it in at all

pkg/config/loader.go
chuckha: I added a comment, the language came from https://github.com/kubernetes/client-go/tree/master/examples 

in cluster is stuff running in the cluster, like sonobuoy https://github.com/heptio/sonobuoy/blob/master/pkg/config/loader.go#L135

and similarly out of cluster is stuff accessing the cluster from outside

liztio: that makes sense! I don't love that it doesn't take any arguments, though. Could it take a kubeconfig filename?

cmd/sonobuoy/app/operations/logs.go
timothysc: I'm ok with it.  Eventually you may want a utility f(n) like prettyPrint for a container with an optional separator arg.

pkg/config/loader.go
timothysc: I believe you can conditionalize to detect in cluster and fallback and we can use in any location then.  

Also //Comment Exported.

cmd/sonobuoy/app/constants/constants.go
```diff
@@ -0,0 +1,5 @@
+package constants
+
+const (
+	SonobuoyPod = "sonobuoy"
```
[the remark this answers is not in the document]
chuckha: yeah true, i'll collapse now and we can split out later when we need it.

PR #226, opened 2018-02-01
`sonobuoy generate` first pass

liztio:
I don't know how to test this because seemingly nothing can handle multiple YAML documents?


cmd/sonobuoy/app/utils/mode/mode.go
```diff
@@ -0,0 +1,99 @@
+package mode
```
chuckha: duplicate package 📦 📦

chuckha: I'd also be ok dropping utils and adding mode/ as a directory under app/, but up to you

cmd/sonobuoy/app/operations/gen.go
```diff
@@ -17,20 +17,52 @@ limitations under the License.
...
 type GenConfig struct {
-	Path string
-	opts RunConfig
+	Path     string
+	ModeName mode.Name
+	Image    string
+}
+
+type templateValues struct {
+	E2EFocus       string
```
timothysc: we'll need skip and provider too.  

Ideally, I was hoping gen/run could use @chuckha's tooling to query the cluster and get the provider settings and fill it in properly for us.

liztio: Neither of those are used by any of the examples right now. What's skip and provider for?

timothysc: skip is the regex list passed to the image, provider right now is plumbing but will be leveraged a lot more in the not-to-distant future.  

We get asked all the time by folks customizing the focus, and skip vars... 
https://github.com/heptio/kube-conformance/blob/master/Dockerfile#L32

PR #227, opened 2018-02-07
update examples to use `master`, generated from `sonobuoy gen`.

liztio:


chuckha:
Should this be the master tag?

https://github.com/heptio/sonobuoy/blob/master/cmd/sonobuoy/app/gen.go#L47 to use :master?

chuckha:
hmm, maybe not, but gen won't work without providing `--sonobuoy-image` until our next release, which is probably ok.

Can you add these files as make targets with the appropriate command to generate them? then lgtm

PR #234, opened 2018-02-09
Copy

chuckha:
not as terrible as last time...


cmd/sonobuoy/app/cp.go
```diff
@@ -37,15 +43,64 @@ func init() {
...
+
+	// This means we left CopyResults early and will report an error.
+	if reader == nil {
+		close(errc)
+		wg.Wait()
+		os.Exit(1)
+	}
+
+	// Extract the tar'd output into a local directory under the prefix.
+	err = ops.UntarAll(reader, outDir, prefix)
```
timothysc: I don't think we should untar, that's up to them to determine.

cmd/sonobuoy/app/operations/cp.go
```diff
@@ -17,10 +17,148 @@ limitations under the License.
...
+		}
+	}()
+	return reader
+}
+
+/** Everything below this marker has been copy/pasta'd from k8s/k8s. The only modification is exporting UntarAll **/
+
+// UntarAll expects a reader that contains tar'd data. It will untar the contents of the reader and write
+// the output into destFile under the prefix, prefix.
+func UntarAll(reader io.Reader, destFile, prefix string) error {
```
timothysc: Just nix this, they can do that work.

cmd/sonobuoy/app/operations/cp.go
```diff
@@ -17,10 +17,148 @@ limitations under the License.
...
+	podNamespace  = "heptio-sonobuoy"
+	containerName = "kube-sonobuoy"
+	results       = "/tmp/sonobuoy"
 )
 
-func CopyResults(path string) error {
-	// pretty straight forwards
-	return errors.New("not implemented")
+// CopyResults copies results from a sonobuoy run into a Reader in tar format.
+func CopyResults(config *rest.Config, cmderr io.Writer, errc chan error) io.Reader {
```
timothysc: We'll have to pass in namespace and the other consts can be in config

cmd/sonobuoy/app/cp.go
```diff
@@ -37,15 +43,64 @@ func init() {
...
-		errlog.LogError(errors.Wrap(err, "error attempting to copy sonobuoy results"))
-		code = 1
+	outDir, err := cmd.Flags().GetString("output")
+	if err != nil {
+		errlog.LogError(fmt.Errorf("failed to get output flag value: %v", err))
+		os.Exit(1)
+	}
+
+	// Find and load the kubeconfig.
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
```
liztio: this is definitely gonna need to be factored out. Do you wanna do it or leave it up to me? :p

cmd/sonobuoy/app/operations/cp.go
liztio: Why the error channel? Are we going to be reporting many errors?

cmd/sonobuoy/app/operations/cp.go
```diff
@@ -17,10 +17,148 @@ limitations under the License.
...
+			}
+			if err := outFile.Close(); err != nil {
+				return err
+			}
+		}
+	}
+
+	if entrySeq == -1 {
+		//if no file was copied
+		errInfo := fmt.Sprintf("error: %s no such file or directory", prefix)
```
liztio: `fmt.Errorf`

cmd/sonobuoy/app/cp.go
chuckha: This is not untaring the sonobuoy results.

We run the command `tar cf - 'path/to/results'` on the pod which will take the contents of an entire directory (which contains our tar.gz results file) and output each file to stdout which we capture with a reader. I'm sure there are other approaches, but this one seems very clean to me.

liztio: And if we do, this should go in operations. I think `app` should just be gathering config and reporting errors.

chuckha: (each file being the 1 file, sonobuoy results and the meta log that also happens to be in the results directory)

cmd/sonobuoy/app/operations/cp.go
chuckha: The error channel is required for capturing errors in the streaming go routine. CopyResults sets up the request, launches a go routine which streams the data and immediately returns the reader where the data being streamed will be accessible. If the goroutine gets an error our command should overall fail (exit 1).

cmd/sonobuoy/app/cp.go
timothysc: The original config code had this if/elsed

cmd/sonobuoy/app/cp.go
chuckha: I suspect we may not want `ops` to write files to disk which is why I am returning a reader. The app is where we are setting useful defaults for people using the `sonobuoy` command which I assert they very much do want to write results to disk. 

However, if another program were to consume the `cp` command, they may not want to be forced into writing to disk and just want to read into memory and inspect the results there.

cmd/sonobuoy/app/cp.go
chuckha: we can both leave it and add a ticket to factor it out.

cmd/sonobuoy/app/operations/cp.go
```diff
@@ -17,10 +17,148 @@ limitations under the License.
...
+		Stdout:    true,
+		Stderr:    true,
+	}, scheme.ParameterCodec)
+	executor, err := remotecommand.NewSPDYExecutor(config, "POST", req.URL())
+	if err != nil {
+		errc <- fmt.Errorf("unable to get remote executor: %v", err)
+		return nil
+	}
+	reader, writer := io.Pipe()
+	go func() {
```
timothysc: Why not make this blocking?

chuckha: so it came like this and i thought about why not make this blocking:

If the file is large, we can stream it straight to disk with this method and not buffer the entire thing in memory. If we make this blocking we'll have to buffer the entire tarball into memory then write to disk.

If you think the complexity of this code outweighs the likelihood of us caring about the size of the tarball I'll be happy to write it blocking, much code can be removed.

cmd/sonobuoy/app/cp.go
chuckha: @timothysc the original loader has homegrown kubeconfig finder (just binding the `kubeconfig` env var) and doesn't look in the default place of the homeDir unless I'm totally missing something. This uses the same loading rules that kubectl uses, so the experience should be what users already expect.

chuckha: (`NewDefaultClientConfigLoadingRules()` also takes windows into account)

PR #235, opened 2018-02-09
Start to change the readme to reduce the complexity.

timothysc:

Fixes: #230


timothysc:
@chuckha @liztio any quick feedback b4 I push this.  I know not all the tooling is there but I want to get it out there just so folks who are using master don't keep prodding me.

README.md
```diff
@@ -4,118 +4,85 @@
...
-> * **In practice, you should make sure that the Sonobuoy pod writes its results to a Persistent Volume.** The quickstart example writes its output to an `emptyDir` volume for simplicity.
->
-> *Troubleshooting errors from `kubectl logs`*:
->  * If you are able to debug and resolve the issue on your own, *make sure to delete and reapply Sonobuoy's YAML manifests*. Otherwise, [file an issue][10].
->
 
 To view the output, copy the output directory from the main Sonobuoy pod to somewhere local:
 ```
-kubectl cp heptio-sonobuoy/sonobuoy:/tmp/sonobuoy ./archive --namespace=heptio-sonobuoy
+$ sonobuoy cp .
```
chuckha: supremely minor nit, this is the default location for the cp command so a `$ sonobuoy cp` would do the same

README.md
```diff
@@ -4,118 +4,85 @@
...
 ```
 
-This should copy a single `.tar.gz` snapshot from the Sonobuoy pod into your local `./archive` directory. You can extract its contents into `./results` with:
+This should copy a single `.tar.gz` snapshot from the Sonobuoy pod into your local `.` directory. You can extract its contents into `./results` with:
```
chuckha: `sonobuoy cp` also copies the run log file as well. We can adjust behavior if this is not wanted. The tar command in the next line is unaffected.

timothysc:
We're going to fine tooth update the docs, so I expect we'll refine over time.

PR #239, opened 2018-02-12
Simplify logs command

chuckha:
This is a non-optimal improvement of the logs command.


cmd/sonobuoy/app/logs.go
```diff
@@ -43,9 +46,18 @@ func getLogs(cmd *cobra.Command, args []string) {
 		errlog.LogError(errors.Wrap(err, "error getting follow flag"))
 		os.Exit(1)
 	}
-	cfg := config.NewWithDefaults()
-
-	if err := ops.GetLogs(cfg.PluginNamespace, follow); err != nil {
+	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(clientcmd.NewDefaultClientConfigLoadingRules(), &clientcmd.ConfigOverrides{})
```
chuckha: The idea here is to make it look just like `cp` so that we can refactor it more easily.

timothysc: I think we should probably push most of this block into config, b/c it will be reused by other commands.   

They basically just need the fn(client, error)

chuckha: Yeah, that's the end goal, this is just an intermediate step to keep size of PRs down. It's really easy to refactor identical code and the logs config stuff was all wonky so this solves the two issues of 1) logs config is wonky 2) make code we want to refactor look the same

liztio: It's pulled out in #242

PR #241, opened 2018-02-12
Wrap errors in results library

chuckha:

### DIFF
--- a/pkg/results/reader.go
+++ b/pkg/results/reader.go
@@ -6,13 +6,13 @@ import (
 	"compress/gzip"
 	"encoding/json"
 	"encoding/xml"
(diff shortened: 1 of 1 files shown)


pkg/results/reader.go
```diff
@@ -59,19 +58,17 @@ func NewReaderFromBytes(data []byte) (*Reader, error) {
 	r := bytes.NewReader(data)
 	gzipReader, err := gzip.NewReader(r)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error creating new gzip reader: %v", err)
```
liztio: why not `errors.Wrap()`?

PR #245, opened 2018-02-13
SEP-4: Status reporting

liztio:

### DIFF
--- a/docs/enhancements/status-reporting.md
+++ b/docs/enhancements/status-reporting.md
@@ -47,7 +47,7 @@ This status should be secure so only the intended audience can see it.
 
 ### For Sonobuoy Aggregator
 
-An annotation will be attached to the sonobuoy pod, constantly updated by the
+An annotation will be attached to the sonobuoy pod, periodically updated by the
 aggregation server.
 
 The annotation will be `sonobuoy.heptio.com/status`.
--- a/docs/enhancements/status-reporting.md
+++ b/docs/enhancements/status-reporting.md
@@ -55,7 +55,6 @@ The annotation will be `sonobuoy.heptio.com/status`.
 ```json
 {
   "status": "running",
-  "percent_complete": 0.5
   "plugins": [
     {
       "name":"systemd",
@@ -90,7 +89,6 @@ it when reporting status to clients.
 ```json
 {
   "status": "failed",
-  "percent_complete": 0.5
   "plugins": [
     {
       "name":"systemd",
--- a/docs/enhancements/status-reporting.md
+++ b/docs/enhancements/status-reporting.md
@@ -11,7 +11,6 @@
     - [Proposal](#proposal)
         - [For Sonobuoy Aggregator](#for-sonobuoy-aggregator)
         - [For `sonobuoy status`](#for-sonobuoy-status)
-        - [For Scanner](#for-scanner)
         - [Heartbeats (optional/future)](#heartbeats-optionalfuture)
         - [User Stories](#user-stories)
     - [Unresolved Questions](#unresolved-questions)
@@ -35,8 +34,6 @@ This status should be secure so only the intended audience can see it.
 ### Goals
 
 * Create a command `sonobuoy status` that shows the results of an ongoing run.
-* Create an API on Aggregator so Scanner can report status instead of the
-  existing spinner
 * (optional) Workers can report their status to the aggregator
 
 ### Non-goals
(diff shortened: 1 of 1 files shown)


docs/enhancements/status-reporting.md
```diff
@@ -0,0 +1,158 @@
...
+heartbeats or more granular status.
+
+The API for this status will look like:
+
+`GET /status`
+
+```json
+{
+  "status": "running",
+  "percent_complete": 0.5
```
timothysc: I don't know how you can compute percent complete unless we plumb some form of ginkgo hook.

docs/enhancements/status-reporting.md
```diff
@@ -0,0 +1,158 @@
...
+
+* Actually updating Scanner is outside of our scope, though we should make sure
+  the API is useful to that team
+* Complex metrics and monitoring. The first pass should be simple.
+
+## Proposal
+
+### For Sonobuoy Aggregator
+
+In addition to the HTTPS server running for the plugin workers, a second HTTPS
```
timothysc: Do we need to secure a separate status endpoint?  Couldn't we post updates as an annotation on deployment object itself?

docs/enhancements/status-reporting.md
liztio: The percent is just % of nodes which have marked themselves as complete

docs/enhancements/status-reporting.md
```diff
@@ -0,0 +1,117 @@
...
+* Actually updating Scanner is outside of our scope, though we should make sure
+  the API is useful to that team
+* Complex metrics and monitoring. The first pass should be simple.
+* Scanner support
+
+## Proposal
+
+### For Sonobuoy Aggregator
+
+An annotation will be attached to the sonobuoy pod, constantly updated by the
```
timothysc: s/constantly/periodicially

docs/enhancements/status-reporting.md
```diff
@@ -0,0 +1,117 @@
...
+  ]
+}
+
+```
+
+### For `sonobuoy status`
+
+`sonobuoy status` will use the kubeconfig acquisition logic from `sonobuoy run`,
+then lookup the `sonobuoy` pod in the provided namespace. The status annotation can
+then be decoded and displayed to the user.
```
timothysc: +100

docs/enhancements/status-reporting.md
```diff
@@ -0,0 +1,117 @@
...
+      "last-heartbeat" "2008-09-15T15:53:00+05:00",
+    }
+  ]
+}
+```
+
+### User Stories
+
+1. As a user running Sonobuoy, I want feedback on my ongoing runs
+2. As a cluster admin, I want my cluster to fail early in the case of DNS or
```
timothysc: We should add preflight checks for this before we even generate.

docs/enhancements/status-reporting.md
```diff
@@ -0,0 +1,117 @@
...
+   Network misconfiguration
+
+## Unresolved Questions
+
+How useful will this status information actually be? Will it make it easier to diagnose failures?
```
timothysc: I think that's where `status` stops and `logs` begins

docs/enhancements/status-reporting.md
```diff
@@ -0,0 +1,117 @@
...
+* Actually updating Scanner is outside of our scope, though we should make sure
+  the API is useful to that team
+* Complex metrics and monitoring. The first pass should be simple.
+* Scanner support
+
+## Proposal
+
+### For Sonobuoy Aggregator
+
+An annotation will be attached to the sonobuoy pod, periodically updated by the
```
chuckha: What sorts of status will this provide that we don't have already? Our aggregation server already knows if stuff is running (is it waiting for results or not). I am unsure how this this proposal is going to improve the UX for scanner when we already tell the user it's running and then report failure after a timeout.

If this is an infrastructural phase, what do you think about writing out the phases explicitly?

What is the end goal?

liztio: The explicit thing we're adding is per-plugin information. So we can report "e2e plugin done, heptio-e2e plugin is still running.

docs/enhancements/status-reporting.md
```diff
@@ -0,0 +1,117 @@
...
+
+### For `sonobuoy status`
+
+`sonobuoy status` will use the kubeconfig acquisition logic from `sonobuoy run`,
+then lookup the `sonobuoy` pod in the provided namespace. The status annotation can
+then be decoded and displayed to the user.
+
+### Heartbeats (optional/future)
+
+The workers will set their own status annotations similar to the aggregators.
```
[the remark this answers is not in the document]
liztio: The worker containers will need to know only the name of the pod they're running in, and I think we can get that from the downwards API. The Plugins themselves will have to learn how to retrieve this information, but a well-placed label should make this easy to do. Because of the new way the plugin definition works, we can easily add this extra information transparently.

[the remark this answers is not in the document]
timothysc: I'd like to keep it scoped, to very simple status, and possible debug level meta-data.

docs/enhancements/status-reporting.md
liztio: The "second phase" is the health information from the individual workers. That information like "this pod is alive, it hasn't given up" is definitely information we don't have right now: now we basically just have "it hasn't timed out yet" and "it's done."

timothysc: Just knowing the state of which plugins are running is useful UX, and a breadcrumb to call logs.

PR #246, opened 2018-02-13
Capture all e2e results if a timeout occurs

chuckha:
This commit introduces an aditional 'soft' timeout.
If the aggregator hits the softtimeout without the plugins finishing it will
tell have the plugins clean themselvfes up. This gives the plugins a chance to
react to being deleted.

Unfortunately k8s has no way to kill one container within a pod so we use the
preStop lifecycle hook. The hook will block k8s from sending a SIGTERM to all
the containers in the pod until the hook finishes (preStop is blocking). The
hook we use gracefully exits the e2e job and then hangs which gives the sonobuoy
worker time to react to the results that it just wrote.

### DIFF
--- a/cmd/sonobuoy/app/worker.go
+++ b/cmd/sonobuoy/app/worker.go
@@ -22,13 +22,20 @@ import (
 	"encoding/pem"
 	"net/http"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
+	"time"
 
 	"github.com/heptio/sonobuoy/pkg/errlog"
 	"github.com/heptio/sonobuoy/pkg/plugin"
 	"github.com/heptio/sonobuoy/pkg/worker"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+)
+
+const (
+	pauseSecondsBeforeShutdown = 20
 )
 
 func init() {
@@ -90,7 +97,22 @@ func loadAndValidateConfig() (*plugin.WorkerConfig, error) {
 	return cfg, nil
 }
 
+// sigHandler is used to manage graceful cleanups when a TERM signal is received.
+func sigHandler() <-chan struct{} {
+	stop := make(chan struct{})
+	go func() {
+		sigc := make(chan os.Signal, 1)
+		signal.Notify(sigc, syscall.SIGTERM)
+		<-sigc
+		// The worker needs to not shutdown immediately upon receiving SIGTERM.
+		time.Sleep(pauseSecondsBeforeShutdown)
+		close(stop)
+	}()
+	return stop
+}
+
 func runGatherSingleNode(cmd *cobra.Command, args []string) {
+	stop := sigHandler()
 	cfg, err := loadAndValidateConfig()
 	if err != nil {
 		errlog.LogError(err)
@@ -107,7 +129,7 @@ func runGatherSingleNode(cmd *cobra.Command, args []string) {
 	// http://sonobuoy-master:8080/api/v1/results/by-node/node1/systemd_logs
 	url := cfg.MasterURL + "/" + cfg.NodeName + "/" + cfg.ResultType
 
-	err = worker.GatherResults(cfg.ResultsDir+"/done", url, client)
+	err = worker.GatherResults(cfg.ResultsDir+"/done", url, client, stop)
 	if err != nil {
 		errlog.LogError(err)
 		os.Exit(1)
@@ -116,6 +138,7 @@ func runGatherSingleNode(cmd *cobra.Command, args []string) {
 }
 
 func runGatherGlobal(cmd *cobra.Command, args []string) {
+	stop := sigHandler()
 	cfg, err := loadAndValidateConfig()
 	if err != nil {
 		errlog.LogError(err)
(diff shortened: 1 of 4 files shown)


pkg/plugin/aggregation/run.go
```diff
@@ -127,24 +127,31 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 	go aggr.IngestResults(monitorCh)
 
 	// Ensure we only wait for results for a certain time
-	timeout := time.After(time.Duration(cfg.TimeoutSeconds) * time.Second)
+
+	// softTimeout will start by cleaning up all plugins, but we aren't in an error state yet.
+	softTimeout := time.After(time.Duration(cfg.TimeoutSeconds) * time.Second)
+	// hardTimeout means we've waited way too long and are in an error state now. Worst case scenario.
+	hardTimeout := time.After(time.Duration(cfg.TimeoutSeconds+300) * time.Second)
```
chuckha: open to suggestions on the `+300` magic number

pkg/plugin/driver/job/job.go
```diff
@@ -186,15 +186,13 @@ func (p *Plugin) Monitor(kubeclient kubernetes.Interface, _ []v1.Node, resultsCh
 // Cleanup cleans up the k8s Job and ConfigMap created by this plugin instance
 func (p *Plugin) Cleanup(kubeclient kubernetes.Interface) {
 	p.cleanedUp = true
-	gracePeriod := int64(1)
```
chuckha: removed because I actually want the default grace period of 30 seconds to give our pods enough time to react. See: https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods for subtlety.

pkg/plugin/aggregation/run.go
liztio: at least make it a constant

plugins.d/e2e.yml
```diff
@@ -14,3 +14,7 @@ spec:
   - mountPath: /tmp/results
     name: results
     readOnly: false
+  lifecycle:
+    preStop:
+      exec:
+        command: ["/bin/sh", "-c", "kill $(pgrep -f e2e.test); sleep 10"]
```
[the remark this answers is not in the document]
liztio: We could probably delete it, it's just `sonobuoy gen --mode quick`

chuckha:
I am interested in this signal escalation, it sounds like something i could use here but am unfamiliar with the vocabulary. I can only find https://github.com/kubernetes/kubernetes/issues/6951. Could you point me to some docs/information?

cmd/sonobuoy/app/worker.go
```diff
@@ -87,7 +90,22 @@ func loadAndValidateConfig() (*plugin.WorkerConfig, error) {
...
 
+// sigHandler is used to manage graceful cleanups when a TERM signal is received.
+func sigHandler() <-chan struct{} {
+	stop := make(chan struct{})
+	go func() {
+		sigc := make(chan os.Signal, 1)
+		signal.Notify(sigc, syscall.SIGTERM)
+		<-sigc
+		// The worker needs to not shutdown immediately upon receiving SIGTERM.
+		time.Sleep(30)
```
chuckha: During a graceful shutdown k8s will send a TERM to all the containers of the pod. The pod contains 1) the worker 2) the e2e test pod. We sleep for a little bit before actually exiting to capture and transmit results.

I was looking into IPC but I wasn't sure it was worth the effort vs a slight delay in shutting down.

timothysc: Maybe make this 20 b/c termination timeout is 30.

timothysc: Shouldn't the worker hold for less then 30?  e.g. 20 in order to send the data back.

pkg/plugin/aggregation/run.go
```diff
@@ -30,6 +30,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	gracefulShutdownPeriod = 45
```
timothysc: Do we need the distinction of can clean up just be the default case for normal timeout conditions?

cmd/sonobuoy/app/worker.go
chuckha: Yeah, 20 probably makes more sense because of the broken case:

If we use 30 seconds: In a normal case I'd expect the plugin to write data to the done file and have the worker pick it up and send it out and clean itself up like normal (because it finished all the work it cared about). In the broken case where the results are not written out in 30 seconds, it's going to simultaneously send a message on the `stop` channel as well as receive a SIGKILL from k8s (because the grace period has expired). Sometimes we'd exit from a SIGKILL sometimes we'd exit cleanly from a message on the `stop` channel.

In the broken case, if 20 seconds elapsed and we still hadn't received results, we still have 10 seconds of graceful cleanup. We'd send a message on the stop channel signaling which would shut the server down and exit with a nil error status. If we still hadn't exited in 10 seconds (unlikely given the code path, but 💻s) then we'd get a SIGKILL and exit.

pkg/plugin/aggregation/run.go
chuckha: A normal timeout shuts down the aggregation server. Instead of that, we want to clean up the plugins first and give them a chance to report the data back to the aggregation server. After some amount of time we can then clean up the aggregation server.

We could bake this logic into the normal shutdown path. This would remove the newly introduced `for{select{` block which might be a good thing but will increase the complexity of a timeout (likely we'd have to sleep for this grace period anyway to give the plugins a chance to respond to being shut down).

chuckha: I think it makes more sense to keep them separate. If we add all this logic to the end cleanup, we have to do something to pause to give the plugins time to send results to the aggregator. If we do something like a sleep instead of this kind of loop, we might miss errors that come back since we'd be blocking this thread on the sleep and not the select statement.

chuckha:
fixing this up based on cleanup branch

PR #258, opened 2018-02-15
Annotation updater

liztio:
Implements [SEP4](https://github.com/heptio/sonobuoy/blob/master/docs/enhancements/status-reporting.md)


pkg/plugin/aggregation/aggregator.go
```diff
@@ -76,11 +76,11 @@ func NewAggregator(outputDir string, expected []plugin.ExpectedResult) *Aggregat
 }
 
 // Wait blocks until all expected results have come in.
-func (a *Aggregator) Wait(stop chan bool) {
+func (a *Aggregator) Wait(stop chan bool, results chan<- *plugin.Result) {
```
liztio: I don't love this. Anyone have ideas?

pkg/backplane/status/status.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Plugin represents the current status of an individual plugin
+type Plugin struct {
+	Plugin string `json:"plugin"`
+	Node   string `json:"node"`
+	Status string `json:"status"`
+}
+
+// Status represents the current status of a Sonobuoy run
+type Status struct {
```
liztio: `status.Status.Status` isn't... great. Any ideas for breaking this up?

pkg/backplane/status/status.go
```diff
@@ -0,0 +1,45 @@
+package status
```
timothysc: Why not call it update?  

e.g. - update.Status

pkg/backplane/status/status.go
timothysc: update.Run.Status ?

pkg/plugin/aggregation/aggregator.go
timothysc: I'm kind of confused what this has todo with an annotation change?

liztio: We need somewhere to tap into the results we're getting so we can update the annotation when we recieve a result.

timothysc: Isn't this very churny on a deamonset plugin, vs. updating a data structure and have periodic batched update?  

e.g. on a 1000 node cluster you'd send 1000 updates.

pkg/backplane/status/status.go
liztio: I think `update` is a bit too vague. Update what? Could be status, plugin versions, kubernetes versions...

timothysc: This is all Run / Job Specific , but I'm also terrible at naming and usually redux it over time.  I don't have strong opinions so whatever makes sense for now is fine with a //TODO - Find a better name...

chuckha: I was reading through some go docs and I think I was wrong to say `status.Status` is bad. The stutter I was confusing it with would be something like `status.NewStatus`. I'm ok with `status.Status` as an exported type.

pkg/backplane/status/update.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Serialize json-encodes the status object
+func (u *Updater) Serialize() (string, error) {
+	u.RLock()
+	defer u.RUnlock()
+	bytes, err := json.Marshal(u.status)
+	return string(bytes), errors.Wrap(err, "couldn't marshall status")
+}
+
+// Annotate serialises the status json, then updates the Sonobuoy pod with the annotation
```
chuckha: how about `...updates the Aggregator pod...`?

pkg/plugin/aggregation/run.go
```diff
@@ -154,3 +167,32 @@ func Cleanup(client kubernetes.Interface, plugins []plugin.Interface) {
...
+			Status: state,
+		}
+
+		if err := updater.Receive(&update); err != nil {
+			logrus.WithFields(
+				logrus.Fields{
+					"node":   update.Node,
+					"plugin": update.Plugin,
+					"status": state,
+					"at":     "sendUpdates",
```
chuckha: If the location of the failure is important we could include the error stacktrace using `pkg/errors`: https://godoc.org/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper

pkg/plugin/aggregation/run.go
```diff
@@ -154,3 +167,32 @@ func Cleanup(client kubernetes.Interface, plugins []plugin.Interface) {
...
+				},
+			).Error(err)
```
chuckha: I think this should be `.WithError(err).WithFields(...).Info("could not receive update")` since we don't really use the Error log level at all. At some point in the distant past we talked about logging and agreed [this logic was pretty good][log].

[log]: https://dave.cheney.net/2015/11/05/lets-talk-about-logging

pkg/plugin/aggregation/run.go
```diff
@@ -154,3 +167,32 @@ func Cleanup(client kubernetes.Interface, plugins []plugin.Interface) {
...
+		}
+	}
+	if err := updater.Annotate(); err != nil {
+		logrus.WithField("at", "sendUpdates").Error(err)
```
chuckha: Error => WithError & Info (see above comment)

pkg/plugin/aggregation/run.go
```diff
@@ -154,3 +167,32 @@ func Cleanup(client kubernetes.Interface, plugins []plugin.Interface) {
 		p.Cleanup(client)
 	}
 }
+
+func sendUpdates(updater *status.Updater, ag *Aggregator) {
```
chuckha: This feels like a pretty important function floating around. It looks like it might make sense to live on the aggregator struct or on the updater struct. I think it would be ok for now. If you think it makes sense to find a home for this, maybe add a TODO. If I'm just over thinking it feel free to ignore

pkg/plugin/aggregation/run.go
liztio: this is an old habit from Heroku's coding style. I can ditch it.

PR #261, opened 2018-02-16
Handle signals from kubernetes

### LINKED ISSUE #214: Scan runs >1 hr, then error occurs and all results are lost
Environment:
```
bootstrapper: minikube
version: 0.25.0
K8s: 1.9.0
Networking: Calico
Container Runtime: CRI-O
```

Test go on for >1hour, then when tries to report back, get the following error (and proceeds to cleanup, which destroys the results):

```
INFO: Failed to get logs from node "minikube" pod "client-envvars-e110f82f-0472-11e8-b965-0a580a010080" container "env3cont": the server rejected our request for an unknown reason (get pods client-envvars-e110f82f-0472-11e8-b965-0a580a010080)
```

chuckha:
* Handles a SIGTERM (pod deletion) more cleanly
* Increased job grace period to 60 seconds
* Tells plugins to cleanup 60 seconds before shutting down the aggregation server

fixes #214

### DIFF
--- a/cmd/sonobuoy/app/worker.go
+++ b/cmd/sonobuoy/app/worker.go
@@ -112,7 +112,6 @@ func runGatherSingleNode(cmd *cobra.Command, args []string) {
 		errlog.LogError(err)
 		os.Exit(1)
 	}
-
 }
 
 func runGatherGlobal(cmd *cobra.Command, args []string) {
--- a/pkg/plugin/constants.go
+++ b/pkg/plugin/constants.go
@@ -1,0 +1,22 @@
+/*
+Copyright 2017 Heptio Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package plugin
+
+const (
+	// GracefulShutdownPeriod is how long plugins have to cleanly finish before they are terminated.
+	GracefulShutdownPeriod = 60
+)
--- a/pkg/worker/worker.go
+++ b/pkg/worker/worker.go
@@ -22,9 +22,12 @@ import (
 	"mime"
 	"net/http"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"syscall"
 	"time"
 
+	"github.com/heptio/sonobuoy/pkg/plugin"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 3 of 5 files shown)


cmd/sonobuoy/app/worker.go
```diff
@@ -90,7 +98,23 @@ func loadAndValidateConfig() (*plugin.WorkerConfig, error) {
...
+// sigHandler is used to manage graceful cleanups when a TERM signal is received.
+func sigHandler() <-chan struct{} {
+	stop := make(chan struct{})
+	go func() {
+		sigc := make(chan os.Signal, 1)
+		signal.Notify(sigc)
+		sig := <-sigc
+		// The worker needs to not shutdown immediately upon receiving SIGTERM.
+		logrus.WithField("signal", sig).Info("got a signal, waiting then sending the real shutdown signal")
+		time.Sleep(pauseBeforeShutdown * time.Second)
```
timothysc: putting the pause in the sig-handler is a little weird imo.

pkg/worker/worker.go
```diff
@@ -39,27 +39,30 @@ func init() {
...
 	for {
-		if inputFileName, err = ioutil.ReadFile(waitfile); err == nil {
-			break
+		select {
+		case <-ticker:
+			if inputFileName, err := ioutil.ReadFile(waitfile); err == nil {
+				return handleWaitFile(string(inputFileName), url, client)
+			}
+		case <-stop:
+			logrus.Info("Forced shutdown. Stopping.")
```
timothysc: Yeah I'd rather see the sleep here than in the handler.

cmd/sonobuoy/app/worker.go
chuckha: The alternative (and most correct) solution here is for the worker to wait until the other process is finished. I was couldn't find a nice way to do this. IPC might be a possibility, but that seemed like a lot of effort for not much gain when a sleep will do just fine.

If you have other suggestions for how to improve the robustness of this I would love to hear them since this is also not my favorite part of the code.

pkg/worker/worker.go
chuckha: ah. ignore that other comment then, that sounds totally reasonable.

cmd/sonobuoy/app/worker.go
```diff
@@ -107,15 +124,15 @@ func runGatherSingleNode(cmd *cobra.Command, args []string) {
...
+	err = worker.GatherResults(cfg.ResultsDir+"/done", url, client, stop)
 	if err != nil {
 		errlog.LogError(err)
 		os.Exit(1)
 	}
-
 }
 
 func runGatherGlobal(cmd *cobra.Command, args []string) {
+	stop := sigHandler()
```
chuckha: should I inline this into 152?

pkg/plugin/constants.go
```diff
@@ -0,0 +1,22 @@
+/*
```
chuckha: I added this file because this constant is actually shared with both aggregation and driver subpackages.

cmd/sonobuoy/app/worker.go
timothysc: If it's only used inside of GatherResults, then I don't see why it needs to be passed in.  From what I see that looks like the only place.

chuckha: my logic there was to make it obvious from the start that signal handling was happening, but i'll move it deeper and see what it looks like

PR #262, opened 2018-02-20
SEP-e2e-commandline proposal

chuckha:

### DIFF
--- a/docs/enhancements/e2e-command-line.md
+++ b/docs/enhancements/e2e-command-line.md
@@ -1,0 +1,81 @@
+# Sub Command for E2E Results Viewing
+
+## Table of Contents
+
+* [Summary](#summary)
+* [Objectives](#objectives)
+  * [Goals](#goals)
+  * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+  * [User Stories](#user-stories)
+* [Unresolved Questions](#unresolved-questions)
+
+## Summary
+
+End to end (e2e) tests in the Kubernetes ecosystem are built on top of ginkgo
+and output junit XML. XML is not designed for human readability and thus it is
+very hard to grok. Users can use scanner but if they have results there is no
+way for a user to upload those results to scanner to get the nice output.
+
+## Objectives
+
+Make the sonobuoy e2e output be more useful to users.
+
+### Goals
+
+- Output a list of e2e tests (filterable by passed/failed).
+- Generate YAML that can rerun failed tests.
+
+### Non-Goals
+
+- Replace scanner.heptio.com
+- Updating scanner features
+- Re-implement gen
+- Make a CLI per plugin. This is a special case that generates a lot of
+  questions and we think this will improve the user experience by a lot.
+
+## Proposal
+
+Add a subcommand to `sonobuoy` that has arguments/subcommands that satisfies the
+goals above:
+
+`sonobuoy e2e <path/to/results_archive>` will show a list of failed tests from
+the sonobuoy archive file.
+
+`sonobuoy e2e --show=passed` will only display passed tests.
+
+`sonobuoy e2e --rerun-failed <path/to/archive>` will figure out the failed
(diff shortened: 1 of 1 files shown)


docs/enhancements/e2e-command-line.md
```diff
@@ -0,0 +1,78 @@
...
+
+1. download the data from the blocking sonobuoy run on the cluster
+2. Parse the results file (or error if it cannot find e2e results)
+3. Display the list of failed tests followed by thee list of passed tests.
+   Ignore skipped tests.
+
+`sonobuoy e2e <path/to/results_archive>` will do steps 2 and 3 from above.
+
+`sonobuoy e2e --show=passed` will only display passed tests `sonobuoy e2e
+--show=failed` will only show failed tests
```
liztio: where do we actually generate the yaml?

docs/enhancements/e2e-command-line.md
```diff
@@ -0,0 +1,78 @@
...
+way for a user to upload those results to scanner to get the nice output.
+
+## Objectives
+
+Make the sonobuoy e2e output be more useful to users.
+
+### Goals
+
+- Output a list of e2e tests (filterable by passed/failed).
+- Generate YAML that can rerun failed tests.
```
timothysc: plugin specific YAML

timothysc: Also, how would this feed into gen & run ?

docs/enhancements/e2e-command-line.md
timothysc: I need to think more about the UX there...

docs/enhancements/e2e-command-line.md
```diff
@@ -0,0 +1,78 @@
...
+   Ignore skipped tests.
+
+`sonobuoy e2e <path/to/results_archive>` will do steps 2 and 3 from above.
+
+`sonobuoy e2e --show=passed` will only display passed tests `sonobuoy e2e
+--show=failed` will only show failed tests
+
+### User Stories
+
+#### Tinkerer
```
timothysc: POC - Operator

docs/enhancements/e2e-command-line.md
chuckha: This will be the first user of the operators for gen and run. Presumably this will reuse the gen operator to generate some yaml and the run operator to run it. If you were thinking of a different approach I'd be happy to hear it and take your thoughts into account as I update this.

docs/enhancements/e2e-command-line.md
chuckha: @liztio do you mean for the rerun command? It would be in the operator form the e2e subcommand I'm guessing unless as i'm writing it it fits better in the cmd.

@timothysc Can you be a bit more wordy about what you're thinking? My thought behind --show was that I hate the word "filter" because does it mean filter out or filter for? I think show 
is very explicit for this, but am not at all wed to it. Happy to take suggestions and brainstorm if you can tell me what you think the problem is/what you want to optimize for.

chuckha:
@liztio are you looking for implementation details? I don't understand what you mean by flesh out.

I added a comment with how I think it's going to be implemented, but I think the SEP should avoid detailed implementation since that should be left up to the implementer.

Perhaps a few more words to justify would do well:

The generated YAML would reuse the gen operator and generate very targeted YAML for only rerunning specific e2e tests. It would not recreate the original sonobuoy scan yaml but a very reduced (think scanner YAML) subset of features to target individual e2e tests.

I suspect there will be a bit of infrastructure work to get this feature in place, like templatizing more pieces of the gen template.

docs/enhancements/e2e-command-line.md
timothysc: It's just a flag default question.  Most folks who run this will want to see a summary of what failed, or if it all passed.  

So I think defaulting `--show=failed` makes sense, with an options plugin.d YAML output.

PR #263, opened 2018-02-20
Sonobuoy status

### LINKED ISSUE #209: Implement status subcommand

liztio:
Example status:
```
+--------------+-----------------------------+----------+
|    PLUGIN    |            NODE             |  STATUS  |
+--------------+-----------------------------+----------+
| e2e          |                             | running  |
| systemd_logs | ip-10-0-18-159.ec2.internal | complete |
| systemd_logs | ip-10-0-22-130.ec2.internal | complete |
| systemd_logs | ip-10-0-22-8.ec2.internal   | complete |
+--------------+-----------------------------+----------+
|                                              RUNNING  |
+--------------+-----------------------------+----------+
```

closes #209

### DIFF
--- a/cmd/sonobuoy/app/status.go
+++ b/cmd/sonobuoy/app/status.go
@@ -17,9 +17,11 @@ limitations under the License.
 package app
 
 import (
+	"fmt"
 	"os"
+	"strings"
+	"text/tabwriter"
 
-	"github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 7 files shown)


liztio:
I think adding the tablewriter greatly enhances usability for users, but we could do it more messily if we don't want the extra dependency.

cmd/sonobuoy/app/status.go
```diff
@@ -19,28 +19,61 @@ package app
...
 	"os"
 
-	ops "github.com/heptio/sonobuoy/cmd/sonobuoy/app/operations"
-	"github.com/heptio/sonobuoy/pkg/errlog"
+	"github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/heptio/sonobuoy/cmd/sonobuoy/app/args"
```
chuckha: This shouldn't be pulling in args, see https://github.com/heptio/sonobuoy/pull/264

pkg/plugin/aggregation/update.go
```diff
@@ -13,7 +13,10 @@ import (
 	"github.com/heptio/sonobuoy/pkg/plugin"
 )
 
-const annotationName = "sonobuoy.hept.io/status"
+const (
+	StatusAnnotationName = "sonobuoy.hept.io/status"
+	StatusPodName        = "sonobuoy"
+)
 
 type key struct {
```
chuckha: Please document this composite key. Maybe something about how a plugin is uniquely defined by the node it runs on and its name (which isn't actually true as we would also need the sonobuoy id to uniquely define a plugin).

pkg/plugin/aggregation/update.go
```diff
@@ -22,21 +25,20 @@ type key struct {
 // Updater manages setting the Aggregator annotation with the current status
 type Updater struct {
```
chuckha: Does anything here need to be exported?

cmd/sonobuoy/app/status.go
liztio: I thought `cmd` was the right place for args and `operations` was the wrong place for args.

liztio:
We don't support multiple sonobuoy runs in the same namespace, aiui. Parallel sonobuoy runs would be in different namespaces, and you pass in a `--namespace` flag for that reason.

cmd/sonobuoy/app/status.go
chuckha: yeah, my mistake.

chuckha:
ah ha ok, knew i must have been missing something

pkg/plugin/aggregation/update.go
liztio: probably not!

chuckha:
could we use https://golang.org/pkg/text/tabwriter/ instead of the tablewriter? would that still greatly enhance usability while not pulling in an external dep?

PR #264, opened 2018-02-21
Shuffles gen args back to cmd and operations to prevent dep-graph issues

timothysc:
This commit shuffles the args back closer to the commands and moves around some of the objects. 

I isolated at just gen and added a bunch of //TODOs for follow-on PRs.  


cmd/sonobuoy/app/gen.go
```diff
@@ -39,21 +42,40 @@ var GenCommand = &cobra.Command{
...
-	args.AddModeFlag(&genopts.ModeName, GenCommand)
-	args.AddNamespaceFlag(&genopts.Namespace, GenCommand)
-
+	AddGenFlags(&genopts, GenCommand)
 	RootCmd.AddCommand(GenCommand)
 }
 
+// AddGenFlags adds generation flags to a command
+func AddGenFlags(gen *ops.GenConfig, cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVarP(
```
chuckha: Do these need to be persistent flags? That to me implies there is a subcommand and this is a global option flag for this command and all children commands.

cmd/sonobuoy/app/gen.go
```diff
@@ -39,21 +42,40 @@ var GenCommand = &cobra.Command{
...
+	)
+	// TODO: This variable default needs saner image defaults from ops.f(n) or config
+	cmd.PersistentFlags().StringVarP(
+		&gen.Namespace, "namespace", "n", "heptio-sonobuoy",
+		"The namespace to run Sonobuoy in. Only one Sonobuoy run can exist per namespace simultaneously.",
+	)
+	// TODO: Need to provide ability to override config structure and allow for sane defaults
+	// TODO: Need to provide ability to override e2e-focus
+	// TODO: Need to provide ability to override e2e-skip
+}
```
chuckha: Out of scope for this PR, but what do you think about being able to include/exclude plugins with this command? something like `sonobuoy gen --plugins=e2e --e2e-mode=quick` (Then we get into a weird space where a plugin author could integrate with the sonobuoy command by providing a flag set to add their plugin flags `--plugins=e2e,bulkhead --bulkhead-option=abcd`).

cmd/sonobuoy/app/gen.go
```diff
@@ -39,21 +42,40 @@ var GenCommand = &cobra.Command{
...
 	RootCmd.AddCommand(GenCommand)
 }
 
+// AddGenFlags adds generation flags to a command
+func AddGenFlags(gen *ops.GenConfig, cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVarP(
+		&mode, "e2e-mode", "", string(ops.Conformance),
+		fmt.Sprintf("What mode to run sonobuoy in. [%s]", strings.Join(ops.GetModes(), ", ")),
+	)
+	// TODO: This variable default needs saner image defaults from ops.f(n) or config
```
chuckha: We've been trying to keep the suggestion provided by @davecheney to annotate TODOs with people so context can be gleaned for future developers. Adding a name does not at all imply ownership, simply the name of the person that wrote the comment. If we all like this idea we should formalize it along with other decisions we've made (this is on my backlog).

example:
`// TODO(chuckha) make this code more betterer.`

cmd/sonobuoy/app/gen.go
timothysc: Doesn't hurt if we somehow expanded, but if they were local then we would need to shuffle them all.

cmd/sonobuoy/app/gen.go
timothysc: Let's tackle the plugin question separately, I have thoughts there and we should leverage config.

cmd/sonobuoy/app/operations/gen.go
```diff
@@ -45,10 +44,10 @@ type templateValues struct {
 }
 
 // GenerateManifest fills in a template with a Sonobuoy config
-func GenerateManifest(cfg GenConfig) ([]byte, error) {
-	mode := cfg.ModeName.Get()
+func (g *GenConfig) GenerateManifest() ([]byte, error) {
```
liztio: I don't know that I like this. Having a `Config` object have methods to execute feels weird. What does this gain us over the old style?

cmd/sonobuoy/app/gen.go
```diff
@@ -19,16 +19,19 @@ package app
...
 
-	"github.com/heptio/sonobuoy/cmd/sonobuoy/app/args"
 	ops "github.com/heptio/sonobuoy/cmd/sonobuoy/app/operations"
 	"github.com/heptio/sonobuoy/pkg/errlog"
 )
 
 var genopts ops.GenConfig
+var mode string
+
+type foo string
```
liztio: what's this for?

cmd/sonobuoy/app/gen.go
```diff
@@ -39,21 +42,40 @@ var GenCommand = &cobra.Command{
...
+	)
+	// TODO: Need to provide ability to override config structure and allow for sane defaults
+	// TODO: Need to provide ability to override e2e-focus
+	// TODO: Need to provide ability to override e2e-skip
+}
+
 func genManifest(cmd *cobra.Command, args []string) {
-	code := 0
-	bytes, err := ops.GenerateManifest(genopts)
+	err := genopts.ModeName.Set(mode)
```
liztio: this could be collapsed into a single if statement.

cmd/sonobuoy/app/gen.go
```diff
@@ -19,16 +19,19 @@ package app
...
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
-	"github.com/heptio/sonobuoy/cmd/sonobuoy/app/args"
 	ops "github.com/heptio/sonobuoy/cmd/sonobuoy/app/operations"
 	"github.com/heptio/sonobuoy/pkg/errlog"
 )
 
 var genopts ops.GenConfig
```
liztio: I don't think this needs to be a global variable anymore. It only was before so we could use it from `init`.

cmd/sonobuoy/app/gen.go
```diff
@@ -39,21 +42,40 @@ var GenCommand = &cobra.Command{
...
 }
 
+// AddGenFlags adds generation flags to a command
+func AddGenFlags(gen *ops.GenConfig, cmd *cobra.Command) {
+	cmd.PersistentFlags().StringVarP(
+		&mode, "e2e-mode", "", string(ops.Conformance),
+		fmt.Sprintf("What mode to run sonobuoy in. [%s]", strings.Join(ops.GetModes(), ", ")),
+	)
+	// TODO: This variable default needs saner image defaults from ops.f(n) or config
+	cmd.PersistentFlags().StringVarP(
```
liztio: I think it'd be better to either use gen.* everywhere or nowhere. Makes it easier to know where to add new options.

cmd/sonobuoy/app/gen.go
```diff
@@ -39,21 +42,40 @@ var GenCommand = &cobra.Command{
...
+	// TODO: Need to provide ability to override config structure and allow for sane defaults
+	// TODO: Need to provide ability to override e2e-focus
+	// TODO: Need to provide ability to override e2e-skip
+}
+
 func genManifest(cmd *cobra.Command, args []string) {
-	code := 0
-	bytes, err := ops.GenerateManifest(genopts)
+	err := genopts.ModeName.Set(mode)
 	if err == nil {
```
liztio: this could be the more familiar `if err != nil { os.exit() }`

cmd/sonobuoy/app/run.go
```diff
@@ -36,9 +36,9 @@ func init() {
 		Run:   submitSonobuoyRun,
 		Args:  cobra.ExactArgs(0),
 	}
-	args.AddModeFlag(&runopts.GenConfig.ModeName, cmd)
-	args.AddSonobuoyImageFlag(&runopts.GenConfig.Image, cmd)
-	args.AddNamespaceFlag(&runopts.GenConfig.Namespace, cmd)
+	AddGenFlags(&runopts.GenConfig, cmd)
+
+	// TODO: reduce this too.
```
liztio: what do you mean by `reduce`?

cmd/sonobuoy/app/gen.go
timothysc: Updated.

cmd/sonobuoy/app/operations/run.go
```diff
@@ -20,7 +20,9 @@ import (
 	"bytes"
 	"io"
 
+	// TODO - remove kubeconfig too.
```
liztio: We need a kubeconfig somewhere so we can actually execute. What do you intend to replace it withL

cmd/sonobuoy/app/gen.go
timothysc: Whoops leftover experiment.

cmd/sonobuoy/app/gen.go
timothysc: It can't actually or else I would need var at function scope.  

It's used below.

liztio: ah, ignore me!

cmd/sonobuoy/app/operations/gen.go
timothysc: I was planning to expand the number of operations GenConfig struct, such as LoadDefaults.  I could make it more procedural, but every operation would be passing in a GenConfig. 

I was also thinking about the consumer of the library and general UX.  When they include a library, Object[.]F(n) will be auto-discovered vs hunting a function.

cmd/sonobuoy/app/gen.go
timothysc: I "could" flip the logic, but this reduces more cleanly.

cmd/sonobuoy/app/gen.go
timothysc: I must be missing something, it is still used by init.

timothysc:
I'll submit follow on PRs for some of the todos and we should do a walk through as well.

PR #266, opened 2018-02-21
Refactor and consolidate command arguments

liztio:
My proposal for how arguments could work in the future.

Doesn't work 100% yet (mode fails)

### DIFF
--- a/cmd/sonobuoy/app/status.go
+++ b/cmd/sonobuoy/app/status.go
@@ -26,15 +26,14 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/heptio/sonobuoy/cmd/sonobuoy/app/args"
 	ops "github.com/heptio/sonobuoy/cmd/sonobuoy/app/operations"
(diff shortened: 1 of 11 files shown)


cmd/sonobuoy/app/status.go
```diff
@@ -26,15 +26,14 @@ import (
...
-	"github.com/heptio/sonobuoy/cmd/sonobuoy/app/args"
 	ops "github.com/heptio/sonobuoy/cmd/sonobuoy/app/operations"
 	"github.com/heptio/sonobuoy/pkg/errlog"
 )
 
-var statusOpts struct {
-	namespace string
-	config    args.Kubeconfig
-}
+var (
```
liztio: what do we think about this? The advantage of the anonymous struct is it namespaces these variables.

timothysc: I don't have strong opinions outside of the client interface.  

Once it gets to the client we will likely want to wrap in a struct as signatures have a tendency to change over time.

timothysc:
I'm going to merge this so I can rebase my changes.

PR #269, opened 2018-02-22
Cleanup safe Exit(0) scenarios

timothysc:
/cc @liztio

### DIFF
--- a/cmd/sonobuoy/app/master.go
+++ b/cmd/sonobuoy/app/master.go
@@ -45,7 +45,6 @@ func init() {
 }
 
 func runMaster(cmd *cobra.Command, args []string) {
-	exit := 0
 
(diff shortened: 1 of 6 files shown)


cmd/sonobuoy/app/master.go
```diff
@@ -61,14 +60,12 @@ func runMaster(cmd *cobra.Command, args []string) {
...
-	}
+	errcount := discovery.Run(kubeClient, cfg)
 
 	if noExit {
 		logrus.Info("no-exit was specified, sonobuoy is now blocking")
 		select {}
 	}
 
-	os.Exit(exit)
+	os.Exit(errcount)
```
liztio: I don't love this? I think exit code should be 1 no matter what, unless we want to assign extra meanings for the exit codes

PR #278, opened 2018-02-26
Remove docs and we will use godoc for client side interface tooling.

timothysc:

/cc @Bradamant3 @chuckha


timothysc:
We already chatted on slack so just merging this one.

PR #281, opened 2018-02-26
The following PR changes how the master kubeconfig is loaded

timothysc:
and eliminates the kubeconfig settings in sonobuoys config struct.

I still need to finish my testing on it and am breaking up commits to each part. 

/cc @liztio

### DIFF
--- a/cmd/sonobuoy/app/kubeconfig.go
+++ b/cmd/sonobuoy/app/kubeconfig.go
@@ -37,6 +37,7 @@ func (c *Kubeconfig) Set(str string) error {
 
 // Get returns a rest Config, possibly based on a provided config
 func (c *Kubeconfig) Get() (*rest.Config, error) {
+
 	if c.ClientConfigLoadingRules == nil {
 		c.ClientConfigLoadingRules = clientcmd.NewDefaultClientConfigLoadingRules()
 	}
--- a/cmd/sonobuoy/app/master.go
+++ b/cmd/sonobuoy/app/master.go
@@ -25,9 +25,12 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+
+	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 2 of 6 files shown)


cmd/sonobuoy/app/kubeconfig.go
```diff
@@ -37,6 +37,14 @@ func (c *Kubeconfig) Set(str string) error {
 
 // Get returns a rest Config, possibly based on a provided config
 func (c *Kubeconfig) Get() (*rest.Config, error) {
+
+	// detect if in-cluster configuration makes sense
+	if len(c.ExplicitPath) > 0 {
```
timothysc: whoops this is backwards

cmd/sonobuoy/app/master.go
```diff
@@ -25,9 +25,12 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+
```
liztio: why the space here?

cmd/sonobuoy/app/kubeconfig.go
liztio: probably easier just to compare to empty string.

cmd/sonobuoy/app/master.go
timothysc: logical separator from the line below.

cmd/sonobuoy/app/kubeconfig.go
```diff
@@ -37,6 +37,14 @@ func (c *Kubeconfig) Set(str string) error {
 
 // Get returns a rest Config, possibly based on a provided config
 func (c *Kubeconfig) Get() (*rest.Config, error) {
+
+	// detect if in-cluster configuration makes sense
```
chuckha: This is already handled by https://github.com/heptio/sonobuoy/blob/master/cmd/sonobuoy/app/kubeconfig.go#L46

That call to ClientConfig will return in cluster config if it needs to.

PR #287, opened 2018-02-27
Fix log streaming

### LINKED ISSUE #284: Sonobuoy logs not showing plugin containers.
I can access the logs through kubectl but I I get an error on any non-aggregator pods : 

`ERRO[0000] error attempting to get sonobuoy logs: failed to stream logs: could not copy request body: <nil>`

chuckha:
Closes #284

### DIFF
--- a/pkg/client/logs.go
+++ b/pkg/client/logs.go
@@ -69,6 +69,9 @@ func streamLogs(client kubernetes.Interface, namespace, podName string, logOptio
 		return fmt.Errorf("could not stream the request: %v", err)
 	}
 	defer readCloser.Close()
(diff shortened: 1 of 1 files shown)


liztio:
This looks good, but how does adding more error logging "fix" anything?

chuckha:
This function was originally only for following logs so if io.Copy ever *finished* there must have been an error. It became reused to get all logs for all containers that sonobuoy has created. When the reuse happened, this function is expected to return and not hang forever on io.Copy but it only ever returned an error even when there was no error.

This now returns nil when nothing has gone wrong, or returns an error when something has gone wrong.

liztio:
Ahh, that makes sense!

PR #289, opened 2018-02-27
Preflight check for a DNS pod

### LINKED ISSUE #273: Add detection and pre-flight checks to run
Prior to running we should detect api-enabled (rbac) and preflight (fail-check) on known bad deployments 

// TODO(timothysc) Need to add checks which include (detection-rbac, preflight-DNS, ...)

liztio:
Closes #273

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -51,6 +51,8 @@ type E2EConfig struct {
 // RunConfig are the input options for running Sonobuoy.
 type RunConfig struct {
 	GenConfig
+	// SkipPreflight means don't run any checks before kicking off the Sonobuoy run.
+	SkipPreflight bool
 }
 
 // RetrieveConfig are the options passed to RetrieveResults.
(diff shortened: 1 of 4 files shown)


pkg/client/run.go
```diff
@@ -23,23 +23,34 @@ import (
...
+		}
+		if err := preflightCheck(client); err != nil {
+			return errors.Wrap(err, "preflight check failed")
+		}
+	}
+
 	manifest, err := c.GenerateManifest(&cfg.GenConfig)
 	if err != nil {
-		return errors.Wrap(err, "couldn't run invalid manifest")
+		return errors.Wrap(err, "")
```
chuckha: return err at this point instead of wrapping i think

pkg/client/interfaces.go
```diff
@@ -51,6 +51,7 @@ type E2EConfig struct {
 // RunConfig are the input options for running Sonobuoy.
 type RunConfig struct {
 	GenConfig
+	SkipPreflight bool
```
chuckha: godoc please!

PR #292, opened 2018-02-27
Fix default resultsdir for aggregator.

timothysc:


timothysc:
I'll log an issue around the defaults, b/c we need to const a bunch of things.

PR #295, opened 2018-02-27
Delete cmd

liztio:
```
$ sonobuoy run --mode quick
INFO[0000] created object                                name=sonobuoy-serviceaccount namespace= resource=clusterrolebindings
INFO[0000] created object                                name=sonobuoy-serviceaccount namespace= resource=clusterroles
INFO[0000] created object                                name=sonobuoy-config-cm namespace=heptio-sonobuoy resource=configmaps
INFO[0000] created object                                name=sonobuoy-plugins-cm namespace=heptio-sonobuoy resource=configmaps
INFO[0000] created object                                name=sonobuoy namespace=heptio-sonobuoy resource=pods
INFO[0000] created object                                name=sonobuoy-master namespace=heptio-sonobuoy resource=services
$ sonobuoy delete
$ k apply -f /tmp/file.yaml
namespace "e2e-tests-container-probe-fdqfr" configured
$ sonobuoy delete
INFO[0000] deleted namespace                             namespace=heptio-sonobuoy
INFO[0000] deleted clusterrolebindings and clusterroles
INFO[0000] deleted E2E namespace                         namespace=e2e-tests-container-probe-fdqfr
INFO[0000] deleted E2E namespace                         namespace=e2e-tests-pods-qlpdc
$
```

### DIFF
--- a/pkg/client/delete.go
+++ b/pkg/client/delete.go
@@ -17,10 +17,65 @@ limitations under the License.
 package client
 
 import (
+	"strings"
+
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	clusterRoleFieldName  = "component"
+	clusterRoleFieldValue = "sonobuoy"
+
+	e2eNamespacePrefix = "e2e-"
+)
+
 func (c *SonobuoyClient) Delete(cfg *DeleteConfig, client kubernetes.Interface) error {
-	return errors.New("not yet implemented")
+	// Delete the namespace
+	if err := client.CoreV1().Namespaces().Delete(cfg.Namespace, &metav1.DeleteOptions{}); err != nil {
+		return errors.Wrap(err, "failed to delete namespace")
+	}
+	logrus.WithField("namespace", cfg.Namespace).Info("deleted namespace")
+
+	if cfg.EnableRBAC {
+		// ClusterRole and ClusterRoleBindings aren't namespaced, so delete them seperately
+		selector := metav1.AddLabelToSelector(
+			&metav1.LabelSelector{},
+			clusterRoleFieldName,
+			clusterRoleFieldValue,
+		)
+
+		deleteOpts := &metav1.DeleteOptions{}
+		listOpts := metav1.ListOptions{
+			LabelSelector: metav1.FormatLabelSelector(selector),
+		}
+
+		if err := client.RbacV1().ClusterRoleBindings().DeleteCollection(deleteOpts, listOpts); err != nil {
+			return errors.Wrap(err, "failed to delete cluster role binding")
+		}
+
+		// ClusterRole and ClusterRole bindings aren't namespaced, so delete them manually
+		if err := client.RbacV1().ClusterRoles().DeleteCollection(deleteOpts, listOpts); err != nil {
+			return errors.Wrap(err, "failed to delete cluster role")
+		}
+		logrus.Info("deleted clusterrolebindings and clusterroles")
+	}
+
+	// Delete any dangling E2E namespaces
+	namespaces, err := client.CoreV1().Namespaces().List(metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to list namespaces")
+	}
+
+	for _, namespace := range namespaces.Items {
+		if strings.HasPrefix(e2eNamespacePrefix, namespace.Name) {
+			if err := client.CoreV1().Namespaces().Delete(namespace.Name, &metav1.DeleteOptions{}); err != nil {
+				return errors.Wrap(err, "failed to delete 2e2 namespace")
+			}
+			logrus.WithField("namespace", namespace.Name).Info("deleted E2E namespace")
+		}
+	}
+	return nil
 }
--- a/pkg/client/delete.go
+++ b/pkg/client/delete.go
@@ -1,0 +1,26 @@
+/*
+Copyright 2018 Heptio Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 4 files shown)


pkg/client/delete.go
```diff
@@ -0,0 +1,81 @@
...
+			clusterRoleFieldName,
+			clusterRoleFieldValue,
+		)
+
+		deleteOpts := &metav1.DeleteOptions{}
+		listOpts := metav1.ListOptions{
+			LabelSelector: metav1.FormatLabelSelector(selector),
+		}
+
+		if err := client.RbacV1().ClusterRoleBindings().DeleteCollection(deleteOpts, listOpts); err != nil {
```
timothysc: what happens if you fail a delete and retry?

liztio: aiui DeleteCollection() will delete zero or more, but you're right that the explict delete at the beginning will cause everything to bail. What do you think we should do in that case?

timothysc: Try to fire and log all the issues as we roll through it, and either return a true/false or a list of errors.

liztio:
@timothysc 

```
$ sonobuoy delete
INFO[0000] already deleted                               kind=namespace namespace=heptio-sonobuoy
ERRO[0000] failed to delete sonobuoy resources: couldn't delete namespace: namespaces "heptio-sonobuoy" not found
zsh: exit 1     sonobuoy delete
$ go install
go install  4.34s user 0.91s system 121% cpu 4.326 total
$ sonobuoy delete
INFO[0000] already deleted                               kind=namespace namespace=heptio-sonobuoy
INFO[0000] deleted                                       kind=clusterrolebindings
INFO[0000] deleted                                       kind=clusterroles
$ k apply -f /tmp/file.yaml
namespace "e2e-tests-container-probe-fdqfr" created
$ sonobuoy delete
INFO[0000] already deleted                               kind=namespace namespace=heptio-sonobuoy
INFO[0000] deleted                                       kind=clusterrolebindings
INFO[0000] deleted                                       kind=clusterroles
INFO[0000] deleted                                       kind=namespace namespace=e2e-tests-container-probe-fdqfr
$ sonobuoy delete
INFO[0000] already deleted                               kind=namespace namespace=heptio-sonobuoy
INFO[0000] deleted                                       kind=clusterrolebindings
INFO[0000] deleted                                       kind=clusterroles
INFO[0000] delete in progress                            error="Operation cannot be fulfilled on namespaces \"e2e-tests-container-probe-fdqfr\": The system is ensuring all content is removed from this namespace.  Upon completion, this namespace will automatically be purged by the system." kind=namespace namespace=e2e-tests-container-probe-fdqfr
```

pkg/client/delete.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		// ClusterRole and ClusterRole bindings aren't namespaced, so delete them manually
+		err = client.RbacV1().ClusterRoles().DeleteCollection(deleteOpts, listOpts)
+		if err := logDelete(logrus.WithField("kind", "clusterroles"), err); err != nil {
+			return errors.Wrap(err, "failed to delete cluster role")
+		}
+	}
+
+	// Delete any dangling E2E namespaces
```
chuckha: I don't think we should do this, this is the responsibility of ginko. I plumbed through the signals we needed to make ginko clean up correctly and if it isn't, that indicates a problem somewhere. This would hide that symptom from us.

pkg/client/delete.go
```diff
@@ -0,0 +1,107 @@
...
+	case err == nil:
+		log.Info("deleted")
+	case kubeerror.IsNotFound(err):
+		log.Info("already deleted")
+	case kubeerror.IsConflict(err):
+		log.WithError(err).Info("delete in progress")
+	case err != nil:
+		return err
+	}
+	return nil
```
chuckha: could this be the default case inside the switch statement? i guess it doesn't really matter

pkg/client/delete.go
```diff
@@ -0,0 +1,107 @@
...
+			err := client.CoreV1().Namespaces().Delete(namespace.Name, &metav1.DeleteOptions{})
+			if err := logDelete(log, err); err != nil {
+				return errors.Wrap(err, "couldn't delete namespace")
+			}
+		}
+	}
+	return nil
+}
+
+func logDelete(log logrus.FieldLogger, err error) error {
```
chuckha: i'm not super stoked on passing a logrus.FieldLogger in here...I see why you did it like this, but it's not my favorite.

This function feels like it's doing too much maybe, it's both logging and handling the kube error. Maybe the logging could happen after the delete takes place, something like:

```
if err := checkError(deleteNamespace()); err != nil {
    return err
}
logrus.Info("deleted namespace")
```

and then we don't have to worry about passing the log around? Another argument for this is that I don't think the user cares if it's already deleted vs deleted vs delete in progress.

pkg/client/delete.go
timothysc: The data would still be in the tarball, so I don't think it's hidden.  But it does raise a question of ownership, e.g. what happens if someone creates their own test jiggery based on upstream, we'd be destroying it.  Perhaps we can add a --all flag for this...?

timothysc: `sb delete --all` ~= yolo / dust off nuke it from orbit!

chuckha: So long as deleting e2e namespaces is not the default I am not picky about how this gets restructured.

pkg/client/delete.go
liztio: The default case is actually return err, this return is for all the cases that don't return anything.

pkg/client/delete.go
liztio: I guess it feels normal to me, because in other projects I've worked on the logger is _always_ the first argument you use. 

What you proposed is not equivalent, because then we'd be missing the `kind=namespace namespace=heptio-test`

chuckha: It would be straightforward to add that extra data to the logrus.Info line as it would all be in scope.

I think we should talk about or logging strategy in sonobuoy. We likely used the standard logger for expedience (I was hoping this project could get away with the standard logger) but maybe it's time to formalize it more and define a logger to use. 

I still think this function is doing both error checking and logging and should be broken up, but it's ok to move forward. I need to get the demo to working today.

PR #297, opened 2018-02-27
Remove logrus hooks at the end of the run.

chuckha:
Sonobuoy has captured everything interesting and has created the results.
No more logs should be written to disk at this point.

We are stuck referencing the standard logger as that's what we've been using across the rest of sonobuoy.


chuckha:
Yes, this was tested:

```
test salazar:sonobuoy cha$ go run main.go retrieve tmp
tar: removing leading '/' from member names
test salazar:sonobuoy cha$ ls tmp/
201802272122_sonobuoy_6a64e409-8771-4f32-80db-bbe28cf1c034.tar.gz
```

PR #298, opened 2018-02-27
More friendly status messages

liztio:
```
$ sonobuoy status
PLUGIN		NODE				STATUS
e2e						running
systemd_logs	ip-10-0-11-88.ec2.internal	complete
systemd_logs	ip-10-0-12-235.ec2.internal	complete
systemd_logs	ip-10-0-19-40.ec2.internal	complete

Sonobuoy is still running. Runs can take up to 60 minutes.
```

### DIFF
--- a/cmd/sonobuoy/app/status.go
+++ b/cmd/sonobuoy/app/status.go
@@ -79,10 +79,10 @@ func getStatus(cmd *cobra.Command, args []string) {
 		errlog.LogError(errors.Wrap(err, "couldn't write status out"))
 		os.Exit(1)
 	}
-	fmt.Printf("\n%s\n", statusAsString(status.Status))
+	fmt.Printf("\n%s\n", humanReadableStatus(status.Status))
 }
 
-func statusAsString(str string) string {
+func humanReadableStatus(str string) string {
 	switch str {
 	case "running":
 		return "Sonobuoy is still running. Runs can take up to 60 minutes."
--- a/cmd/sonobuoy/app/status.go
+++ b/cmd/sonobuoy/app/status.go
@@ -19,7 +19,6 @@ package app
 import (
 	"fmt"
 	"os"
-	"strings"
 	"text/tabwriter"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


cmd/sonobuoy/app/status.go
```diff
@@ -75,9 +74,23 @@ func getStatus(cmd *cobra.Command, args []string) {
...
 		errlog.LogError(errors.Wrap(err, "couldn't write status out"))
 		os.Exit(1)
 	}
+	fmt.Printf("\n%s\n", statusAsString(status.Status))
+}
+
+func statusAsString(str string) string {
+	switch str {
+	case "running":
+		return "Sonobuoy is still running. Runs can take up to 60 minutes."
```
timothysc: Only if it's conformance or extended does it take that long.

liztio: @timothysc we haven't rolled anything up yet

liztio:
@timothysc we haven't rolled anything up yet, this just adds the friendly status messages.

cmd/sonobuoy/app/status.go
```diff
@@ -75,9 +74,23 @@ func getStatus(cmd *cobra.Command, args []string) {
...
-	fmt.Fprintf(tw, "\t\t%s\n", strings.ToUpper(status.Status))
+
 	if err := tw.Flush(); err != nil {
 		errlog.LogError(errors.Wrap(err, "couldn't write status out"))
 		os.Exit(1)
 	}
+	fmt.Printf("\n%s\n", statusAsString(status.Status))
+}
+
+func statusAsString(str string) string {
```
chuckha: Why isn't this on the Status struct? Could be the String() string method even

liztio: The status struct has a bunch of information on it, not just the overall status. So having it as String() doesn't make much sense I think? But I could move it there for sure.

timothysc:
I'm ok with just merging this and we can make more modifications l8r.

cmd/sonobuoy/app/status.go
chuckha: I think you're right, status is the status of a plugin not the overall status. Maybe this could be named `overallStatusMessage` or something like that? I think it's the `AsString` bit that I'm not a fan of since it is redundant with the function signature.

liztio: how about "humanReadableStatus"?

PR #302, opened 2018-02-27
Implements e2e subcommand

### LINKED ISSUE #248: Creation of an e2e sub-command and tarball interface layer
## Problem statement 

We need an interface library above the tarball interface that provides some base level aggregate functions for getting summary information from e2e style runs, across multiple e2e-style-plugins.  

The abstraction layer should provide very basic rolled up data at 1st and we can refine the requirements over time.  

For example, a common case would be to roll up passing and failing, with a single boolean variable.  If the boolean is false a list of the failed tests are provided for automation.  

A string list that can be `regex or'd |` together and could be passed back into run via a focus varible to execute just those failed tests.   

I'd argue that this is a degenerative use case and should generally be discouraged, but hey "stuff happens" and this is hard. 

Prior to beginning this work, we should create a SEP that outlines some of these details and references this issue. 

/cc @fabioy @k8k

chuckha:
This subcommand allows users to more easily inspect their
results archive. They can see which tests passed, failed or
were skipped. This also allows users to rerun the tests that
failed.

Closes #248

### DIFF
--- a/cmd/sonobuoy/app/e2e.go
+++ b/cmd/sonobuoy/app/e2e.go
@@ -1,0 +1,118 @@
+/*
+Copyright 2018 Heptio Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package app
+
+import (
+	"compress/gzip"
+	"fmt"
+	"os"
+
+	"github.com/heptio/sonobuoy/pkg/client"
+	"github.com/heptio/sonobuoy/pkg/errlog"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/kubernetes"
+)
+
+var e2eFlags struct {
+	show  string
+	rerun bool
+}
+
+func init() {
+	cmd := &cobra.Command{
+		Use:   "e2e archive.tar.gz",
+		Short: "Inspect e2e test results. Optionally rerun failed tests",
+		Run:   e2es,
+		Args:  cobra.ExactArgs(1),
+	}
+	AddGenFlags(&runopts.GenConfig, cmd)
+	AddModeFlag(&runFlags.mode, cmd)
+	AddSonobuoyConfigFlag(&runFlags.sonobuoyConfig, cmd)
+	AddKubeconfigFlag(&runFlags.kubecfg, cmd)
+	// Default to detect since we need a kubeconfig regardless
+	AddRBACModeFlags(&runFlags.rbacMode, cmd, DetectRBACMode)
+	AddSkipPreflightFlag(&runopts.SkipPreflight, cmd)
+
+	cmd.PersistentFlags().StringVar(&e2eFlags.show, "show", "failed", "Defines which tests to show, options are [passed, failed (default) or all]. Cannot be combined with --rerun-failed.")
+	cmd.PersistentFlags().BoolVar(&e2eFlags.rerun, "rerun-failed", false, "Rerun the failed tests reported by the archive. The --show flag will be ignored.")
+
+	RootCmd.AddCommand(cmd)
+}
+
+func e2es(cmd *cobra.Command, args []string) {
+	f, err := os.Open(args[0])
+	if err != nil {
+		errlog.LogError(errors.Wrapf(err, "could not open sonobuoy archive: %v", args[0]))
+		os.Exit(1)
+	}
+	defer f.Close()
+	// As documented, ignore show if we are doing a rerun of failed tests.
+	if e2eFlags.rerun {
+		e2eFlags.show = "failed"
+	}
+	gzr, err := gzip.NewReader(f)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "could not make a gzip reader"))
+		os.Exit(1)
+	}
+	defer gzr.Close()
+	sonobuoy := client.NewSonobuoyClient()
+	testCases, err := sonobuoy.GetTests(gzr, e2eFlags.show)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "could not get tests from archive"))
+		os.Exit(1)
+	}
+
+	// If we are not doing a rerun, print and exit.
+	if !e2eFlags.rerun {
+		fmt.Printf("%v tests\n", e2eFlags.show)
+		fmt.Println(client.PrintableTestCases(testCases))
+		return
+	}
+
+	restConfig, err := runFlags.kubecfg.Get()
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "couldn't get REST client"))
+		os.Exit(1)
+	}
+
+	clientset, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "couldn't create Kubernetes clientset"))
+		os.Exit(1)
+	}
+
+	rbacEnabled, err := runFlags.rbacMode.Enabled(clientset)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "couldn't detect RBAC mode"))
+		os.Exit(1)
+	}
+	runopts.GenConfig.EnableRBAC = rbacEnabled
+
+	runopts.Config = GetConfigWithMode(&runFlags.sonobuoyConfig, runFlags.mode)
+	runopts.E2EConfig = &client.E2EConfig{
+		Focus: client.Focus(testCases),
+		Skip:  "",
+	}
+
+	fmt.Printf("Rerunning %d tests:\n", len(testCases))
+	if err := sonobuoy.Run(&runopts, restConfig); err != nil {
+		errlog.LogError(errors.Wrap(err, "error attempting to rerun failed tests"))
+		os.Exit(1)
+	}
+}
--- a/pkg/client/e2e.go
+++ b/pkg/client/e2e.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2018 Heptio Inc.
+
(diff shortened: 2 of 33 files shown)


pkg/client/interfaces.go
```diff
@@ -91,4 +91,6 @@ type Interface interface {
 	GetStatus(namespace string, client kubernetes.Interface) (*aggregation.Status, error)
 	// GetLogs streams logs from the sonobuoy pod by default to stdout.
 	GetLogs(cfg *LogConfig, client kubernetes.Interface) error
+	// GetTests returns a collection of tests extracted from a sonobuoy archive.
+	GetTests(reader io.Reader, show string) (TestCases, error)
```
chuckha: idk how i feel about this being part of the interface. Seems like we were on the fence including it in the CLI anyway, does it make sense to be part of the SB interface or is this a special case?

pkg/client/e2e.go
```diff
@@ -0,0 +1,119 @@
...
+	r := strings.NewReplacer("[", `\\[`, "]", `\\]`)
+	testNames := make([]string, len(t))
+	for i, tc := range t {
+		testNames[i] = r.Replace(tc.Name)
+	}
+	return strings.Join(testNames, "|")
+}
+
+// TestCase is a single junit test result.
+type TestCase struct {
```
chuckha: These are defined in ginko as well, I could pull in their structs but didn't want to include ginko as a dep.

cmd/sonobuoy/app/e2e.go
```diff
@@ -0,0 +1,100 @@
...
+		Short: "inspect e2e results",
+		Run:   e2es,
+		Args:  cobra.ExactArgs(1),
+	}
+	RootCmd.AddCommand(cmd)
+	AddKubeconfigFlag(&runFlags.kubecfg, cmd)
+	cmd.PersistentFlags().StringVar(&show, "show", "failed", "Defines which tests to show, options are [passed, failed (default) or all]. Cannot be combined with --rerun-failed.")
+	cmd.PersistentFlags().BoolVar(&rerunFailed, "rerun-failed", false, "Rerun the failed tests reported by the archive. The --show flag will be ignored.")
+}
+
```
timothysc: One thing that isn't clear to me and we should probably outline in a comment block is the extension model for the cli & plugins.  e2e is a special case right now but in the future we should make it pluggable.

cmd/sonobuoy/app/e2e.go
```diff
@@ -0,0 +1,100 @@
...
+		os.Exit(0)
+	}
+
+	// Create the necessary configs for rerunning.
+	runOptions := &client.RunConfig{
+		GenConfig: client.GenConfig{
+			E2EConfig: &client.E2EConfig{
+				Focus: testCases.Focus(),
+			},
+			// Quick with a custom Focus config will work.
```
timothysc: you shouldn't need any of the options below, they should default.   If not, we should fix that.

cmd/sonobuoy/app/e2e.go
```diff
@@ -0,0 +1,100 @@
...
+	cmd := &cobra.Command{
+		Use:   "e2e",
+		Short: "inspect e2e results",
+		Run:   e2es,
+		Args:  cobra.ExactArgs(1),
+	}
+	RootCmd.AddCommand(cmd)
+	AddKubeconfigFlag(&runFlags.kubecfg, cmd)
+	cmd.PersistentFlags().StringVar(&show, "show", "failed", "Defines which tests to show, options are [passed, failed (default) or all]. Cannot be combined with --rerun-failed.")
+	cmd.PersistentFlags().BoolVar(&rerunFailed, "rerun-failed", false, "Rerun the failed tests reported by the archive. The --show flag will be ignored.")
```
timothysc: This gets weird with extended tests (sonobuoy-plugins)

cmd/sonobuoy/app/e2e.go
```diff
@@ -0,0 +1,100 @@
...
+	testCases, err := sonobuoy.GetTests(gzr, show)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "could not get tests from archive"))
+		os.Exit(1)
+	}
+
+	// If we are not doing a rerun, print and exit.
+	if !rerunFailed {
+		fmt.Print(testCases)
+		os.Exit(0)
```
liztio: I think we've been using `return` for the exit(0) case.

pkg/client/e2e.go
```diff
@@ -0,0 +1,119 @@
...
+	// Since it's not and it's a subchild that has no elements or contents, the best we
+	// can do is see if it exists. If it exists, we make a struct. If it's not it is nil.
+	Skipped   *struct{} `xml:"skipped"`
+	Failure   string    `xml:"failure"`
+	SystemOut string    `xml:"system-out"`
+}
+
+// jUnitResults is a collection of test cases and metadata.
+type jUnitResults struct {
+	XMLName   xml.Name  `xml:"testsuite" json:"-"`
```
liztio: why does just this one have a JSON attribute?

pkg/client/e2e.go
```diff
@@ -0,0 +1,119 @@
...
+
+// GetTests extracts the junit results from a sonobuoy archive and returns the requested tests.
+func (c *SonobuoyClient) GetTests(reader io.Reader, show string) (TestCases, error) {
+	read := results.NewReaderWithVersion(reader, "irrelevant")
+	junitResults := jUnitResults{}
+	err := read.WalkFiles(func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		return results.ExtractFileIntoStruct("plugins/e2e/results/junit_01.xml", path, info, &junitResults)
```
timothysc: I think we may want to add a //TODO here about extensibility for things like the extended tests/sonobuoy-plugins

pkg/client/e2e.go
```diff
@@ -0,0 +1,119 @@
...
+		out = append(out, junitResults.filter(failed)...)
+	}
+	if show == "skipped" || show == "all" {
+		out = append(out, junitResults.filter(skipped)...)
+	}
+	sort.Sort(out)
+	return out, nil
+}
+
+type TestCases []TestCase
```
timothysc: This almost seems like this belongs with the tarball utility library, and not really part of the external client.

pkg/client/e2e.go
timothysc: https://github.com/onsi/ginkgo/blob/master/reporters/junit_reporter.go seems pretty bounded, I'd move this code into the tarball utils and yank it in.

pkg/client/interfaces.go
timothysc: Put a //TODO around it b/c long term it should be it's own thing.

cmd/sonobuoy/app/e2e.go
chuckha: Weird in what way?

pkg/client/e2e.go
chuckha: I'll open a ticket to capture the idea of extending this command to work with our extended tests.

cmd/sonobuoy/app/e2e.go
chuckha: Pluggable meaning other people can write sonobuoy command line extensions for their own plugins?

pkg/client/e2e.go
chuckha: https://github.com/heptio/sonobuoy/issues/306

cmd/sonobuoy/app/e2e.go
```diff
@@ -0,0 +1,118 @@
...
+func init() {
+	cmd := &cobra.Command{
+		Use:   "e2e archive.tar.gz",
+		Short: "Inspect e2e test results. Optionally rerun failed tests",
+		Run:   e2es,
+		Args:  cobra.ExactArgs(1),
+	}
+	AddGenFlags(&runopts.GenConfig, cmd)
+	AddModeFlag(&runFlags.mode, cmd)
+	AddSonobuoyConfigFlag(&runFlags.sonobuoyConfig, cmd)
```
liztio: could you make another local struct for this? Using runFlags is fine by me, but I think show and rerun failed should go in there too (since the namespace is still shared)

cmd/sonobuoy/app/e2e.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// If we are not doing a rerun, print and exit.
+	if !rerunFailed {
+		fmt.Printf("%v tests\n", show)
+		fmt.Println(client.PrintableTestCases(testCases))
+		return
+	}
+
+	restConfig, err := runFlags.kubecfg.Get()
```
liztio: Might be worth breaking this out into its own function

pkg/client/e2e.go
```diff
@@ -0,0 +1,80 @@
...
+// GetTests extracts the junit results from a sonobuoy archive and returns the requested tests.
+func (c *SonobuoyClient) GetTests(reader io.Reader, show string) ([]reporters.JUnitTestCase, error) {
+	read := results.NewReaderWithVersion(reader, "irrelevant")
+	junitResults := reporters.JUnitTestSuite{}
+	err := read.WalkFiles(func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		// TODO(chuckha) consider reusing this function for any generic e2e-esque plugin results.
+		return results.ExtractFileIntoStruct(results.PluginsDir+e2e.ResultsSubdirectory+e2e.JUnitResultsFile, path, info, &junitResults)
```
liztio: could you use paths.Join() here?

cmd/sonobuoy/app/e2e.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "couldn't get REST client"))
+		os.Exit(1)
+	}
+
+	clientset, err := kubernetes.NewForConfig(restConfig)
```
liztio: we don't need to do it now, but I'll bet there's a lot of code duplicated from Run() here. Maybe a TODO to break it out into a helper function?

pkg/client/e2e.go
chuckha: I'll make a TODO because this will require some other refactoring and since sonobuoy is in control of writing the tarball, it doesn't seem too likely that it will bite us (famous last words).

cmd/sonobuoy/app/e2e.go
chuckha: https://github.com/heptio/sonobuoy/issues/312

PR #303, opened 2018-02-28
Hardcodes version so `go get` has some state.

timothysc:

/cc @liztio


liztio:
:+1:

PR #311, opened 2018-03-01
Use constants for `sonobuoy status` to avoid completed errors

liztio:
@timothysc I can't remember if we're trying to do any dep limiting in cmd, but I think `sonobuoy master` is gonna pull in aggregation anyway.


liztio:
```
$ sonobuoy status
PLUGIN	NODE	STATUS
e2e		complete

Sonobuoy has completed. Use `sonobuoy retrieve` to get results.
```

PR #315, opened 2018-03-02
Refactor flag handling

liztio:
This should make it easier to add flags to commands that "depend" on other commands

### DIFF
--- a/cmd/sonobuoy/app/gen.go
+++ b/cmd/sonobuoy/app/gen.go
@@ -35,36 +35,39 @@ type genFlags struct {
 	rbacMode       RBACMode
 	kubecfg        Kubeconfig
 	e2eflags       *pflag.FlagSet
+	namespace      string
+	sonobuoyImage  string
 }
 
-var genopts ops.GenConfig
 var genflags genFlags
 
-func (g *genFlags) AddFlags(flags *pflag.FlagSet, cfg *ops.GenConfig, rbac RBACMode) {
+func GenFlagSet(cfg *genFlags, rbac RBACMode) *pflag.FlagSet {
 	genset := pflag.NewFlagSet("generate", pflag.ExitOnError)
-	AddModeFlag(&g.mode, genset)
-	AddSonobuoyConfigFlag(&g.sonobuoyConfig, genset)
-	AddKubeconfigFlag(&g.kubecfg, genset)
-	g.e2eflags = AddE2EConfigFlags(genset)
-	AddRBACModeFlags(&g.rbacMode, genset, rbac)
-	flags.AddFlagSet(genset)
+	AddModeFlag(&cfg.mode, genset)
+	AddSonobuoyConfigFlag(&cfg.sonobuoyConfig, genset)
+	AddKubeconfigFlag(&cfg.kubecfg, genset)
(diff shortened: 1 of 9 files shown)


cmd/sonobuoy/app/gen.go
```diff
@@ -22,21 +22,50 @@ import (
...
+	AddKubeconfigFlag(&g.kubecfg, genset)
+	AddE2EConfigFlags(genset)
+	AddRBACModeFlags(&g.rbacMode, genset, rbac)
+	flags.AddFlagSet(genset)
+
+	AddNamespaceFlag(&cfg.Namespace, flags)
+	AddSonobuoyImage(&cfg.Image, flags)
+}
+
+func (g *genFlags) FillConfig(cmd *cobra.Command, cfg *ops.GenConfig) error {
```
chuckha: would it work if this returned a `*ops.GenConfig` directly instead of relying on a global var that we pass in?

Maybe there's a way to do this without having to pass in `*cobra.Command` -- it looks like that's only used for the e2e flags. Those look like special cases, do they need to be?

cmd/sonobuoy/app/gen.go
```diff
@@ -22,21 +22,50 @@ import (
...
 	sonobuoyConfig SonobuoyConfig
 	mode           ops.Mode
 	rbacMode       RBACMode
 	kubecfg        Kubeconfig
 }
 
+var genopts ops.GenConfig
+var genflags genFlags
+
+func (g *genFlags) AddFlags(flags *pflag.FlagSet, cfg *ops.GenConfig, rbac RBACMode) {
```
chuckha: What do you think about a slight restructure where this returned a flagset? Then we could have something like:

```
cmd.Flags().AddFlagSet(GenFlagSet(&genopts))
```

This could give us a nice set of functions (GenFlagSet, RunFlagSet, etc) that returned configured flagsets for each command (and also be able to build on top of each other).

RunFlagSet might be like
```
runFlagSet := new(pflag.FlagSet)
runFlagSet.AddFlagSet(GenFlagSet(&genOptions))
runFlagSet.AddFlag(runflag1 ...)
runFlagSet.AddFlag(runflag2 ...)
return runFlagSet
```

I think the the solution you've provided is a solid way to go and a huge improvement, but I find it a little difficult to follow mutating pointers around. I've found it's best to avoid pointers as arguments if it can be avoided.

cmd/sonobuoy/app/gen.go
liztio: > would it work if this returned a *ops.GenConfig directly instead of relying on a global var that we pass in?

It's passed in so it works with embedded structs

> Maybe there's a way to do this without having to pass in *cobra.Command -- it looks like that's only used for the e2e flags. Those look like special cases, do they need to be?
They do :( We need access to `Changed()` and we have to get that through `cmd`

cmd/sonobuoy/app/gen.go
liztio: I can give that a try, I think that looks good!

cmd/sonobuoy/app/gen.go
liztio: Okay, I'm gonna see if I can work around the latter case.

PR #320, opened 2018-03-05
Default Configs for client interfaces

### LINKED ISSUE #286: GenConfig has no defaults
There are no defaults on GenConfig. It would be nice to be able to generate sonobuoy YAML without having to know anything.

Reproduce:

```
	generateOptions := &client.GenConfig{
		E2EConfig: &client.E2EConfig{
			Focus: testCases.Focus(),
		},
		Config: GetConfigWithMode(&SonobuoyConfig{}, client.Quick),
	}

	bytes, err := client.NewSonobuoyClient().GenerateManifest(generateOptions)
```
=> 
```
spec:
  containers:
  - command:
    - /bin/bash
    - -c
    - /sonobuoy master --no-exit=true -v 3 --logtostderr
    image:  # problem is here, it is empty but it should be :latest or :master of sonobuoy
    imagePullPolicy: Always
```

liztio:
Closes #286

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -77,12 +77,14 @@ const (
 // AddE2EConfigFlags adds two arguments: --e2e-focus and --e2e-skip. These are not taken as pointers, as they are only used by GetE2EConfig. Instead, they are returned as a Flagset which should be passed to GetE2EConfig. The returned flagset will be added to the passed in flag set.
 func AddE2EConfigFlags(flags *pflag.FlagSet) *pflag.FlagSet {
 	e2eFlags := pflag.NewFlagSet("e2e", pflag.ExitOnError)
+	modeName := ops.Conformance
+	defaultMode := modeName.Get()
 	e2eFlags.String(
-		e2eFocusFlag, ops.ConformanceModeConfig.E2EConfig.Focus,
+		e2eFocusFlag, defaultMode.E2EConfig.Focus,
 		"Specify the E2E_FOCUS flag to the conformance tests. Overrides --mode.",
 	)
 	e2eFlags.String(
-		e2eSkipFlag, ops.ConformanceModeConfig.E2EConfig.Skip,
+		e2eSkipFlag, defaultMode.E2EConfig.Skip,
 		"Specify the E2E_SKIP flag to the conformance tests. Overrides --mode.",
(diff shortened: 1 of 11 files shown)


cmd/sonobuoy/app/args.go
```diff
@@ -76,14 +76,12 @@ const (
 // AddE2EConfigFlags adds two arguments: --e2e-focus and --e2e-skip. These are not taken as pointers, as they are only used by GetE2EConfig. Instead, they are returned as a Flagset which should be passed to GetE2EConfig. The returned flagset will be added to the passed in flag set.
 func AddE2EConfigFlags(flags *pflag.FlagSet) *pflag.FlagSet {
 	e2eFlags := pflag.NewFlagSet("e2e", pflag.ExitOnError)
-	modeName := ops.Conformance
-	defaultMode := modeName.Get()
 	e2eFlags.String(
-		e2eFocusFlag, defaultMode.E2EConfig.Focus,
+		e2eFocusFlag, ops.ConformanceModeConfig.E2EConfig.Focus,
```
timothysc: Typically referencing consts don't hit globals.

pkg/client/defaults.go
```diff
@@ -0,0 +1,62 @@
...
+limitations under the License.
+*/
+
+package client
+
+import (
+	"github.com/heptio/sonobuoy/pkg/config"
+)
+
+var ()
```
timothysc: left over ?

liztio: huh. Surprised golint doesn't catch that.

pkg/client/defaults.go
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+package client
+
+import (
+	"github.com/heptio/sonobuoy/pkg/config"
+)
+
+// NewGenConfigWithDefaults is a GenConfig using the default config and Conformance mode
+func NewGenConfigWithDefaults() *GenConfig {
```
chuckha: I think it'd be ok to drop the "WithDefaults". I'd assume NewGenConfig() would have defaults embedded in it implicitly, but I'm also fine with NewGenConfigWithDefaults.

liztio: I added the WithDefaults() to align with NewConfigWithDefaults(), but I can drop this

PR #321, opened 2018-03-05
Updates NewSonobuoyClient to take *rest.Config to prevent passing in

### LINKED ISSUE #300: Cleanup and reduce sonobuoy.interface to not pass config/interface all over.
We should modify the sonobuoy client to only use the kuberenetes.iterface as that is the normal mode of operation, if we need lower level structs there are mutation functions we should use.  

This makes it much more consistent with other tools, and a requirement for integration.

timothysc:
Fixes #300 

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -73,11 +73,25 @@ type RetrieveConfig struct {
 }
 
 // SonobuoyClient is a high-level interface to Sonobuoy operations.
-type SonobuoyClient struct{}
+type SonobuoyClient struct {
+	RestConfig *rest.Config
+	Client     kubernetes.Interface
+	// TODO (timothysc) cache dynamic client
+}
 
 // NewSonobuoyClient creates a new SonobuoyClient
-func NewSonobuoyClient() *SonobuoyClient {
-	return &SonobuoyClient{}
+func NewSonobuoyClient(restConfig *rest.Config) (*SonobuoyClient, error) {
+	sc := &SonobuoyClient{
+		RestConfig: restConfig,
+	}
+	// TODO (timothysc) factor this out and provide a means to perform
+	// JIT creation of the client through an accessor function
+	clientset, err := kubernetes.NewForConfig(sc.RestConfig)
+	if err != nil {
+		return nil, err
+	}
+	sc.Client = clientset
+	return sc, nil
 }
 
 // Make sure SonobuoyClient implements the interface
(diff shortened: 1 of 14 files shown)


pkg/client/interfaces.go
```diff
@@ -73,11 +73,22 @@ type RetrieveConfig struct {
...
+}
 
 // NewSonobuoyClient creates a new SonobuoyClient
-func NewSonobuoyClient() *SonobuoyClient {
-	return &SonobuoyClient{}
+func NewSonobuoyClient(restConfig *rest.Config) (*SonobuoyClient, error) {
+	sc := &SonobuoyClient{
+		RestConfig: restConfig,
+	}
+	clientset, err := kubernetes.NewForConfig(sc.RestConfig)
```
liztio: What about the cases where we don't actually need a kube config? `sonobuoy gen` shouldn't require it

pkg/client/interfaces.go
```diff
@@ -89,15 +100,15 @@ var _ Interface = &SonobuoyClient{}
 type Interface interface {
 	// Run generates the manifest, then tries to apply it to the cluster.
 	// returns created resources or an error
-	Run(cfg *RunConfig, restConfig *rest.Config) error
+	Run(cfg *RunConfig) error
```
liztio: 👍 so nice and clean!

pkg/client/interfaces.go
liztio: I think it's weird that creating the dynamic client fails at call time, but creating a kubectl client fails at creation time

timothysc: > What about the cases where we don't actually need a kube config? sonobuoy gen shouldn't require it 

Gen should really be doing dynamic RBAC detection, so although it's not strictly needed right this instant, I think it's a better practice. 

> I think it's weird that creating the dynamic client fails at call time, but creating a kubectl client fails at creation time 

We could cache the dynamic client here too.  I don't have strong feeling on it.

liztio: I think it's good for `sonobuoy gen` to not hit the network. It's a bit like `GET` requests having side effects, I think it would be surprising.

I'm happy to cache a dynamic client as well, my point was just we're initialising a struct only half our commands end up using.

timothysc: Right now gen doesn't actually hit the network, but the whole client having a valid KubeConfig in order to function seems like a reasonable contract.  It's not strictly needed by gen (again right now), but I don't want to open up a loophole in the api to allow for inconsistency.  

I'll add a //TODO to cache the dynamic client.

PR #326, opened 2018-03-06
Rework logging code

chuckha:
This change hides the concurrency from the client to
provide a familiar interface. Clients can now read
a merged stream of logs as if it were a synchronous io.Reader.

### DIFF
--- a/cmd/sonobuoy/app/logs.go
+++ b/cmd/sonobuoy/app/logs.go
@@ -18,16 +18,21 @@ package app
 
 import (
 	"fmt"
+	"io"
 	"os"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
-	ops "github.com/heptio/sonobuoy/pkg/client"
+	"github.com/heptio/sonobuoy/pkg/client"
 	"github.com/heptio/sonobuoy/pkg/errlog"
 )
 
-var logConfig ops.LogConfig
+const (
+	bufSize = 2048
+)
+
+var logConfig client.LogConfig
 var logsKubecfg Kubeconfig
 
 func init() {
@@ -42,7 +47,7 @@ func init() {
 		&logConfig.Follow, "follow", "f", false,
 		"Specify if the logs should be streamed.",
 	)
-
+	logConfig.Out = os.Stdout
 	AddKubeconfigFlag(&logsKubecfg, cmd.Flags())
 	AddNamespaceFlag(&logConfig.Namespace, cmd.Flags())
 	RootCmd.AddCommand(cmd)
@@ -54,13 +59,26 @@ func getLogs(cmd *cobra.Command, args []string) {
 		errlog.LogError(fmt.Errorf("failed to get rest config: %v", err))
 		os.Exit(1)
 	}
-	sbc, err := ops.NewSonobuoyClient(restConfig)
+	sbc, err := client.NewSonobuoyClient(restConfig)
 	if err != nil {
 		errlog.LogError(errors.Wrap(err, "could not create sonobuoy client"))
 		os.Exit(1)
 	}
-	if err := sbc.GetLogs(&logConfig); err != nil {
-		errlog.LogError(errors.Wrap(err, "error attempting to get sonobuoy logs"))
+	logreader, err := sbc.LogReader(&logConfig)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "could not build a log reader"))
 		os.Exit(1)
+	}
+	b := make([]byte, bufSize)
+	for {
+		n, err := logreader.Read(b)
+		if err != nil && err != io.EOF {
+			errlog.LogError(errors.Wrap(err, "error reading logs"))
+			os.Exit(1)
+		}
+		fmt.Fprint(logConfig.Out, string(b[:n]))
+		if err == io.EOF {
+			return
+		}
 	}
 }
--- a/pkg/client/logs.go
+++ b/pkg/client/logs.go
@@ -19,59 +19,212 @@ package client
 import (
 	"fmt"
 	"io"
-	"os"
-	"strings"
+	"sync"
 
-	"github.com/heptio/sonobuoy/pkg/config"
-	"github.com/sirupsen/logrus"
+	"github.com/pkg/errors"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
-var (
-	podLogSeparator = strings.Repeat("-", 79)
+const (
+	bufSize = 4096
 )
 
-// Logs gathers the logs for the containers in the sonobuoy namespace and prints them
+// Reader provides an io.Reader interface to a channel of bytes.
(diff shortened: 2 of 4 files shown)


chuckha:
This is easily break-outable into two commits: 1) adds the client.Reader struct and then 2) uses it. If you'd like I can do this, I know +391 is a pretty big ask.

pkg/client/logs_test.go
```diff
@@ -0,0 +1,185 @@
+package client
```
liztio: are there any tests with the "headers?"

pkg/client/logs.go
```diff
@@ -19,59 +19,211 @@ package client
...
+		n += len(r.overflowBuffer)
+		r.overflowBuffer = []byte{}
+		// This is suboptimal (does not use entire buffer) but gets output to the user faster.
+		return n, nil
 	}
 
-	pods, err := c.Client.CoreV1().Pods(cfg.Namespace).List(metav1.ListOptions{})
+	// One message a time.
+	select {
+	// it's ok to keep reading messages and miss an error.
```
liztio: what does this mean? will we keep returning this error if it's called repeatedly?

pkg/client/logs.go
```diff
@@ -19,59 +19,211 @@ package client
...
+		close(errc)
+	}(&wg, agg, errc)
+
+	return NewReader(applyHeaders(agg), errc), nil
+}
+
+// message represents a buffer of logs from a container in a pod in a namespace.
+type message struct {
+	// header acts as the header message for the log line and an id to know where the message came from.
+	header string
```
liztio: "preamble" might be better here

pkg/client/logs_test.go
```diff
@@ -0,0 +1,185 @@
...
+			for ; ; i++ {
+				n, err := reader.Read(mybuf)
+				if err != nil && err != io.EOF {
+					t.Fatalf("Expected no errors got %v", err)
+				}
+				if err == io.EOF {
+					break
+				}
+				if n > len(mybuf) {
+					t.Fatalf("n is too big: %v mybuf is only %v", n, len(mybuf))
```
liztio: some of these could probably be `Errorf` instead of `Fatalf`. I usually only use the latter when continuing might lead to a panic (like with the eof)

cmd/sonobuoy/app/logs.go
```diff
@@ -54,13 +56,26 @@ func getLogs(cmd *cobra.Command, args []string) {
...
+	logreader, err := sbc.LogReader(&logConfig)
+	if err != nil {
+		errlog.LogError(err)
 		os.Exit(1)
 	}
+	b := make([]byte, 1024)
+	for {
+		n, err := logreader.Read(b)
+		if err != nil && err != io.EOF {
+			errlog.LogError(err)
```
liztio: do these need wrapping?

pkg/client/logs.go
chuckha: this is acknowledging that if both channels are ready, select will randomly choose either errc or bytestream. if it decides to choose bytestream we'll keep reading even though we've encountered an error somewhere, but it's ok to keep reading, eventually select will give us our error and it will get propagated as expected.

pkg/client/logs.go
```diff
@@ -19,59 +19,211 @@ package client
...
+		l.errc <- errors.Wrapf(err, "error streaming logs from container [%v]", l.container)
+		return
 	}
 	defer readCloser.Close()
-	// In the case of -f this will never return unless there is an error.
-	if _, err = io.Copy(os.Stdout, readCloser); err != nil {
-		return fmt.Errorf("could not copy request body: %v", err)
+
+	// newline because logs have new lines in them
+	header := fmt.Sprintf("## Namespace: %v ## Pod: %v ## Container: %v\n", l.ns, l.pod, l.container)
```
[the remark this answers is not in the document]
chuckha: what do you think about

`\n==> namespace.pod.container <==\n`? (a la `tail -f`)

but what were you thinking as far as structured logging? i wasn't thinking about non humans consuming this particular output

pkg/client/logs_test.go
chuckha: all the files should have the headers, i'll add it here and submit another PR for any other missing headers

pkg/client/logs_test.go
chuckha: good call, I'll switch them where appropriate

cmd/sonobuoy/app/logs.go
chuckha: probably

liztio:
I was thinking namespace=heptio pod=sonobouy container=aggregator the way
logrus works. I tend to assume all things for human consumption will
eventually wind up in a log somewhere 😅

chuckha:
haha probably a wise assumption. i'll fix that up

cmd/sonobuoy/app/logs.go
```diff
@@ -42,7 +47,8 @@ func init() {
 		&logConfig.Follow, "follow", "f", false,
 		"Specify if the logs should be streamed.",
 	)
-
+	// TODO(chuckha) make this configurable if necessary.
```
timothysc: nah, from a command line util the can pipe it anywhere.

cmd/sonobuoy/app/logs.go
```diff
@@ -54,13 +60,26 @@ func getLogs(cmd *cobra.Command, args []string) {
...
-	if err := sbc.GetLogs(&logConfig); err != nil {
-		errlog.LogError(errors.Wrap(err, "error attempting to get sonobuoy logs"))
+	logreader, err := sbc.LogReader(&logConfig)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "could not build a log reader"))
 		os.Exit(1)
 	}
+	b := make([]byte, bufferSize)
+	for {
+		n, err := logreader.Read(b)
```
timothysc: love it!

cmd/sonobuoy/app/logs.go
```diff
@@ -18,18 +18,23 @@ package app
...
+	"github.com/heptio/sonobuoy/pkg/client"
 	"github.com/heptio/sonobuoy/pkg/errlog"
 )
 
-var logConfig ops.LogConfig
+var logConfig client.LogConfig
 var logsKubecfg Kubeconfig
 
+const (
+	bufferSize = 4096
```
timothysc: why const this twice, you could export from logs and reference that one.

pkg/client/logs.go
```diff
@@ -19,59 +19,211 @@ package client
...
+	// Always return the error if it is set.
+	if r.err != nil {
+		return 0, r.err
+	}
+
+	// n is the number of bytes written to `p`.
+	n := 0
+
+	// Send any overflow before grabbing new messages.
+	if len(r.overflowBuffer) > 0 {
```
timothysc: There has gotta be a data struct for this but you could //TODO that l8r.

cmd/sonobuoy/app/logs.go
chuckha: they can be independently set, they just happen to be the same. I think exporting & referencing might cause confusion later on "why are these the same reference? do the buffers have to be the same size?"

PR #329, opened 2018-03-07
Return exit code 1 when command not found

### LINKED ISSUE #327: Return code for unknown command is incorrect
`sonobuoy cp .` returns now an error since the command is replaced by `retrieve` but the code returned is 0 (should be 1 or something else).

```
$ go get -u github.com/heptio/sonobuoy
$ sonobuoy cp .
Error: unknown command "cp" for "sonobuoy"
Run 'sonobuoy --help' for usage.
$ echo $?
0
```

Thanks

chuckha:
Closes #327

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,10 +17,16 @@ limitations under the License.
 package main
 
 import (
+	"os"
+
 	"github.com/heptio/sonobuoy/cmd/sonobuoy/app"
 )
 
 // main entry point of the program
 func main() {
-	app.RootCmd.Execute()
+	err := app.RootCmd.Execute()
+	if err != nil {
+		// Execute takes care of printing the error
+		os.Exit(1)
+	}
 }


main.go
```diff
@@ -17,10 +17,16 @@ limitations under the License.
...
+
 	"github.com/heptio/sonobuoy/cmd/sonobuoy/app"
 )
 
 // main entry point of the program
 func main() {
-	app.RootCmd.Execute()
+	err := app.RootCmd.Execute()
+	if err != nil {
+		// Execute takes care of printing the error
```
timothysc: Should we also log the error for consistency?

chuckha: We could but the user would see the same error twice unless we pass in some silencing flag to cobra.

The default output looks like this:

```
test salazar:sonobuoy cha$ go run main.go sdlkfjslkdjf
 Error: unknown command "sdlkfjslkdjf" for "sonobuoy"
Run 'sonobuoy --help' for usage.
ERRO[0000] unknown command "sdlkfjslkdjf" for "sonobuoy" 
exit status 1
```

I think it's cleaner without us logging the error as well as having cobra log the error.

PR #332, opened 2018-03-07
Cache clients for future use

### LINKED ISSUE #325: Cleanup kube.interface client handling
Per discussion, we should shuffle out the actual opening until they are used via accessor f(n)'s.  

Stretch goal for 0.11.0 time permitting.

liztio:
closes #325

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -21,6 +21,8 @@ import (
 
 	"github.com/heptio/sonobuoy/pkg/config"
 	"github.com/heptio/sonobuoy/pkg/plugin/aggregation"
+	"github.com/pkg/errors"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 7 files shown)


pkg/client/interfaces.go
```diff
@@ -77,26 +79,41 @@ type RetrieveConfig struct {
...
+	client        kubernetes.Interface
+	dynamicClient dynamic.ClientPool
 }
 
 // NewSonobuoyClient creates a new SonobuoyClient
 func NewSonobuoyClient(restConfig *rest.Config) (*SonobuoyClient, error) {
 	sc := &SonobuoyClient{
-		RestConfig: restConfig,
+		RestConfig:    restConfig,
+		client:        nil,
```
timothysc: it always defaults to nil, but meh I don't care.

liztio: leaving uninitialised fields makes me nervous 😓

PR #339, opened 2018-03-08
API Version preflight check for sonobuoy run

### LINKED ISSUE #330: Error when running on K8s 1.7.x
Description: When trying to run `sonobuoy run`, it errors out with the following message:

```
[centos@ sonobuoy]$ ./sonobuoy run
ERRO[0000] error attempting to run sonobuoy: failed to create object: failed to create API resource sonobuoy-serviceaccount: the server could not find the requested resource
```
After running `sonobuoy gen > manifest.yaml` and `kubectl apply -f manifest.yaml` thereafter produced the following error meesage:

```
[centos@ sonobuoy]$ kubectl apply -f manifest.yaml
namespace "heptio-sonobuoy" created
serviceaccount "sonobuoy-serviceaccount" created
configmap "sonobuoy-config-cm" created
configmap "sonobuoy-plugins-cm" created
pod "sonobuoy" created
service "sonobuoy-master" created
Error from server (BadRequest): error when creating "manifest.yaml": ClusterRoleBinding in version "v1" cannot be handled as a ClusterRoleBinding: no kind "ClusterRoleBinding" is registered for version "rbac.authorization.k8s.io/v1"
Error from server (BadRequest): error when creating "manifest.yaml": ClusterRole in version "v1" cannot be handled as a ClusterRole: no kind "ClusterRole" is registered for version "rbac.authorization.k8s.io/v1"
```

Sonobuoy Version: v0.11.0-alpha (built from source)
Kubectl Version: 1.7.4
Kubernetes Version: 1.7.8
Host OS: CentOS
Host OS Version: 7.4
Bootstrapper: kubeadm

liztio:
Closes #330

### DIFF
--- a/pkg/client/preflight.go
+++ b/pkg/client/preflight.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2018 Heptio Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package client
+
+import (
+	"fmt"
+
+	version "github.com/hashicorp/go-version"
+	"github.com/heptio/sonobuoy/pkg/buildinfo"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var preflightChecks = []func(kubernetes.Interface) error{
+	preflightDNSCheck,
+	preflightVersionCheck,
+}
+
+// PreflightChecks runs all preflight checks in order, returning the first error encountered.
+func (c *SonobuoyClient) PreflightChecks() []error {
+	client, err := c.Client()
+	if err != nil {
+		return []error{err}
+	}
+
+	errors := []error{}
+
+	for _, check := range preflightChecks {
+		if err := check(client); err != nil {
+			errors = append(errors, err)
+		}
+	}
+	return errors
+}
+
+const (
+	kubeSystemNamespace = "kube-system"
+	kubeDNSLabelKey     = "k8s-app"
+	kubeDNSLabelValue   = "kube-dns"
+)
+
+func preflightDNSCheck(client kubernetes.Interface) error {
+	selector := metav1.AddLabelToSelector(&metav1.LabelSelector{}, kubeDNSLabelKey, kubeDNSLabelValue)
+
+	obj, err := client.CoreV1().Pods(kubeSystemNamespace).List(
+		metav1.ListOptions{LabelSelector: metav1.FormatLabelSelector(selector)},
+	)
+	if err != nil {
+		return errors.Wrap(err, "could not retrieve list of pods")
+	}
+
+	if len(obj.Items) == 0 {
+		return errors.New("no kube-dns tests found")
+	}
+
+	return nil
+}
+
+var (
+	minimumKubeVersion = version.Must(version.NewVersion(buildinfo.MinimumKubeVersion))
+	maximumKubeVersion = version.Must(version.NewVersion(buildinfo.MaximumKubeVersion))
+)
+
+func preflightVersionCheck(client kubernetes.Interface) error {
+	versionInfo, err := client.Discovery().ServerVersion()
+	if err != nil {
+		return errors.Wrap(err, "failed to retrieve server version")
+	}
+
+	serverVersion, err := version.NewVersion(versionInfo.String())
+	if err != nil {
+		return errors.Wrap(err, "couldn't parse version string")
+	}
+
+	if serverVersion.LessThan(minimumKubeVersion) {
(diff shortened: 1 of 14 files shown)


pkg/client/preflight.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+
+	version "github.com/hashicorp/go-version"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apimachineryversion "k8s.io/apimachinery/pkg/version"
+	"k8s.io/client-go/kubernetes"
+)
+
+var preflightChecks = []func(kubernetes.Interface) error{
```
chuckha: whoa did gofmt not put a space here? `error{`

pkg/client/preflight.go
```diff
@@ -0,0 +1,109 @@
...
+	preflightDNSCheck,
+	preflightVersionCheck,
+}
+
+// PreflightChecks runs all preflight checks in order, returning the first error encountered.
```
chuckha: what was the decision behind bailing on the first one and not collecting a list of errors?

pkg/client/preflight.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if len(obj.Items) == 0 {
+		return errors.New("no kube-dns tests found")
+	}
+
+	return nil
+}
+
+var (
+	minimumKubeVersion = version.Must(version.NewVersion("1.8.0"))
```
chuckha: should these be consts up top?

pkg/client/preflight.go
```diff
@@ -0,0 +1,109 @@
...
+	maximumKubeVersion = version.Must(version.NewVersion("1.11.0"))
```
chuckha: 1.10 is the new one coming out and supporting up to a non-existent version is a little confusing, why are we doing that?

pkg/client/preflight.go
```diff
@@ -0,0 +1,109 @@
...
+func preflightVersionCheck(client kubernetes.Interface) error {
+	res, err := client.
+		Discovery().
+		RESTClient().
+		Get().
+		AbsPath("version").
+		Do().
+		Raw()
+	if err != nil {
+		return errors.Wrap(err, "couldn't retrieve version")
```
chuckha: what do you think about saying "can't figure out your version so you may encounter errors. we support versions after <minimum version>" and then passing the check? 

I think I prefer to be more flexible with these preflight checks and only bail if we're sure something isn't going to work. Your call though, I can understand that we'd have more success cases if we can guarantee we support the cluster.

pkg/client/preflight.go
```diff
@@ -0,0 +1,109 @@
...
+	serverVersion, err := version.NewVersion(versionInfo.String())
+	if err != nil {
+		return errors.Wrap(err, "couldn't parse version string")
+	}
+
+	if serverVersion.LessThan(minimumKubeVersion) {
+		return fmt.Errorf("Minimum kubernetes version is %s, got %s", minimumKubeVersion.String(), versionInfo.String())
+	}
+
+	if serverVersion.GreaterThan(maximumKubeVersion) {
```
chuckha: Could we get away with just the minimum? I'm trying to think of why we'd want a maximum and am not coming up with too many good reasons, maybe if k8s gets to 2.0 we might for a time only support up to 1.y -> 1.x and not 2.0? That reason to me doesn't justify having to figure out a maximum version when we bump these numbers. What am I overlooking?

pkg/client/preflight.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+var (
+	minimumKubeVersion = version.Must(version.NewVersion("1.8.0"))
+	maximumKubeVersion = version.Must(version.NewVersion("1.11.0"))
+)
+
+func preflightVersionCheck(client kubernetes.Interface) error {
+	res, err := client.
+		Discovery().
```
timothysc: why not Discovery().ServerVersion() ?

pkg/client/preflight.go
timothysc: I'd almost like to put all the version consts into the versions file, also I don' think we need a maximum.

pkg/client/preflight.go
liztio: We already have a `--skip-preflight` option. I think that's a good enough escape hatch

pkg/client/preflight.go
liztio: I... did not know about that!

pkg/client/preflight.go
liztio: it explicitly removes it. Who knew.

pkg/client/preflight.go
liztio: I try to keep constants near the function that uses them. Maybe that's not a good pattern?

pkg/client/preflight.go
liztio: I guess it's not obvious to me how to return a list of errors. There's no ErrorList that implements Error() is there?

pkg/client/preflight.go
chuckha: 🤯

pkg/client/preflight.go
chuckha: I was thinking []errors a la kubeadm https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/preflight/checks.go

pkg/client/preflight.go
chuckha: nah that's fine

pkg/client/preflight.go
liztio: We'd need to thread that all the way back to client.Interface, is the issue

PR #340, opened 2018-03-08
Add GCP provider to Sonobuoy

liztio:
This is a lot of new lines but I think it's important that we add this


timothysc:
Are you kidding me, this is really a requirement for gcp?

Gopkg.lock
```diff
@@ -365,6 +385,23 @@
   ]
   revision = "e19ae1496984b1c655b8044a65c0300a3c878dd3"
 
+[[projects]]
+  name = "google.golang.org/appengine"
```
timothysc: This just makes me want to cry.

liztio: we could force everyone to use `sonobuoy gen` for gcp if you want...

PR #342, opened 2018-03-08
Print out which plugins are running

### LINKED ISSUE #317: UX improvement to Run
Run should tell me what it's doing -- like running the conformance tests, running the quick smoke test, etc. I'm glad it tells me what it created, but it might be nice to know what it's doing on a higher level.

chuckha:
It's nice to know what is about to run on our cluster
so now the run command prints out the list of plugins
it is configured to tell sonobuoy to run.

Closes #317

### DIFF
--- a/cmd/sonobuoy/app/run.go
+++ b/cmd/sonobuoy/app/run.go
@@ -17,7 +17,9 @@ limitations under the License.
 package app
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/sonobuoy/app/run.go
```diff
@@ -83,7 +85,14 @@ func submitSonobuoyRun(cmd *cobra.Command, args []string) {
 		errlog.LogError(errors.Wrap(err, "could not create sonobuoy client"))
 		os.Exit(1)
 	}
-
+	m := genflags.mode.Get()
+	plugins := []string{}
+	for _, plugin := range m.Selectors {
+		plugins = append(plugins, plugin.Name)
```
timothysc: you could just construct the string with sprintfs here too.

chuckha: you're not wrong, there are many ways to tackle this. I chose this style so I could ignore thinking about when to add the `, ` to the string.

PR #345, opened 2018-03-08
Sonobuoy existing preflight check

liztio:
Checks to see if a Sonobuoy pod exists in this namespace already.

```
$ sonobuoy run
Running plugins: e2e, systemd-logs
INFO[0000] created object                                name=sonobuoy-serviceaccount namespace= resource=clusterrolebindings
INFO[0000] object already exists                         name=sonobuoy-serviceaccount namespace= resource=clusterroles
INFO[0000] created object                                name=sonobuoy-config-cm namespace=heptio-sonobuoy resource=configmaps
INFO[0000] created object                                name=sonobuoy-plugins-cm namespace=heptio-sonobuoy resource=configmaps
INFO[0000] created object                                name=sonobuoy namespace=heptio-sonobuoy resource=pods
INFO[0000] created object                                name=sonobuoy-master namespace=heptio-sonobuoy resource=services
$ sonobuoy run
Running plugins: e2e, systemd-logs
ERRO[0000] Preflight checks failed
ERRO[0000] sonobuoy run already exists in this namespace
```

### DIFF
--- a/pkg/client/preflight.go
+++ b/pkg/client/preflight.go
@@ -22,17 +22,19 @@ import (
 	version "github.com/hashicorp/go-version"
 	"github.com/heptio/sonobuoy/pkg/buildinfo"
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
-var preflightChecks = []func(kubernetes.Interface) error{
+var preflightChecks = []func(kubernetes.Interface, *PreflightConfig) error{
 	preflightDNSCheck,
 	preflightVersionCheck,
+	preflightExistingSonobuoy,
 }
 
 // PreflightChecks runs all preflight checks in order, returning the first error encountered.
-func (c *SonobuoyClient) PreflightChecks() []error {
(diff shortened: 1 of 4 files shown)


pkg/client/preflight.go
```diff
@@ -97,3 +99,18 @@ func preflightVersionCheck(client kubernetes.Interface) error {
...
+	_, err := client.CoreV1().Pods(cfg.Namespace).Get("sonobuoy", metav1.GetOptions{})
+	switch {
+	// Pod doesn't exist: great!
+	case apierrors.IsNotFound(err):
+		return nil
+	case err != nil:
+		return errors.Wrap(err, "error checking for Sonobuoy pod")
+	// No error: pod exists
+	case err == nil:
+		return errors.New("sonobuoy run already exists in this namespace")
```
chuckha: probably out of scope for this, but what do you think about giving the user some action on a failed preflight? for instance in this case, either delete the pod or run it in a different namespace

liztio: Yeah, I thought about that. It doesn't feel right to put that sort of thing inside of a error message, though. Maybe at some point we could make an ErrorWithSolution() type that had both an `Error()` and a `Solution()` method on it, then the CLI could check for the latter?

chuckha: yeah!

PR #365, opened 2018-03-13
Adds submission templates to make sure we are consistent.

timothysc:
I'm a process person, and this will force my hand to write the details. 

This will should hopefully make @liztio and @chuckha happier.


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
+
+**Is this a BUG REPORT or FEATURE REQUEST?**:
```
timothysc: @liztio we can modify this for RFEs to make them more KEP-esk too.

PR #366, opened 2018-03-13
Consider core dns in the preflight DNS check

pipo02mix:
Since now [coredns](https://github.com/coredns/coredns) it is a real alternative for DNS server in Kubernetes I would like to be able to run sonobuoy without modifying default labels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 
 # Development manifest
 examples/dev.yaml
+
+# Unencrypted credentials
+heptio-images-ee4b0474b93e.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,19 +18,10 @@ before_install:
 deploy:
 - provider: script
   skip_cleanup: true
-  script: curl -sL https://git.io/goreleaser | bash
+  script: ./travis-deploy.sh
   on:
     repo: heptio/sonobuoy
-    tags: true
-# Push to GCR
-- provider: script
-  script:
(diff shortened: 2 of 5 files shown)


pkg/client/preflight.go
```diff
@@ -54,20 +54,31 @@ const (
 	kubeSystemNamespace = "kube-system"
 	kubeDNSLabelKey     = "k8s-app"
 	kubeDNSLabelValue   = "kube-dns"
+	coreDNSLabelValue   = "coredns"
```
timothysc: What deployment tool is this sets this label?  Kubeadm overloads for the kube-dns label w/coreos.

timothysc: I think we should check kubespray and make a call, I'll try to see if they do the same.

pipo02mix: There is a fresh PR approved for adding coredns to kubespray. There I dont see any overwriting action

https://github.com/kubernetes-incubator/kubespray/pull/2462

timothysc: @pipo02mix - It appears they add `k8s-app : coredns` and that seems like a good addition

timothysc:
@pipo02mix - we use DCO signoff checker on our repos.  Could you amend your commit with `--signoff` 

/cc @liztio for review as you wrote this section

pipo02mix:
I have screwed it up doing the sign-off. For some reason, I could not sign only last two commits. I close this in favor of a [new PR](https://github.com/heptio/sonobuoy/pull/366)

PR #369, opened 2018-03-13
Consider core dns in the preflight DNS check

pipo02mix:
Since now [coredns](https://github.com/coredns/coredns) it is a real alternative for DNS server in Kubernetes I would like to be able to run sonobuoy without modifying default labels. This PR replace #366

Signed-off-by: Fernando Ripoll <pipo02mix@gmail.com>

### DIFF
--- a/pkg/client/preflight.go
+++ b/pkg/client/preflight.go
@@ -54,20 +54,31 @@ const (
 	kubeSystemNamespace = "kube-system"
 	kubeDNSLabelKey     = "k8s-app"
 	kubeDNSLabelValue   = "kube-dns"
+	coreDNSLabelValue   = "coredns"
 )
 
 func preflightDNSCheck(client kubernetes.Interface, cfg *PreflightConfig) error {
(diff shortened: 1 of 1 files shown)


pkg/client/preflight.go
```diff
@@ -54,20 +54,31 @@ const (
...
+		coreDNSLabelValue,
+	}
 
-	obj, err := client.CoreV1().Pods(kubeSystemNamespace).List(
-		metav1.ListOptions{LabelSelector: metav1.FormatLabelSelector(selector)},
-	)
-	if err != nil {
-		return errors.Wrap(err, "could not retrieve list of pods")
+	var nPods = 0
+	for _, labelValue := range dnsLabels {
```
timothysc: Can you construct a set based label selector https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement  ?  I'd have to dig the upstream test cases for `valueA or valueB in key`

timothysc: So ran through and I guess you can't.

PR #373, opened 2018-03-14
fixes signal handling bug

chuckha:

Fixes a signal handling bug.
No issue created

This is what I wanted to do originally.

When the worker gets a shutdown signal, it waits for $GRACEFUL_SHUTDOWN_PERIOD and then actually cleans up.

What it did before was simply sleep and block the main thread then exit.

**Release note**:
```
release-note
```

### DIFF
--- a/pkg/worker/worker.go
+++ b/pkg/worker/worker.go
@@ -44,8 +44,10 @@ func init() {
 // 3. The done file contains a single string of the results to be sent to the master
 func GatherResults(waitfile string, url string, client *http.Client) error {
 	logrus.WithField("waitfile", waitfile).Info("Waiting for waitfile")
-	stop := sigHandler()
+	signals := sigHandler()
 	ticker := time.Tick(1 * time.Second)
+	stop := make(chan struct{}, 1)
+	// TODO(chuckha) evaluate wait.Until [https://github.com/kubernetes/apimachinery/blob/e9ff529c66f83aeac6dff90f11ea0c5b7c4d626a/pkg/util/wait/wait.go]
 	for {
 		select {
 		case <-ticker:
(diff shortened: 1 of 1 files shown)


pkg/worker/worker.go
```diff
@@ -91,7 +97,6 @@ func sigHandler() <-chan struct{} {
 		signal.Notify(sigc, syscall.SIGTERM)
 		sig := <-sigc
 		logrus.WithField("signal", sig).Info("got a signal, waiting then sending the real shutdown signal")
-		close(stop)
```
chuckha: channels in a select should not be closed because they are always ready and don't give other channels a chance to get read from.

chuckha:
logs showing this works:

```
$ docker logs -f 6ba4c22e23ab
time="2018-03-14T16:10:20Z" level=info msg="Waiting for waitfile" waitfile=/tmp/results/done
time="2018-03-14T16:12:17Z" level=info msg="got a signal, waiting then sending the real shutdown signal" signal=terminated
time="2018-03-14T16:12:24Z" level=info msg="Detected done file, transmitting result file" resultFile=/tmp/results/e2e.tar.gz
```

pkg/worker/worker.go
```diff
@@ -44,19 +44,25 @@ func init() {
...
 		case <-ticker:
 			if resultFile, err := ioutil.ReadFile(waitfile); err == nil {
 				logrus.WithField("resultFile", string(resultFile)).Info("Detected done file, transmitting result file")
 				return handleWaitFile(string(resultFile), url, client)
 			}
+		case <-signals:
+			// Run a goroutine here so we can keep checking the done file before cleaning up.
+			go func() {
+				time.Sleep(plugin.GracefulShutdownPeriod)
+				stop <- struct{}{}
```
timothysc: can't you just set a timer for the stop channel using wait.Until()? 

Looks at upstream examples.

timothysc: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go

chuckha: We could do that, but we still have to coordinate three cases here.

1) We got the done file
2) We got a SIGINT
3) We got a "ok shut everything down message"

If I were to wait.Until() with some callback that executes every period until the stop channel passed in is closed, I think it would be harder to read, but I'm more used to the for/select.

It's your call. I literally wrote the code here so if you prefer this style, I can refactor.

This might also be a good suggestion for https://github.com/heptio/sonobuoy/issues/372 since now we've parameterized the callback as `func()`.

Example:
```go
stop := make(chan struct{}{})
signals :=  sigHandler()
go func() {
    <-signals
    time.Sleep(plugin.GracefulShutdownPeriod)
    close(stop)
}
wait.Until(func(){
         logrus.WithField("resultFile", string(resultFile)).Info("Detected done file, transmitting result file")
        return handleWaitFile(string(resultFile), url, client)
    }
}, 1 *time.Second, stop)
```

timothysc: I'm ok with the fix, please add a `//TODO evaluate wait.Until time permitting` and we can move onto other things.

PR #374, opened 2018-03-14
Refactors worker waiting function

### LINKED ISSUE #372: Refactor worker signal handling to be testable
Feature request.

**What happened**:
Testing signal handling is clumsy and awkward for the worker, but it really needs to be tested.

I didn't expect testing to be so clumsy for this part of the code

**Environment**:
- Kernel (e.g. `uname -a`):

chuckha:
* Adds a test to the cleanup case.

Closes #372

I think this simplifies the logic for the worker.

- Fixes #372

It's ok if you don't think this actually makes the code easier to read/follow. I'm happy to close it without merging.

**Release note**:
```
release-note
```

### DIFF
--- a/cmd/sonobuoy/app/worker.go
+++ b/cmd/sonobuoy/app/worker.go
@@ -22,12 +22,16 @@ import (
 	"encoding/pem"
 	"net/http"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
+	"time"
 
 	"github.com/heptio/sonobuoy/pkg/errlog"
 	"github.com/heptio/sonobuoy/pkg/plugin"
 	"github.com/heptio/sonobuoy/pkg/worker"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/sonobuoy/app/worker.go
```diff
@@ -64,6 +68,21 @@ func runGather(cmd *cobra.Command, args []string) {
...
+// elapses after a SIGTERM is received.
+func sigHandler(timeout time.Duration) <-chan struct{} {
+	stop := make(chan struct{})
+	go func() {
+		sigc := make(chan os.Signal, 1)
+		signal.Notify(sigc, syscall.SIGTERM)
+		sig := <-sigc
+		logrus.WithField("signal", sig).Info("received a signal. Waiting then sending the real shutdown signal.")
+		time.Sleep(timeout)
+		stop <- struct{}{}
```
timothysc: How do you notify to cleanup?  
It looks like it just hard-stops now.

chuckha: I should add cleanups to the stop case. I don't think they are super important here since that is the end of the program, but better to always cleanup rather than sometimes clean up.

timothysc:
need a re-base then feel free to self-merge.

PR #375, opened 2018-03-14
Parallel e2e

jimmidyson:

To speed up test runs by enabling option of parallelism via ginkgo (requires heptio/kube-conformance#17 to be merged and released before this to be functional).

This also enables specifying the kube-conformance image which I needed during development cycles, but could be useful for people running other e2e images.

Plus this switches to use single-quotes YAML strings in templates to fix using e.g. `\[Serial\]` in skip/focus (YAML doesn't like non-escaped backslashes in double quotes strings so easier to switch to single-quoted to not require escaping at all.

Um, I haven't raised one...

We'd love to get this in ASAP so we can start using it for all our CI needs. We currently have hand-rolled e2e test 

**Release note**:
```
Parallel e2e tests can be enabled by specifying `--e2e-parallel=y` flag
```

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -78,8 +78,9 @@ func AddSonobuoyConfigFlag(cfg *SonobuoyConfig, flags *pflag.FlagSet) {
 }
 
 const (
-	e2eFocusFlag = "e2e-focus"
-	e2eSkipFlag  = "e2e-skip"
+	e2eFocusFlag    = "e2e-focus"
+	e2eSkipFlag     = "e2e-skip"
+	e2eParallelFlag = "e2e-parallel"
 )
 
 // AddE2EConfigFlags adds two arguments: --e2e-focus and --e2e-skip. These are not taken as pointers, as they are only used by GetE2EConfig. Instead, they are returned as a Flagset which should be passed to GetE2EConfig. The returned flagset will be added to the passed in flag set.
@@ -95,12 +96,16 @@ func AddE2EConfigFlags(flags *pflag.FlagSet) *pflag.FlagSet {
 		e2eSkipFlag, defaultMode.E2EConfig.Skip,
 		"Specify the E2E_SKIP flag to the conformance tests. Overrides --mode.",
 	)
+	e2eFlags.String(
+		e2eParallelFlag, defaultMode.E2EConfig.Parallel,
+		"Specify the E2E_PARALLEL flag to the conformance tests. Overrides --mode.",
+	)
 	flags.AddFlagSet(e2eFlags)
 	return e2eFlags
 }
 
-// GetE2EConfig gets the E2EConfig from the mode, then overrides them with e2e-focus and e2e-skip if they are provided.
-// We can't rely on the zero value of the flags, as "" is a valid  focus or skip value.
+// GetE2EConfig gets the E2EConfig from the mode, then overrides them with e2e-focus, e2e-skip and e2e-parallel if they
+// are provided. We can't rely on the zero value of the flags, as "" is a valid focus, skip or parallel value.
 func GetE2EConfig(mode ops.Mode, flags *pflag.FlagSet) (*ops.E2EConfig, error) {
 	cfg := mode.Get().E2EConfig
 	if flags.Changed(e2eFocusFlag) {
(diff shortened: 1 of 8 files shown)


jimmidyson:
Should say this requires heptio/kube-conformance#17 to be merged and released first.

cmd/sonobuoy/app/args.go
```diff
@@ -87,12 +96,16 @@ func AddE2EConfigFlags(flags *pflag.FlagSet) *pflag.FlagSet {
 		e2eSkipFlag, defaultMode.E2EConfig.Skip,
 		"Specify the E2E_SKIP flag to the conformance tests. Overrides --mode.",
 	)
+	e2eFlags.String(
```
timothysc: kind of want to hide this for power CI users only, b/c folks can destroy their clusters.

jimmidyson: Marked flag as hidden.

cmd/sonobuoy/app/args.go
```diff
@@ -87,12 +96,17 @@ func AddE2EConfigFlags(flags *pflag.FlagSet) *pflag.FlagSet {
...
+		e2eParallelFlag, defaultMode.E2EConfig.Parallel,
```
chuckha: Please add the defaults to the specified modes https://github.com/heptio/sonobuoy/blob/master/pkg/client/mode.go#L82, I think the default should be "n" or whatever the no value is for all three modes

cmd/sonobuoy/app/args.go
```diff
@@ -70,8 +78,9 @@ func AddSonobuoyConfigFlag(cfg *SonobuoyConfig, flags *pflag.FlagSet) {
...
 
 const (
-	e2eFocusFlag = "e2e-focus"
-	e2eSkipFlag  = "e2e-skip"
+	e2eFocusFlag    = "e2e-focus"
+	e2eSkipFlag     = "e2e-skip"
+	e2eParallelFlag = "e2e-parallel"
 )
 
 // AddE2EConfigFlags adds two arguments: --e2e-focus and --e2e-skip. These are not taken as pointers, as they are only used by GetE2EConfig. Instead, they are returned as a Flagset which should be passed to GetE2EConfig. The returned flagset will be added to the passed in flag set.
```
chuckha: please update the godoc to reflect the new flag and comment on its hidden status and why it is hidden in the UI

pkg/client/gen.go
```diff
@@ -58,14 +61,16 @@ func (c *SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
...
 	tmplVals := &templateValues{
-		E2EFocus:        cfg.E2EConfig.Focus,
-		E2ESkip:         cfg.E2EConfig.Skip,
-		SonobuoyConfig:  string(marshalledConfig),
-		SonobuoyImage:   cfg.Image,
-		Version:         buildinfo.Version,
-		Namespace:       cfg.Namespace,
-		EnableRBAC:      cfg.EnableRBAC,
-		ImagePullPolicy: cfg.ImagePullPolicy,
+		E2EFocus:             strings.Replace(cfg.E2EConfig.Focus, "'", "''", -1),
```
chuckha: I see this covered in #376 -- is this PR pending some updates?

Could you add a comment about why replacing one single quote with two single quotes is the thing to do here? I don't quite understand the replacement after looking at the YAML template change as well.

cmd/sonobuoy/app/args.go
jimmidyson: Defaulted to `"1"` for all 3 modes as discussed in https://github.com/heptio/kube-conformance/pull/17#discussion_r174653310

pkg/client/gen.go
jimmidyson: If we merged #376 then I can rebase this and drop this from this PR. I'll add comments in other PR as it's more focused.

cmd/sonobuoy/app/args.go
jimmidyson: Done.

pkg/client/gen.go
jimmidyson: Added comments in #376.

chuckha: Awesome. That was super helpful. Let's get the other one merged first so you can drop it from here.

PR #376, opened 2018-03-15
Fix up default focus/skip regexps

jimmidyson:
Signed-off-by: Jimmi Dyson <jimmidyson@gmail.com>

Current regexps for focus and skip were too loose and could have matched specs that they weren't supposed to match against. For example, old focus default was `Conformance` which would have matched any specs with `Conformance`, whereas this should match against `[Conformance]` as per conformance suite description:

> The standard set of conformance tests is currently those defined by the [Conformance] tag in the kubernetes e2e suite.

This also updates the default skip regexp, without changing the intended set of specs to be skipped.

- Sorry not created one.

Hope that this is self-explanatory. Note also the switch to single quotes in YAML file to remove the need to escape special characters (e.g. backslash, colon, etc) in regexp strings.

**Release note**:
Not sure this needs one? Behaviour should be unchanged, more helping to prevent issues or weird behaviour in future.

### DIFF
--- a/pkg/client/gen.go
+++ b/pkg/client/gen.go
@@ -19,6 +19,7 @@ package client
 import (
 	"bytes"
 	"encoding/json"
+	"strings"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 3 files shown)


pkg/client/gen.go
```diff
@@ -57,9 +58,17 @@ func (c *SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
 		return nil, errors.Wrap(err, "couldn't marshall selector")
 	}
 
+	// Template values that are regexps (`E2EFocus` and `E2ESkip`) are
```
jimmidyson: @chuckha Here's an (overly verbose) explanation of why single quotes in YAML strings in this case.

PR #412, opened 2018-04-10
WIP: Private registry

### LINKED ISSUE #160: Config option to reference private registry
We are attempting to deploy sonobuoy on a k8s cluster configured behind a firewall with no internet access.

The sonobuoy images are available in a private registry and quickstart.yaml updated to reflect this.  When launching the deployment, the sonobuoy specific pods launch with no issues.

However, after a period of time, the below pods fails to launch in the newly created "e2e-tests" namespace.

A kubectl describe against this:
`kubectl describe pod pod-submit-remove-16bb7011-d055-11e7-ba38-0a580a02306e -n=e2e-tests-pods-jgshk`

shows that it is failing to pull an nginx image as below:

`Failed to pull image "gcr.io/google-containers/nginx-slim-amd64:0.20": rpc error: code = Unknown desc = Error response from daemon: Get https://gcr.io/v1/_ping: dial tcp 173.194.69.82:443: i/o timeout`

As the cluster does not have internet access, it is unable to pull and access this image from gcr.io

This nginx image from gcr.io is not referenced in the deployment yaml so is it possible to add a configuration option where the nginx image can be accessed from a private internal registry?

Something like the below if this is feasible:
`private-registry:443/gcr.io/google-containers/nginx-slim-amd64:0.20`

Any assistance with this would be much appreciated.

lewismarshall:

It informs what's required for using a private registry or running tests in an air-gapped environment.

- Fixes #160 

An early draft - got me out of jail - happy to tidy up if you want to use it.

**Release note**:
```
Adds the --private-registry option to gen and run sub commands and passes through to e2e tests with E2E_REGISTRY_OVERRIDE env option. See dirty hack for upstream - https://github.com/appvia/kube-conformance/pull/1/files
```

### DIFF
--- a/pkg/templates/manifest.go
+++ b/pkg/templates/manifest.go
@@ -154,6 +154,8 @@ spec:
       valueFrom:
         fieldRef:
           fieldPath: status.podIP
+    - name: SONOBUOY_IMAGE
+      value: {{.SonobuoyImage}}
     image: {{.SonobuoyImage}}
     imagePullPolicy: {{.ImagePullPolicy}}
     name: kube-sonobuoy
--- a/pkg/templates/manifest.go
+++ b/pkg/templates/manifest.go
@@ -89,6 +89,8 @@ data:
         value: '{{.E2ESkip}}'
       - name: E2E_PARALLEL
         value: '{{.E2EParallel}}'
+      - name: E2E_REGISTRY_OVERRIDE
+        value: '{{.PrivateRegistry}}'
       command: ["/run_e2e.sh"]
       image: {{.KubeConformanceImage}}
       imagePullPolicy: {{.ImagePullPolicy}}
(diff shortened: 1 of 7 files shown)


timothysc:
@lewismarshall Thank you for the patch.  

Is your goal to rebuild all the dependent images and host from a local registry?

lewismarshall:
@timothysc the goal was to run a kube-conformance e2e run in an air-gapped environment, which I have now done, with this change and the related kube-conformance and kubernetes hacks.

Unfortunatly, I have to work on other things now but I'm providing the kube-conformance hack as a proof of concept for "can the e2e.test binary work air-gapped". It crucially added this kubernetes patch file: https://github.com/appvia/kube-conformance/pull/1/files#diff-724e99cba5f5ce487c2cae2c447786f8

pkg/templates/manifest.go
```diff
@@ -113,7 +115,7 @@ data:
         value: /tmp/results
       - name: CHROOT_DIR
         value: /node
-      image: gcr.io/heptio-images/sonobuoy-plugin-systemd-logs:latest
+      image: {{.PrivateRegistry}}/sonobuoy-plugin-systemd-logs:latest
```
lewismarshall: I can see already that this will fail for the normal case - I need to add a specific template variable (or if prefered add a `--systemd-logs-image param`)

timothysc:
Dropping this breadcrumb here when we loop back to this - https://github.com/kubernetes/kubernetes/issues/38067

johnSchnake:
#612 Resolved #160 so I think this can be closed now. Try out the new functionality and let us know what you think.

timothysc:
Thx @johnSchnake

PR #453, opened 2018-06-27
Retrieve error handling

### LINKED ISSUE #407: Make 'sonobuoy retrieve' cli command aware if tarball is available or not
FEATURE REQUEST

**What happened**:
I prepared script to run sonobuoy and gather results. Basically script has steps:
1. `sonobuoy run`
2. in a loop wait for all plugins to complete 
3. `sonobuoy retrieve`
4. `tar xzf *.tar.gz`

Last step failed because there is no tarball after retrieve, just a folder (with UID as name). Sonobuoy returned 0 exit code.

It would be better UX if sonobuoy cli retrieve would return some errorCodes, f.e.:
0 - ok, tarball downloaded
1 - error, cannot retrieve the tarball, not recoverable error
2 (or other number) - tarball not ready, you can try again later

I think it would make scripting easier (now I can just check if there's a tarball in specified folder)

**Environment**:
- Kernel (e.g. `uname -a`):

liztio:
Right now `sonobuoy retrieve` provides very bad error handling. This improves it:

```
$ sonobuoy retrieve
error retrieving results: error sending request: Post https://liz-k8s-liz-d22b5dmgmt.eastus.cloudapp.azure.com/api/v1/namespaces/heptio-sonobuoy/pods/sonobuoy/exec?command=%2Fusr%2Fbin%2Fenv&command=bash&command=-c&command=tar+cf+-+%2Ftmp%2Fsonobuoy%2F%2A.tar.gz&container=kube-sonobuoy&container=kube-sonobuoy&stdout=true: dial tcp: lookup liz-k8s-liz-d22b5dmgmt.eastus.cloudapp.azure.com: no such host
zsh: exit 2     sonobuoy retrieve
```

```
$ sonobuoy retrieve
Results not ready yet. Check `sonobuoy status` for status.
zsh: exit 1     sonobuoy retrieve
```

- Fixes #407

**Release note**:
```
better sonobuoy retrieve error handling
client.Retrieve returns error channel
```

### DIFF
--- a/pkg/client/retrieve.go
+++ b/pkg/client/retrieve.go
@@ -24,8 +24,6 @@ import (
 	"path"
 	"path/filepath"
 
-	"github.com/sirupsen/logrus"
-
 	"github.com/heptio/sonobuoy/pkg/config"
 	"github.com/pkg/errors"
 
@@ -34,10 +32,19 @@ import (
 	"k8s.io/client-go/tools/remotecommand"
 )
 
-func (c *SonobuoyClient) RetrieveResults(cfg *RetrieveConfig) (io.Reader, error) {
+var tarCommand = []string{
+	"/usr/bin/env",
+	"bash",
+	"-c",
+	fmt.Sprintf("tar cf - %s/*.tar.gz", config.MasterResultsPath),
+}
+
+func (c *SonobuoyClient) RetrieveResults(cfg *RetrieveConfig) (io.Reader, <-chan error) {
+	ec := make(chan error, 1)
 	client, err := c.Client()
 	if err != nil {
-		return nil, err
+		ec <- err
+		return nil, ec
 	}
 	restClient := client.CoreV1().RESTClient()
 	req := restClient.Post().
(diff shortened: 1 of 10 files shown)


pkg/client/retrieve.go
```diff
@@ -34,10 +32,19 @@ import (
 	"k8s.io/client-go/tools/remotecommand"
 )
 
-func (c *SonobuoyClient) RetrieveResults(cfg *RetrieveConfig) (io.Reader, error) {
+var tarCommand = []string{
```
chuckha: why do we need to wrap this call with `bash -c`?

pkg/client/retrieve.go
```diff
@@ -34,10 +32,19 @@ import (
...
+	"/usr/bin/env",
+	"bash",
+	"-c",
+	fmt.Sprintf("tar cf - %s/*.tar.gz", config.MasterResultsPath),
+}
+
+func (c *SonobuoyClient) RetrieveResults(cfg *RetrieveConfig) (io.Reader, <-chan error) {
```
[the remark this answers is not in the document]
liztio: This can't really be blocking if we want to keep the io.Reader contract. In the correct case, the error code isn't available until after execution is completed. In the error case, it could be an arbitrary amount of time (10 or more seconds) until we get the report of the status code.

pkg/client/retrieve.go
liztio: because the `sh` execution doesn't handle the `*` globbing

pkg/client/retrieve.go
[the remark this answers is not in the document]
timothysc: I don't mind blocking for (X) with a ctx arg provided, that standard go.  Otherwise, we need to revisit the whole interface.

pkg/client/retrieve.go
chuckha: were you running into issues with the existing code?

liztio: Yes. If you ran the command before sonobuoy had finished running, it would tar up and retrieve a temporary directory that didn't have anything useful in it. Users had to detect if the untarred file ended in .tar.gz to determine if it was successful.

PR #459, opened 2018-07-03
Client go update

### LINKED ISSUE #450: Update / vet our usage of kubernetes API for 1.11 update
We need to go through our API usage and use the lowest API group available to our supported versions

1.9, 1.10 and 1.11.

chuckha:
This PR updates our codebase to use 1.11 client-go and friends.

I also checked API versions and we're still exactly where we need to be with regard to API changes. No API we care about changed in 1.11

- Fixes #450 

**Release note**:
```
NONE
```


cmd/sonobuoy/app/retrieve.go
```diff
@@ -59,12 +59,18 @@ func retrieveResults(cmd *cobra.Command, args []string) {
...
 
-	restConfig, err := rcvFlags.kubecfg.Get()
+	cfg, err := rcvFlags.kubecfg.Get()
 	if err != nil {
-		errlog.LogError(fmt.Errorf("failed to get kubernetes client: %v", err))
+		errlog.LogError(errors.Wrap(err, "failed to get kubernetes client"))
 		os.Exit(1)
 	}
-	sbc, err := client.NewSonobuoyClient(restConfig)
+	skc, err := dynamic.NewApiHelper(cfg)
```
timothysc: Can probably wrap this repeated section behind the common .Get() operation.

pkg/client/interfaces.go
```diff
@@ -78,19 +78,24 @@ type PreflightConfig struct {
 	Namespace string
 }
 
+// SonobuoyKubeAPIClient is the interface Sonobuoy uses to communicate with a kube-apiserver.
+type SonobuoyKubeAPIClient interface {
```
timothysc: Really?

pkg/dynamic/client.go
```diff
@@ -0,0 +1,89 @@
...
+
+type ApiHelper struct {
+	Discover discovery.DiscoveryInterface
+	Client   dynamic.Interface
+}
+
+func NewApiHelper(c *rest.Config) (*ApiHelper, error) {
+	client, err := dynamic.NewForConfig(c)
+	if err != nil {
+		return nil, fmt.Errorf("could not create dynamic client: %v", err)
```
timothysc: This looks like you want to use https://godoc.org/github.com/pkg/errors#Wrap

pkg/dynamic/client.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, fmt.Errorf("could not create discovery client: %v", err)
+	}
+	return &ApiHelper{
+		Discover: discover,
+		Client:   client,
+	}, nil
+}
+
+func (a *ApiHelper) CreateObject(obj *unstructured.Unstructured) error {
+	groupResources, err := restmapper.GetAPIGroupResources(a.Discover)
```
timothysc: Just based on the flow of data in this function you could probably reduce this into a set of private or public functions and separate out some of the behavior into sub-components + with tests.  

Did you copy some of this from upstream examples in the latest client-go?

pkg/client/interfaces.go
chuckha: Better than having the sonobuoy client require a struct imo

but yes, the naming needs work

pkg/dynamic/client.go
chuckha: I couldn't find any examples of the dynamic client upstream. this was the result of reading a lot of code fairly quickly

chuckha:
```
test salazar:sonobuoy cha$ sonobuoy run --mode quick
Running plugins: e2e
INFO[0002] created object                                name=heptio-sonobuoy namespace= resource=namespaces
INFO[0002] created object                                name=sonobuoy-serviceaccount namespace=heptio-sonobuoy resource=serviceaccounts
INFO[0002] created object                                name=sonobuoy-serviceaccount-heptio-sonobuoy namespace= resource=clusterrolebindings
INFO[0003] created object                                name=sonobuoy-serviceaccount namespace= resource=clusterroles
INFO[0003] created object                                name=sonobuoy-config-cm namespace=heptio-sonobuoy resource=configmaps
INFO[0003] created object                                name=sonobuoy-plugins-cm namespace=heptio-sonobuoy resource=configmaps
INFO[0003] created object                                name=sonobuoy namespace=heptio-sonobuoy resource=pods
INFO[0003] created object                                name=sonobuoy-master namespace=heptio-sonobuoy resource=services
```

sonobuoy built with this commit.

pkg/client/run.go
```diff
@@ -67,44 +55,31 @@ func (c *SonobuoyClient) Run(cfg *RunConfig) error {
...
-		err := createObject(c.DynamicClientPool(), &obj, mapper)
+		name, err := c.dynamicClient.Name(obj)
 		if err != nil {
+			return errors.Wrap(err, "could not get object name")
+		}
+		namespace, err := c.dynamicClient.Namespace(obj)
+		if err != nil {
+			return errors.Wrap(err, "could not get object namespace")
+		}
+		newObj, err := c.dynamicClient.CreateObject(obj)
```
timothysc: It looks like you're ignoring this error....?

pkg/dynamic/client.go
```diff
@@ -0,0 +1,95 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+)
+
+// A scoped down meta.MetadataAccessor
+type MetadataAccessor interface {
```
timothysc: This is much better, thank you!

pkg/dynamic/client.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get name for object")
+	}
+	namespace, err := a.Accessor.Namespace(obj)
+	if err != nil {
+		return nil, errors.Wrapf(err, "couldn't get namespace for object %s", name)
+	}
+
+	rsc := a.Client.Resource(restMapping.Resource)
+	ri := rsc.Namespace(namespace)
```
timothysc: is there any chance rsc could be nil?

pkg/client/run.go
chuckha: the error gets handled on line 75

i agree though it's a weird pattern and may warrant a comment.

pkg/dynamic/client.go
chuckha: There sure is. I'll fix that, good catch

PR #463, opened 2018-07-03
Return a non-zero status code for failed plugins

### LINKED ISSUE #404: Get non-zero exit code from sonnobuoy cli in case of returned failed status
Feature Request

**What happened**:
I run: 
`sonobuoy status` 
and get information about failed plugin. 
But when I run 
`echo $?` 
I got 
`0` 
exit code.

If status is failed I'd like to get non zero exit code so I can handle it properly in case of using sonobuoy cli in bash scripts. Without it I need to parse the output.

I run sonobuoy v0.11.1

**Environment**:
- Kernel (e.g. `uname -a`):

chuckha:
Fixes #404

I can't really write a test for this, so it's kind of a best effort deal.

This returns a non-zero exit code for `sonobuoy status` when the status is that a plugin has failed.

- Fixes #404 

**Release note**:
```
NONE
```

### DIFF
--- a/cmd/sonobuoy/app/status.go
+++ b/cmd/sonobuoy/app/status.go
@@ -80,9 +80,20 @@ func getStatus(cmd *cobra.Command, args []string) {
 	} else {
 		err = printSummary(os.Stdout, status)
 	}
+
 	if err != nil {
 		errlog.LogError(err)
 		os.Exit(1)
+	}
+	os.Exit(exitCode(status))
+}
+
+func exitCode(status *aggregation.Status) int {
+	switch status.Status {
+	case aggregation.FailedStatus:
+		return 1
+	default:
+		return 0
 	}
 }
 


cmd/sonobuoy/app/status.go
```diff
@@ -80,10 +80,21 @@ func getStatus(cmd *cobra.Command, args []string) {
...
 	if err != nil {
 		errlog.LogError(err)
 		os.Exit(1)
 	}
+	os.Exit(exitCode(status))
+}
+
+func exitCode(status *aggregation.Status) int {
+	switch status.Status {
+	case aggregation.FailedStatus:
```
timothysc: Why is this not an error on GetStatus?

chuckha: because it's not an error getting the status -- the error is on the plugin and is reported through the status.

chuckha: I don't have an example, but it will look like this (hand crafted output):

```
test salazar:sonobuoy cha$ sonobuoy status
PLUGIN		STATUS		COUNT
e2e		complete	1
systemd_logs	failed	1

Sonobuoy has failed. You can see what happened with `sonobuoy logs`.

test salazar:sonobuoy cha$ echo $?
1
```

PR #464, opened 2018-07-05
Improve error logging SEP

chuckha:

Adds a SEP for error logging.

### DIFF
--- a/docs/enhancements/error-logging-improvement.md
+++ b/docs/enhancements/error-logging-improvement.md
@@ -1,0 +1,62 @@
+# Error Logging Improvement
+
+## Table of Contents
+
+* [Summary](#summary)
+* [Objectives](#objectives)
+  * [Goals](#goals)
+  * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+  * [User Stories](#user-stories)
+* [Unresolved Questions](#unresolved-questions)
+
+# Summary
+
+The log produced by Sonobuoy is used to be able to diagnose clusters and
+troubleshoot users. This makes Sonobuoy's error logging a requirement, not a
+nice-to-have. A requirement must have automated testing or it's going to break.
+Error logging in Sonobuoy, at the time of writing this document, is difficult if
+not impossible to unit test.
+
+# Objectives
+
+The main objective is to unify how we write to the error log and treat the error
+log as a proper dependency.
+
+## Goals
+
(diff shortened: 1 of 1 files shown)


docs/enhancements/error-logging-improvement.md
```diff
@@ -0,0 +1,62 @@
...
+not impossible to unit test.
+
+# Objectives
+
+The main objective is to unify how we write to the error log and treat the error
+log as a proper dependency.
+
+## Goals
+
+* Write unit tests for functions that call the error log.
```
timothysc: I don't mind writing unit tests on errors, but it's generally faux pas to write UT's on error logs.  Unless you have a utility to splice the error from the log.

docs/enhancements/error-logging-improvement.md
```diff
@@ -0,0 +1,62 @@
...
+* Treat info/debug logging as a proper dependency. This would be nice but is not
+  the focus.
+* Get 100% unit test coverage on error log call-sites. Again, nice but probably
+  not worth the effort.
+
+# Proposal
+
+## Features of the error log
+
+1. The error log should always (some exceptions are ok, like power disruption) be
```
timothysc: I think having a standard for them is good.  I think having an enumerated errors list is also good.

docs/enhancements/error-logging-improvement.md
chuckha: generally, I agree that it's a faux pas to test error logs. But this SEP outlines why we *should* test our error log calls which, to restate, is because we depend heavily on that *feature* and it needs to not have bugs like the one I squashed recently. The error log actually provides value to the consumer and therefore is a requirement.

docs/enhancements/error-logging-improvement.md
chuckha: The point I was trying to make here was that we are logging to the error log in various ways. Sometimes in callback, sometimes in groups, sometimes directly. Ideally we log to the error log in one way and not many. There is a place where we collect an errorgroup (which is good) but we also, while we're collecting errors, log to the error log. It's *so* easy to make a mistake this way.

chuckha:
@timothysc happy to reword or close, let me know what you'd like to see here.

I think this speaks to a specific manifestation of our general dependency managing issue in the code base. I can make this wording more generic, but my main focus is improving error logging.

PR #479, opened 2018-07-11
Autodetect version

### LINKED ISSUE #444: RFE: Autodetect kubernetes version and set kube-conformance image accordingly
FEATURE REQUEST

**What happened**:
I run Sonobuoy Scanner tool and Sonobuoy CLI on a k8s v1.8.5, and v1.8.13 cluster. The scans all result in the same failed e2e conformance tests verified by viewing the e2e.log contained inside the tarball.

All of the e2e conformance tests to pass if the cluster is v1.8+   Would like for the Sonobuoy scanner to auto-detect what version of kube-conformance container that it needs to run.

NAME=test-cluster.k8s.local
kops create cluster \
    --node-count 3 \
    --zones us-east-1a,us-east-1b,us-east-1c \
    --master-zones us-east-1a,us-east-1b,us-east-1c \
    --node-size t2.medium \
    --master-size t2.medium \
    --networking weave \
    --image 383156758163/k8s-1.8-debian-jessie-amd64-hvm-ebs-2018-05-27 \
    --kubernetes-version 1.8.5 \
    ${NAME}

Run the Sonobuoy Scanner tool (browser-based) and save logs.
Run the Sonobuoy CLI (./sonobuoy run) and save logs.

The Problem persisted when upgrading from k8s v.1.8.5 to 1.8.13 patch. However, after upgrading to k8s v1.9.6, the e2e conformance tests all succeed.
The customer stated that the scans did not fail in the past.

**Impact on the customer**:
The customer was not sure if there was an actual problem in the cluster.  

**ZenDesk ticket ID**:  ticket 203

**ZenDesk ticket priority/urgency**: Normal

**Environment**:
- Kubernetes version (use `kubectl version`): 1.8.5
- Cloud provider or hardware configuration: AWS
- OS (e.g. from /etc/os-release): Debian GNU/Linux 8 (jessie)
- Kernel (e.g. `uname -a`): 4.4.133-k8s
[201806181429_sonobuoy_14d33a51-d436-4e5c-a9e7-96aeba55354f.tar.gz](https://github.com/heptio/sonobuoy/files/2115621/201806181429_sonobuoy_14d33a51-d436-4e5c-a9e7-96aeba55354f.tar.gz)

liztio:
Sonobuoy will now detect the server's version and try to run the appropriate conformance container

- Fixes #444 

**Release note**:
```
sonobuoy run automatically detects server version, gen uses latest

--kube-conformance-image-version has several possible arguments:
auto queries the server version and uses that
latest uses the `latest` tag
a literal version (e.g. v1.10.2) always uses that
```

### DIFF
--- a/cmd/sonobuoy/app/gen.go
+++ b/cmd/sonobuoy/app/gen.go
@@ -23,27 +23,30 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/heptio/sonobuoy/pkg/client"
+	"github.com/heptio/sonobuoy/pkg/config"
 	"github.com/heptio/sonobuoy/pkg/errlog"
 )
 
 type genFlags struct {
-	sonobuoyConfig       SonobuoyConfig
-	mode                 client.Mode
-	rbacMode             RBACMode
-	kubecfg              Kubeconfig
-	e2eflags             *pflag.FlagSet
-	namespace            string
-	sonobuoyImage        string
-	kubeConformanceImage string
-	imagePullPolicy      ImagePullPolicy
+	sonobuoyConfig              SonobuoyConfig
+	mode                        client.Mode
+	rbacMode                    RBACMode
+	kubecfg                     Kubeconfig
+	e2eflags                    *pflag.FlagSet
+	namespace                   string
+	sonobuoyImage               string
+	kubeConformanceImage        string
+	kubeConformanceImageVersion ConformanceImageVersion
+	imagePullPolicy             ImagePullPolicy
 }
 
 var genflags genFlags
 
-func GenFlagSet(cfg *genFlags, rbac RBACMode) *pflag.FlagSet {
+func GenFlagSet(cfg *genFlags, rbac RBACMode, version ConformanceImageVersion) *pflag.FlagSet {
 	genset := pflag.NewFlagSet("generate", pflag.ExitOnError)
 	AddModeFlag(&cfg.mode, genset)
 	AddSonobuoyConfigFlag(&cfg.sonobuoyConfig, genset)
@@ -55,6 +58,7 @@ func GenFlagSet(cfg *genFlags, rbac RBACMode) *pflag.FlagSet {
 	AddNamespaceFlag(&cfg.namespace, genset)
 	AddSonobuoyImage(&cfg.sonobuoyImage, genset)
 	AddKubeConformanceImage(&cfg.kubeConformanceImage, genset)
+	AddKubeConformanceImageVersion(&cfg.kubeConformanceImageVersion, genset, version)
 
 	return genset
 }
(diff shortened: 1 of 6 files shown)


cmd/sonobuoy/app/gen.go
```diff
@@ -65,14 +69,46 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
 		return nil, errors.Wrap(err, "could not retrieve E2E config")
 	}
 
+	kubeclient, kubeError := maybeGetClient(&g.kubecfg)
+
+	rbacEnabled, err := genflags.rbacMode.Enabled(kubeclient)
+	if err != nil {
+		if errors.Cause(err) == ErrRBACNoClient {
```
chuckha: This type of error checking is a bit fragile and also requires errors to be exported for testing. An alternative approach to consider is defining a behavior on the custom error and then type checking to see if the returned error implements that behavior:

https://golang.org/src/os/error.go#L22

This makes it easy to create an error that implements that behavior without having to do a string comparison.

cmd/sonobuoy/app/gen.go
```diff
@@ -65,14 +69,46 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
...
+			return nil, errors.Wrap(err, kubeError.Error())
+		}
+		return nil, err
+	}
+
+	var discoveryClient discovery.ServerVersionInterface
+	var image string
+
+	if g.kubeConformanceImage != "" {
```
chuckha: This part could be at the very top, if kubeConformanceImage is defined, return the config. Otherwise, we can do the client song and dance and then return the config. This would drop the indentation level, be easier to read, and remove some of the error checking here I think.

cmd/sonobuoy/app/gen.go
```diff
@@ -65,14 +69,46 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
...
+	} else {
+		if kubeclient != nil {
+			discoveryClient = kubeclient.DiscoveryClient
+		}
+
+		imageVersion, err := g.kubeConformanceImageVersion.Get(discoveryClient)
+		if err != nil {
+			if errors.Cause(err) == ErrImageVersionNoClient {
+				return nil, errors.Wrap(err, kubeError.Error())
+			} else {
```
chuckha: drop the `else`, it's very rarely used in go

cmd/sonobuoy/app/gen.go
```diff
@@ -65,14 +69,46 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
 		return nil, errors.Wrap(err, "could not retrieve E2E config")
 	}
 
+	kubeclient, kubeError := maybeGetClient(&g.kubecfg)
```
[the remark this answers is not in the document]
liztio: I know it's unusual, but there's not much choice. The client is only required in some situtations, so erroring out when the configuration isn't valid is the wrong behaviour. I'm pretty committed to having `sonobuoy gen` work without access to a cluster, and I don't know any other ways to support this.

cmd/sonobuoy/app/gen.go
liztio: We also need the client song and dance for RBAC detection.

cmd/sonobuoy/app/gen.go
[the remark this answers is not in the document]
timothysc: +1 re naming getClient.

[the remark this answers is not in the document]
timothysc: The logic flow below this can be reduced based on whether there is a client and error below.

pkg/config/config.go
```diff
@@ -30,7 +30,7 @@ const (
 	// DefaultNamespace is the namespace where the master and plugin workers will run (but not necessarily the pods created by the plugin workers).
 	DefaultNamespace = "heptio-sonobuoy"
 	// DefaultKubeConformanceImage is the URL of the docker image to run for the kube conformance tests.
-	DefaultKubeConformanceImage = "gcr.io/heptio-images/kube-conformance:latest"
+	DefaultKubeConformanceImage = "gcr.io/heptio-images/kube-conformance:%s"
```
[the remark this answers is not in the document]
timothysc: s/const/var

cmd/sonobuoy/app/gen.go
```diff
@@ -65,14 +69,45 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
 		return nil, errors.Wrap(err, "could not retrieve E2E config")
 	}
 
+	kubeclient, kubeError := getClient(&g.kubecfg)
+
+	rbacEnabled, err := genflags.rbacMode.Enabled(kubeclient)
```
timothysc: Comment on the semantics here and open an issue on auto-detection then lgtm.

cmd/sonobuoy/app/gen.go
chuckha: ughhhhh yeah true

PR #485, opened 2018-07-12
use 'Enable' vs 'Enabled' for RBAC mode

jramos:
The usage information for `--rbac` is incorrect. It specifies `Enable` instead of `Enabled` as the correct argument value. This updates the `EnabledRBACMode` value from `Enabled` to `Enable` to maintain consistency with the `Disable` and `Detect` RBAC modes.


liztio:
Hi @jramos! Thank you so much for your PR.

 Could you amend your PR with --sign-off? `git commit --amend --sign-off` should do it, then force push. That's how Heptio does our contributor agreement.

jramos:
@liztio -- sure, done.

PR #486, opened 2018-07-13
Sets version image to use vMajor.Minor b/c we don't match all .Patch releases

timothysc:
Bug fix on run b/c we don't release *.patch versions

See https://github.com/heptio/kube-conformance/issues/25

**Release note**:
```
None
```

/assign @chuckha @liztio

### DIFF
--- a/cmd/sonobuoy/app/imageversion.go
+++ b/cmd/sonobuoy/app/imageversion.go
@@ -17,6 +17,7 @@ limitations under the License.
 package app
 
 import (
+	"fmt"
 	"strings"
 
 	version "github.com/hashicorp/go-version"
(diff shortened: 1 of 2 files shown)


cmd/sonobuoy/app/imageversion.go
```diff
@@ -79,7 +80,10 @@ func (c *ConformanceImageVersion) Get(client discovery.ServerVersionInterface) (
...
 		}
 
-		return version.GitVersion, nil
+		// NOTE: Until the kube-conformance container is pushed upstream we can't
+		// guarantee alignment with exact versioning see https://github.com/heptio/kube-conformance/issues/25
+		// for more details
+		return fmt.Sprintf("v%s.%s", version.Major, version.Minor), nil
 	}
 	return string(*c), nil
 }
```
liztio: Do we want to constrain validateVersion to minor releases too? Or is that fine as is

timothysc: It was fine for validation so I didn't change it.  Ideally I'd like to move the code back to your original PR in the future if we can push upstream to publish the testing container per release.

PR #489, opened 2018-07-18
add arm support for sonobuoy

### LINKED ISSUE #181: ARM versions available ?
Do you know if ARM images exist ?
I try that but it fails : 
```
$ docker pull gcr.io/heptio-images/sonobuoy-arm:v0.10.0
Error response from daemon: manifest for gcr.io/heptio-images/sonobuoy-arm:v0.10.0 not found
```

lubinsz:
Signed-off-by: Bin Lu <bin.lu@arm.com>

Add arm support

- Fixes #181

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,12 +13,14 @@
 # limitations under the License.
 #
 
+BINARY = sonobuoy
 TARGET = sonobuoy
 GOTARGET = github.com/heptio/$(TARGET)
 REGISTRY ?= gcr.io/heptio-images
 IMAGE = $(REGISTRY)/$(TARGET)
 DIR := ${CURDIR}
 DOCKER ?= docker
+LINUX_ARCH := amd64 arm64
 
 GIT_VERSION ?= $(shell git describe --always --dirty)
 IMAGE_VERSION ?= $(shell git describe --always --dirty)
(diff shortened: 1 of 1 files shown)


lubinsz:
@chuckha 
Can you check it？ 
Thanks.

liztio:
@lubinsz there's several components to running sonobuoy in arm. One is the binaries that are used for say `sonobuoy run`. Those aren't much trouble, you can compile them yourselves, or we could add them to goreleaser. Then there's the Docker images that Sonobuoy uses to run the tests and aggregate the results. Those will need to be build as [multi-arch docker images](https://container-solutions.com/multi-arch-docker-images/). Lastly, all the images the tests rely on, including [`kube-conformance`](https://github.com/heptio/kube-conformance) and [all the images run as part of the Kubernetes E2E tests](https://github.com/kubernetes/kubernetes/blob/master/test/utils/image/manifest.go#L51-L99). Those all need to be updated to be multi-arch as well.

This PR will only cause the binaries for those arm files to be built then discarded. Unfortunately, adding arm support is more involved. Feel free to drop by #sonobuoy on the kubernetes slack to chat about it!

lubinsz:
@liztio 
I see. 
This pr is only  a part of the patches for  #181
I will deliver other patches for it.

PR #491, opened 2018-07-19
push images with multi-arch to support amd64/arm64

### LINKED ISSUE #181: ARM versions available ?

lubinsz:
Signed-off-by: Bin Lu <bin.lu@arm.com>

add arm64 support 

- Fixes #181

**Release note**:
```
release-note
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.10-alpine
+FROM BASEIMAGE
 MAINTAINER Timothy St. Clair "tstclair@heptio.com"
 
-RUN apk add --no-cache ca-certificates bash
-ADD sonobuoy /sonobuoy
+CMD1
+
+ADD BINARY /sonobuoy
 ADD scripts/run_master.sh /run_master.sh
 ADD scripts/run_single_node_worker.sh /run_single_node_worker.sh
 WORKDIR /
--- a/Makefile
+++ b/Makefile
@@ -12,13 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+EMPTY :=
+SPACE := $(EMPTY) $(EMPTY)
+COMMA := $(EMPTY),$(EMPTY)
 
+BINARY = sonobuoy
 TARGET = sonobuoy
 GOTARGET = github.com/heptio/$(TARGET)
+GOPATH = $(shell go env GOPATH)
 REGISTRY ?= gcr.io/heptio-images
 IMAGE = $(REGISTRY)/$(TARGET)
 DIR := ${CURDIR}
 DOCKER ?= docker
+LINUX_ARCH := amd64 arm64
+DOCKERFILE :=
+PLATFORMS := $(subst $(SPACE),$(COMMA),$(foreach arch,$(LINUX_ARCH),linux/$(arch)))
 
 GIT_VERSION ?= $(shell git describe --always --dirty)
 IMAGE_VERSION ?= $(shell git describe --always --dirty)
@@ -30,8 +38,6 @@ VERBOSE_FLAG = -v
 endif
 BUILDMNT = /go/src/$(GOTARGET)
 BUILD_IMAGE ?= golang:1.10-alpine
-BUILDCMD = CGO_ENABLED=0 go build -o $(TARGET) $(VERBOSE_FLAG) -ldflags "-X github.com/heptio/sonobuoy/pkg/buildinfo.Version=$(GIT_VERSION)"
-BUILD = $(BUILDCMD) $(GOTARGET)
 
 TESTARGS ?= $(VERBOSE_FLAG) -timeout 60s
 TEST_PKGS ?= $(GOTARGET)/cmd/... $(GOTARGET)/pkg/...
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -91,6 +127,31 @@ push:
 		$(DOCKER) push $(REGISTRY)/$(TARGET):latest; \
 	fi
 
+push_manifest:
+	/go/bin/manifest-tool push from-args --platforms $(PLATFORMS) --template $(REGISTRY)/$(TARGET)-ARCH:$(VERSION) --target  $(REGISTRY)/$(TARGET):$(VERSION)
+
+push: pre container
+	for arch in $(LINUX_ARCH); do \
+		$(MAKE) push_images TARGET="sonobuoy-$$arch"; \
```
liztio: `An image does not exist locally with the tag: gcr.io/heptio-images/sonobuoy-amd64`

When should these be created?

Dockerfile
```diff
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.10-alpine
+FROM BASEIMAGE
```
liztio: I think seperate Dockerfiles would be easier to reason about

Makefile
```diff
@@ -71,17 +76,48 @@ lint:
...
+       -t $(REGISTRY)/$(TARGET):$(IMAGE_BRANCH) \
+       -t $(REGISTRY)/$(TARGET):$(GIT_REF) \
+       -f $(DOCKERFILE) \
 		.
 
-sonobuoy:
-	$(DOCKER_BUILD) '$(BUILD)'
+container: sonobuoy
+	for arch in $(LINUX_ARCH); do \
+		if [ $$arch = amd64 ]; then \
```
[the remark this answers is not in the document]
lubinsz: hi @liztio 
In my opinion, inconvenient maintenance in the later stage if use seperate Dockerfiles.
For example: if we need to add a new feature in Dockerfile, we should modify them twice.
3 differences between the two substitutions:
1, baseimage
2, command: In x86, we can't use any 'RUN' command in Dockerfile to build target for CROSS-platform(arm, power). 
3, binary file: different binary files correspond to different platforms

Dockerfile
lubinsz: Inconvenient maintenance in the later stage if use seperate Dockerfiles as I think.
Also, same approach in k8s project: single Dockerfile, and use 'sed' to change it for each platform.

Makefile
```diff
@@ -71,17 +76,48 @@ lint:
...
+			sed 's|BASEIMAGE|alpine:3.7|g' Dockerfile > Dockerfile-$$arch; \
```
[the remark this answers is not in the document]
lubinsz: Hi, 
I have fixed it. Please check it.

Makefile
```diff
@@ -91,6 +127,31 @@ push:
 		$(DOCKER) push $(REGISTRY)/$(TARGET):latest; \
 	fi
 
+push_manifest:
+	/go/bin/manifest-tool push from-args --platforms $(PLATFORMS) --template $(REGISTRY)/$(TARGET)-ARCH:$(VERSION) --target  $(REGISTRY)/$(TARGET):$(VERSION)
```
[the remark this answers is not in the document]
lubinsz: Hi,
I have fixed it. Please check it.

Makefile
```diff
@@ -12,26 +12,33 @@
...
+LINUX_ARCH := amd64 arm64
+DOCKERFILE :=
+PLATFORMS := $(subst $(SPACE),$(COMMA),$(foreach arch,$(LINUX_ARCH),linux/$(arch)))
 
 GIT_VERSION ?= $(shell git describe --always --dirty)
 IMAGE_VERSION ?= $(shell git describe --always --dirty)
 IMAGE_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD | sed 's/\///g')
 GIT_REF = $(shell git rev-parse --short=8 --verify HEAD)
 
+GOBIN = /go/bin
```
liztio: I don't want to need to be root to run this script. `GOBIN ?= /go/bin` maybe? Are you building this on a machine that doesn't have go set up?

lubinsz: hi Liztio,
I have updated it, please check it.
Thanks.

lubinsz:
@liztio 
Any other question?

Makefile
```diff
@@ -71,17 +81,49 @@ lint:
 vet:
 	$(DOCKER_BUILD) '$(VET)'
 
-container: sonobuoy
+pre:
+	curl -sSL https://github.com/estesp/manifest-tool/releases/download/$(MANIFEST_TOOL_VERSION)/manifest-tool-linux-amd64 > $(MANIFEST_TOOL_DIR)/manifest-tool
```
liztio: This doesn't work on OSX: `/var/folders/v9/30xb9zqj20nf56xbvmz6qpym0000gn/T/tmp.cdp3AjCw/manifest-tool: /var/folders/v9/30xb9zqj20nf56xbvmz6qpym0000gn/T/tmp.cdp3AjCw/manifest-tool: cannot execute binary file`

Why did you switch from `go get`?

lubinsz: @liztio 
"go get" is back. Please check it.

liztio:
Hi there @lubinsz! It took me a little bit of fiddling this morning but I managed to get a test push to work. Because I work from osx, I needed to make the following change:

```diff
diff --git a/Makefile b/Makefile
index 40a982d..965c831 100644
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ push_images:
        fi

 push_manifest:
-       $(GOPATH)/bin/manifest-tool push from-args --platforms $(PLATFORMS) --template $(REGISTRY)/$(TARGET)-ARCH:$(VERSION) --target  $(REGISTRY)/$(TARGET):$(VERSION)
+       $(GOPATH)/bin/manifest-tool -username oauth2accesstoken --password "`gcloud auth print-access-token`" push from-args --platforms $(PLATFORMS) --template $(REGISTRY)/$(TARGET)-ARCH:$(VERSION) --target  $(REGISTRY)/$(TARGET):$(VERSION)

 push: pre container
        for arch in $(LINUX_ARCH); do \
```

If you could add that, we should be good to add this!

lubinsz:
@liztio 
Done.

liztio:
Sorry for all the back and forth! We should be cutting v1.11.5 soon, and hopefully that'll be the first release that supports arm!

kalyxin02:
@liztio Could we know when will the v1.11.5 release happen? Thanks!

liztio:
@kalyxin02 we recently release v1.11.6  due to an issue with .5, it should be live now, give it a go!

johnSchnake:
@lubinsz I'm just looking into the makefile/images a bit since we need to update them for #609 

Why did we need to use different base images at all? From my short trial, it seems like we can just use the basic `GOOS=<> GOARCH=<> go build` and the binary created from my mac was even able to run on an `arm64v8:busybox` container.

I'm just wondering if I'll only notice the need in a certain situation; I dont actually have an arm64 device to actually test on either.

PR #500, opened 2018-07-23
Quick fix for an empty config SEGV on nil rest.Config bug

### LINKED ISSUE #498: Sonobuoy 0.11.4 panics when analyzing results tarball
I ran `sonobuoy run` against a Kubernetes 1.11.1 cluster on CentOS 7.4 on AWS. After `sonobuoy status` reported the tests complete, I retrieved the results using `sonobuoy retrieve`. Upon attempting to analyze the results using `sonobuoy e2e <archive>`, Sonobuoy panics and exits.

I expected Sonobuoy to report on the results of the tests.

I've reproduced this error using the Linux build of Sonobuoy 0.11.4 on both Fedora 27 and CentOS 7.4. I have not yet tested the macOS build.

**Environment:**

- Sonobuoy version: 0.11.4
- Kubernetes version: (use `kubectl version`): 1.11.1
- Kubernetes installer & version: `kubeadm` 1.11.1
- Cloud provider or hardware configuration: AWS (no cloud provider configured)
- OS (e.g. from `/etc/os-release`): CentOS Linux 7
[201807212345_sonobuoy_bd9a041a-8a5a-4885-aab8-719bd67786ac.tar.gz](https://github.com/heptio/sonobuoy/files/2219369/201807212345_sonobuoy_bd9a041a-8a5a-4885-aab8-719bd67786ac.tar.gz)

timothysc:
Fixes a condition where getSonobuoyClient(cfg *rest.Config) can be called with a nil cfg 

Which is a normal condition that can occur for offline tarball analysis.

- Fixes #498

none

**Release note**:
```
None
```

/assign @chuckha 
/cc @scottslowe

### DIFF
--- a/cmd/sonobuoy/app/common.go
+++ b/cmd/sonobuoy/app/common.go
@@ -9,9 +9,13 @@ import (
 )
 
 func getSonobuoyClient(cfg *rest.Config) (*client.SonobuoyClient, error) {
-	skc, err := sonodynamic.NewAPIHelperFromRESTConfig(cfg)
-	if err != nil {
-		return nil, errors.Wrap(err, "couldn't get sonobuoy api helper")
+	var skc *sonodynamic.APIHelper
+	var err error
+	if cfg != nil {
+		skc, err = sonodynamic.NewAPIHelperFromRESTConfig(cfg)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't get sonobuoy api helper")
+		}
 	}
 	return client.NewSonobuoyClient(cfg, skc)
 }


chuckha:
I'm ok with this as a quick fix

I'm starting to think about two sonobuoy clients. one that interacts with stuff locally gen (sometimes) and e2e that do not require a kubeconfig and one that interacts with kubernetes, run, status, also gen, etc.

The only other thing I'll suggest is that this go directly in https://github.com/heptio/sonobuoy/blob/master/pkg/dynamic/client.go to return a local-only sonobuoy client.

Issue #507, opened 2018-07-24
Sonobuoy failing to test clusters with custom taints on some nodes.


mauilion:
**What steps did you take and what happened:**

I have a set of nodes that have the following taints:
```
terraform-az1-0
  effect:NoExecute key:node-role.kubernetes.io/etcd
  effect:NoSchedule key:node-role.kubernetes.io/etcd
terraform-az1-1
  effect:NoSchedule key:node-role.kubernetes.io/master]
terraform-az1-2
terraform-az2-0
  effect:NoExecute key:node-role.kubernetes.io/etcd
  effect:NoSchedule key:node-role.kubernetes.io/etcd
terraform-az2-1
  effect:NoSchedule key:node-role.kubernetes.io/master
terraform-az2-2
terraform-az3-0
  effect:NoExecute key:node-role.kubernetes.io/etcd
  effect:NoSchedule key:node-role.kubernetes.io/etcd
terraform-az3-1
  key:node-role.kubernetes.io/master effect:NoSchedule
terraform-az3-2
```

When I run sonobuoy against this cluster it fails the tests.
In debugging I can see that part of this is related to the fact that not all nodes are addressable with the default tolerations of:

```
      tolerations:
      - effect: NoSchedule
        key: node-role.kubernetes.io/master
        operator: Exists
      - key: CriticalAddonsOnly
        operator: Exists
```

I've tried to work around the issue by specifying a toleration of 

```
  tolerations:
  - operator: "Exists"
```
in the sonobuoy.yaml file. example here: https://gist.github.com/mauilion/3e13dfbf649cc4e30fa56ea2fd258b12

**What did you expect to happen:**
I expected that either the default tolerations would be very permissive. Or that when provided an additional set of tolerations the provided tolerations would be merged into the spec deployed by sonobuoy.

**Environment:**

- Sonobuoy version: v0.11.4
- Kubernetes version: (use `kubectl version`):
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.1", GitCommit:"b1b29978270dc22fecc592ac55d903350454310a", GitTreeState:"clean", BuildDate:"2018-07-17T18:53:20Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #510, opened 2018-07-26
When inspecting another changeset I found our internal template to be old

timothysc:
Our internal daemonset templates are really old, and I was surprised it worked.  

This partially addresses overlaps with #507 as well.

I need to test this some more to verify. 

**Release note**:
```
NONE
```

/assign @liztio

### DIFF
--- a/pkg/discovery/queries.go
+++ b/pkg/discovery/queries.go
@@ -27,7 +27,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


timothysc:
@mauilion @joshrosso @carldanley - I don't recall is a sonobuoy scan pre-upgrade part of the workflow.  If so I may augment this change, to support 1.8 just-in-case.

timothysc:
I looked over instructions and the answer was no.

pkg/discovery/queries.go
```diff
@@ -118,41 +117,13 @@ func queryNsResource(ns string, resourceKind string, opts metav1.ListOptions, ku
 	case "ConfigMaps":
 		return kubeClient.CoreV1().ConfigMaps(ns).List(opts)
 	case "ControllerRevisions":
-		lst, err := kubeClient.AppsV1().ControllerRevisions(ns).List(opts)
```
liztio: Getting rid of this feels good :+1:

PR #539, opened 2018-10-03
Add SSH capabilities to CLI

### LINKED ISSUE #536: Allow SSH keys to be put into the plugin containers via CLI
**Motivation**
There are numerous e2e tests which require SSH keys to be present, otherwise they are skipped. There should be a way from the CLI to run those tests, given that you have the appropriate private key available.

`sonobuoy run --e2e-focus SSH --private-key=./path/to/private_key`

This can currently be achieved manually but has a few downsides.
Process:
 -  Run `sonobuoy gen` with your focus/skip/etc options to generate a yaml file for deployment.
 - add a secret to the namespace which you have to manually base64 encode
      OR create the namespace and secret manually via kubectl first and remove the namespace from the yaml so you dont duplicate it.
 - add the volume and volume mount to the e2e pod

**Environment:**
I've done these steps manually on Mac and Linux; mainly against a cluster that was in Openstack but nothing is OS/platform/version specific. There is a tiny difference when creating the secret when you are base64 encoding if you are using the OS tools though since Mac/Linux have different default behavior/flags for newlines (see https://kubernetes.io/docs/concepts/configuration/secret/)

I am going to put up a PR to try and automate this but please let me know about agreement with this at a high level or if there are any nuances I need to be aware of.

johnSchnake:
- Add ssh-key and ssh-user flags to CLI to specify
a key/user to SSH to containers in the cluster.
 - Use k8s secrets to store the secret then mount
into the containers
 - Specify env vars so different cloud providers
can find the key at the intended location.

Adds flags and gen support for SSH keys/user so that Kubernetes E2E tests that require SSH keys can execute.

- Fixes #536

I added a new test utilizing golden files for `gen.go`. There was one existing test which was looking for one specific component by decoding everything and discarding all but the data of interest. Since the data is varied in type and extensive, using golden files seemed to be the most straight forward way to test the `gen.go` functionality and I think a follow up ticket could even add more tests of this nature (there was a todo in the code regarding testing more parts of the manifest and the current object oriented approach would be too verbose to execute well IMO).

Per the norm with golden files, I added an `update` flag to the test so that you could update the golden files by running the tests with the new flag.

Lastly, I added multiple env vars for the SSH key location which actually behave a little differently. In some cases it uses just the base name for the file (id_rsa) but in others it uses the full path (/root/.ssh/id_rsa). The reason for this is that Kubernetes currently treats the keys differently on different providers. I have a ticket/PR to fix this behavior as well and if it merges then we could update this to also be consistent. (See https://github.com/kubernetes/kubernetes/issues/68687 )

**Release note**:
```
release-note
New flags added to "sonobuoy run": "ssh-user" and "ssh-key" which enable embedding your SSH key into the container so that tests requiring SSH can be run.
```


johnSchnake:
@ncdc @timothysc So there was a unit test I added (notes in the PR message) but I also wanted to mention that I ran this manually with the following command:
```sonobuoy run --e2e-focus="should propagate mounts to the host" --ssh-user=<ssh key user> --ssh-key <path to my key> --sonobuoy-image gcr.io/heptio-images/sonobuoy:v0.11.6```

Had to used the tagged image and based my changes from the last release since there was an issue running from master (known issue): https://github.com/heptio/sonobuoy/issues/538

**Comment by timothysc** [comment]
@liztio they are linked in the issue he referenced.  I really don't like it, but it's what is there today.

**Comment by johnSchnake** [comment]
@liztio One of the simplest is: `should propagate mounts to the host`; that will run and can pass just fine. If you choose an SSH one at random it may pass/fail based on your setup. I have another k8s issue/pr up because the SSH tests are inconsistent with whether or not they require externalIPs to be specified on the nodes.

File: pkg/templates/manifest.go [hunk 1]
```diff
@@ -97,8 +120,19 @@ data:
...
         name: results
         readOnly: false
+{{- if .SSHKey }}
+      - mountPath: /root/.ssh
+        name: sshkey-vol
+{{- end}}
       tolerations:
         - operator: "Exists"
+{{- if .SSHKey }}
+    extra-volumes:
```
liztio: I think this should be `volumes`.

pkg/templates/manifest.go
```diff
@@ -97,8 +120,19 @@ data:
...
+    - name: sshkey-vol
+      secret:
+        secretName: ssh-key
+        defaultMode: 256
```
liztio: this is a weird mode. `--w-r-xrw-`. I don't think this'll work, usually `600` is the right mode for SSH keys.

liztio:
I ran this with `sonobuoy run --e2e-focus "should propagate mounts to the host" --ssh-user ubuntu --ssh-key ~/.ssh/id_rsa --sonobuoy-image gcr.io/heptio-images/sonobuoy:v0.11.0`. It started up the sonobuoy container, but then it error'd out:

```
time="2018-10-04T15:28:56Z" level=info msg="Running plugin" plugin=e2e
time="2018-10-04T15:28:56Z" level=info msg="starting aggregation server" address=0.0.0.0 port=8080
time="2018-10-04T15:28:56Z" level=error msg="error running plugins: error running plugin e2e: could not create Job resource for Job plugin e2e: Pod \"sonobuoy-e2e-job-cd26497c7474458d\" is invalid: spec.containers[0].volumeMounts[1].name: Not found: \"sshkey-vol\""
```

I've made some comments on the PR with what I think the issues are.

pkg/templates/manifest.go
johnSchnake: This is for the sonobuoy manifest though so it seems to go through this intermediate `extra-volumes` field that you've created. https://github.com/heptio/sonobuoy/blob/master/docs/plugins.md references this same field for adding more volumes to the plugins.

pkg/templates/manifest.go
johnSchnake: Its a little wonky in the documentation because they go back and forth but it seems to require decimal notation:
> Note that the JSON spec doesn’t support octal notation, so use the value 256 for 0400 permissions. If you use yaml instead of json for the pod, you can use octal notation to specify permissions in a more natural way.

If you look at their examples though, even when using yaml they use the decimal notation (so 256->0400 which is sufficient for this purpose)

https://kubernetes.io/docs/concepts/configuration/secret/

pkg/templates/manifest.go
liztio: ugh, right you are. sorry, it's been a while since I touched this stuff

pkg/templates/manifest.go
liztio: huh, TIL

johnSchnake:
@liztio Hm, I'm unsure of why you'd get that error. Are there more logs that you could send to help debug this? 

It seems to be saying that the volume itself (which is based on the secret) does not exist so I'm wondering if the secret failed to be created or if just the `extra-volume` failed. Like the comments above indicated, the syntax seems to be correct (and works for me) to add the extra volume and mount it.

I've tried:
 - running from my own repo/branch
 - running from heptio repo and checking out this PR
 - using an absolute path for the key
 - using a path referencing `~` like your example

I am using `make build_sonobuoy` and using that binary which is for linux so I'm running it inside another docker container. Does that mirror what you're doing or are you building for mac perhaps?

I don't think the yaml/code is the issue though since the unit test (using a goldenfile) works. That just leaves something about the clusters that is behaving differently which we can only see from the logs.

liztio:
Ohh, I know what I did wrong. I forgot I needed to build a container for use remotely too, I was just compiling remotely. I'm happy to approve this now.

Issue #437, opened 2018-06-05
Support new parameter --wait-until-finish


openstacker:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST
**What happened**:
Currently, user has to run command like this 
 `sonobuoy --kubeconfig /path/to/kubeconfig run && sleep 5200 && sonobuoy retrieve results && sonobuoy e2e results/*.tar.gz`

**What you expected to happen**:
As a user, I'd like to have a parameter like --wait-until-finish, so that I can run sonobuoy command like this
`sonobuoy --kubeconfig /path/to/kubeconfig --wait-until-finish run`

In this case, I'd like to see the command will print out the output based on the cases it just finished until all done. 

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Sonobuoy tarball (which contains * below)
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):

PR #557, opened 2018-11-06
Adding a runner for automating sonobuoy runs.

TrevinTeacutter:
We were wanting to automate sonobuoy runs. So I added a "runner" command that does a few things:
* Kicks off a `sonobuoy run`
* Polls `sonobuoy status` for completion of tests
* If status is bad, or something wrong happened, runs `sonobuoy delete`
* If status is good and the run is complete, will try and retrieve results and output them to `stdout` since at least for us, we are expecting to have our logging solution grab them and persist them.

I figured we wouldn't be the only ones interested in automating this process a bit more so rather than maintain a fork I'm submitting upstream.

### DIFF
--- a/cmd/sonobuoy/app/runner.go
+++ b/cmd/sonobuoy/app/runner.go
@@ -1,0 +1,177 @@
+package app
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/heptio/sonobuoy/pkg/client"
+	"github.com/heptio/sonobuoy/pkg/errlog"
+	"github.com/heptio/sonobuoy/pkg/plugin/aggregation"
+)
+
+type runnerFlags struct {
+	genFlags
+	showAll         bool
+	skipPreflight   bool
+	pollingInterval time.Duration
+	statusLoopDelay time.Duration
+	retrievalDelay  time.Duration
+	timeout         time.Duration
+}
+
+var (
+	runnerflags    runnerFlags
+	sonobuoyClient *client.SonobuoyClient
+)
+
+func RunnerFlagSet(cfg *runnerFlags) *pflag.FlagSet {
+	runnerset := pflag.NewFlagSet("runner", pflag.ExitOnError)
+	// Default to detect since we need kubeconfig regardless
+	runnerset.AddFlagSet(GenFlagSet(&cfg.genFlags, DetectRBACMode, ConformanceImageVersionAuto))
+	AddSkipPreflightFlag(&cfg.skipPreflight, runnerset)
+
+	runnerset.DurationVar(
+		&cfg.timeout, "runner-timeout", 6*time.Hour,
+		"Length of time to give the runner before giving up. (Default: 6h)",
+	)
+	runnerset.DurationVar(
+		&cfg.pollingInterval, "polling-interval", 5*time.Minute,
+		"Duration of time between polling sonobuoy for status. (Default: 5m)",
+	)
+	runnerset.BoolVar(
+		&cfg.showAll, "show-all", false,
+		"Don't summarize plugin statuses, show all individually",
+	)
+
+	return runnerset
+}
+
+func (r *runnerFlags) Config() (*client.RunConfig, error) {
+	gencfg, err := r.genFlags.Config()
+	if err != nil {
+		return nil, err
+	}
+	return &client.RunConfig{
+		GenConfig: *gencfg,
+	}, nil
+}
+
+func init() {
+	cmd := &cobra.Command{
+		Use:   "runner",
+		Short: "Submits a sonobuoy run and waits for it's completion",
+		Run:   startSonobuoyRunner,
+		Args:  cobra.ExactArgs(0),
+	}
+
+	cmd.Flags().AddFlagSet(RunnerFlagSet(&runnerflags))
+	RootCmd.AddCommand(cmd)
+}
+
+func startSonobuoyRunner(cmd *cobra.Command, args []string) {
+	var signals chan os.Signal
+	signals = make(chan os.Signal, 1)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+
+	cfg, err := runnerflags.kubecfg.Get()
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "couldn't get REST client"))
+		os.Exit(1)
+	}
+
+	runCfg, err := runnerflags.Config()
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "could not retrieve E2E config"))
+		os.Exit(1)
+	}
+	sonobuoyClient, err = getSonobuoyClient(cfg)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "could not create sonobuoy client"))
+		os.Exit(1)
+	}
+
+	err = run(runCfg)
+	if err != nil {
+		errlog.LogError(err)
+		os.Exit(1)
+	}
+
+	ticker := time.NewTicker(runnerflags.pollingInterval)
+	go func() {
+		for range ticker.C {
+			status, err := status()
+			if err != nil {
+				errlog.LogError(err)
+				continue
+			}
+
+			switch status.Status {
+			case aggregation.RunningStatus:
+				continue
+			case aggregation.FailedStatus:
+				errlog.LogError(errors.New("Aggregator has status of failed"))
+				os.Exit(1)
+			case aggregation.CompleteStatus:
+				fmt.Printf("Run completed")
+				os.Exit(0)
+			default:
+				errlog.LogError(errors.New("Unknown status for aggregator"))
+				os.Exit(1)
+			}
+		}
+	}()
+
+	time.Sleep(runnerflags.timeout)
+	ticker.Stop()
+	errlog.LogError(errors.New("Runner timed out waiting for tests to complete"))
+	os.Exit(1)
+}
+
+func run(runCfg *client.RunConfig) error {
+	plugins := make([]string, len(runCfg.Config.PluginSelections))
+	for i, plugin := range runCfg.Config.PluginSelections {
+		plugins[i] = plugin.Name
+	}
+
+	if len(plugins) > 0 {
+		fmt.Printf("Running plugins: %v\n", strings.Join(plugins, ", "))
+	}
+
+	if !runnerflags.skipPreflight {
+		if errs := sonobuoyClient.PreflightChecks(&client.PreflightConfig{Namespace: runnerflags.namespace}); len(errs) > 0 {
+			for _, err := range errs {
+				errlog.LogError(err)
+			}
+			return errors.New("Preflight checks failed")
+		}
+	}
+
+	if err := sonobuoyClient.Run(runCfg); err != nil {
+		return errors.Wrap(err, "error attempting to run sonobuoy")
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


timothysc:
@TrevinTeacutter Thank you for the patch!  Could you please signoff on your commit? 

I'm not sold on the UX of this approach 
/assign @liztio for review.

TrevinTeacutter:
Yeah, I don't have any opinions on the format of what it outputs to stdout so long as we get the json object into a single line since logging events are typically separated by new lines (or maybe that should be configurable in some way).

ncdc:
@timothysc since you pinged in slack: I'd like an easy way to kick off tests and wait up to a user-specified timeout for them to complete. I'm ok scripting retrieving the results afterwards, or including it as an option using a flag. If the tests don't complete within the timeout, it might be useful to have it retrieve whatever diagnostic data it can, but that might be too specific of a use case.

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+			errlog.LogError(err)
+			continue
+		}
+
+		switch status.Status {
+		case aggregation.FailedStatus:
+			signals <- syscall.SIGUSR1
+			return
+		case aggregation.CompleteStatus:
+			// Results aren't immediately available even though status says complete=
```
liztio: extra `=` here

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+			time.Sleep(runnerflags.retrievalDelay)
```
liztio: I don't really like this fixed timeout. Why not just make a couple attempts at retrieval until it succeeds?

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+
+		status, err := status()
+		if err != nil {
+			errlog.LogError(err)
+			continue
+		}
+
+		switch status.Status {
+		case aggregation.FailedStatus:
+			signals <- syscall.SIGUSR1
```
liztio: I don't love this. I get that passing in SIGUSR1 makes a lot of sense for the end user, but using it internally feels like a hack.

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+	if failed {
+		os.Exit(1)
+	}
+
+	go func() {
+		_, open := <-signals
+		delete()
+
+		if open {
+			close(signals)
```
liztio: why close here but not in the other case?

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+
+func RunnerFlagSet(cfg *runnerFlags) *pflag.FlagSet {
+	runnerset := pflag.NewFlagSet("runner", pflag.ExitOnError)
+	// Default to detect since we need kubeconfig regardless
+	runnerset.AddFlagSet(GenFlagSet(&cfg.genFlags, DetectRBACMode, ConformanceImageVersionAuto))
+	AddSkipPreflightFlag(&cfg.skipPreflight, runnerset)
+
+	runnerset.DurationVar(
+		&cfg.timeout, "runner-timeout", 6*time.Hour,
+		"Length of time to give the runner before giving up. (Default: 6h)",
```
liztio: where did these times come from?

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+			time.Sleep(runnerflags.pollingInterval - delta.Sub(time.Now()))
+		default:
+			errlog.LogError(errors.New("Unknown status for aggregator"))
+			signals <- syscall.SIGUSR1
+			return
+		}
+	}
+}
+
+func run(runCfg *client.RunConfig) bool {
```
liztio: why doesn't this just return an error, like the other methods?

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+
+	if err := sonobuoyClient.Run(runCfg); err != nil {
+		errlog.LogError(errors.Wrap(err, "error attempting to run sonobuoy"))
+		return true
+	}
+
+	return false
+}
+
+func retrieve(outDir string) bool {
```
liztio: same here, re:bool

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+
+	err = untar(gzipReader, printTarballResults)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func printTarballResults(header *tar.Header, tarReader io.Reader) error {
```
liztio: are you just printing the untarrred result to stdout?

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	return &client.RunConfig{
+		GenConfig: *gencfg,
+	}, nil
+}
+
+func init() {
+	cmd := &cobra.Command{
+		Use:   "runner",
+		Short: "Submits a sonobuoy run and manages the lifetime of that run",
```
liztio: I'd like to see a longer description here. What does 'manage the lifetime'  mean? what are the expected outputs?

cmd/sonobuoy/app/runner.go
TrevinTeacutter: Just some arbitrary time for how long some of our Sonobuoy runs have taken. If you have a better default timeout I'm definitely open for suggestions.

TrevinTeacutter:
I can add options for whether or not to delete, that's fine.
As for outputting files, my only requirement is that each JSON within the tarball be able to be outputted in some way to stdout where each json file would be treated as a separate event by a logging daemon or driver (such as fluentd), thus why in this case each file is separated by a single line.

cmd/sonobuoy/app/runner.go
TrevinTeacutter: Yes, with each file on a separate line so that is recognized as a single event by logging daemons/drivers. Our current idea is to just use logging as our way to persist results of successful runs.

cmd/sonobuoy/app/runner.go
BenTheElder: So then this is a race and if the sleep isn't long enough we don't try again ...? :-1: 
Kubernetes would retry with backoff in this situation.

cmd/sonobuoy/app/runner.go
BenTheElder: That sounds a bit terrible :confused: 
Kubernetes's own CI would not want to use this, we'd upload the tarball contents separately to persistent storage (GCS) rather than dump everything into the logs.

cmd/sonobuoy/app/runner.go
```diff
@@ -0,0 +1,330 @@
...
+
+		if open {
+			close(signals)
+			os.Exit(1)
+		} else {
+			os.Exit(0)
+		}
+	}()
+
+	time.Sleep(runnerflags.statusLoopDelay)
```
BenTheElder: if you use a switch and ticker instead it can be more responsive to signals

BenTheElder: (and cleanly exit / print something to the logs rather than just aborting on signal directly ..?)

BenTheElder:
IIRC the tarball should contain results that are not JSON, which are useful to retreive, why not make your retrieval and print mechanism a separate command that your tools call after waiting for the tests to finish? That's more focused / composable and makes it easy for other users to not dump the tarball to the logs, I don't suspect that to be a popular option.

TrevinTeacutter:
The problem is still there is no automated way to check for completion because Sonobuoy itself offers no other output mechanisms for any of it's commands other than plain text (meaning regex is the only way to get info on whether it passed or not because exit codes aren't useful here either). There is also no equivalent of a `rollout status` command to sonobuoy to avoid having to poll.

cmd/sonobuoy/app/runner.go
TrevinTeacutter: We do not have GCS, it would feel more apt to allow multiple ways to output the result to various backends and not just GCS. At current though we don't really have usage scenarios to even test adding those different persistence stores as backends. At the same time, someone within our team also brought up that it would be nice if we had a way to parse the results to know if there were failures within the runs without having to manually scan the json files.

TrevinTeacutter:
Here is the story that will maybe help understand why there is desire to make this all one command:
We want to automate runs of sonobuoy on our clusters, so that no human is starting them or sifting through results to ensure it passed all tests or not. If it fails of course human interaction would be needed (interaction depending on the failure) which means grabbing some summary to know what tests failed or grabbing the actual tarball which is persisted somehow.

I don't want to write some hacky bash script that is regexing the output of each sonobuoy run (if I had a choice, I'd much rather rely on jq'ing the results than doing magical regex) either because that makes the automation extremely fragile.

If you have recommendations outside of making this PR then I'll happily look into them.

cmd/sonobuoy/app/runner.go
BenTheElder: I don't think sonobuoy should upload to GCS, I think your automation should upload the output files to $storage ideally.

BenTheElder: As for checking if there are failures, the exit code of the command should be sufficient without dumping the results to the logs.

BenTheElder:
Most CI / testing is built around exit codes AFAIK.. so my suggested route:
- have the run command give a useful exit code
- let some user specific tool handle persisting the tarball

Then your script is just:
- obtain cluster
- run sonobuoy, save the exit code
- ask sonobuoy for the tarball, persist it via your preferred method
- exit with the code / output a string based on the exit code for your logging tools to find

Dumping the tarball to output seems like a huge hack, but also one that would be easy to implement in another tool.

TrevinTeacutter:
Is the run considered failed if something bad happened or if one of the e2e tests (or any other test) also failed? That seems....problematic because it still means someone has to manually pull that tarball and either run some separate (custom) command to interpret them or manually interpret them.

BenTheElder:
- we can use different exit codes for the failure modes as necessary
- your small tool can pull the tarball and dump it to logs after run exits.
It doesn't need to be part of run though

Needing to wait for the run to end seems common, needing to dump the full
results to logs less so. The latter is also especially trivial and less
error prone than cleanly blocking on the run.

TrevinTeacutter:
I'd say that is more of a result of how sonobuoy is sending results in a manner that isn't easily serialized into the client to process those results than it is because of processing the results itself as a part of the same command. I'll just get rid of the dumping story for now.

johnSchnake:
It seems that now that the delete and retrieve functionality have been removed, the main thing this does is wait for the finish.

This matches up with https://github.com/heptio/sonobuoy/issues/437 which has the more desirable UX of just adding a flag to the run itself. Under the hood it would still be polling for a terminal state but increases  the scriptability of sonobuoy without adding new runners and complicating the UX.

I'd be happy to get #437 included which seems to benefit you as much as this PR with a much smaller code/UX change. Let me know if you want to handle that (preferably in a new PR to clarify history/discussion).

TrevinTeacutter:
I have no preference, so if you wanna do the change that works for me.

johnSchnake:
https://github.com/heptio/sonobuoy/pull/592 has merged so you now have the option of doing:
```
sonobuoy run --wait && sonobuoy retrieve && sonobuoy delete --wait
```
which should run (up to a day), get the results, then delete everything.

Obviously you can add some more scripting in there if you want to otherwise handle errors/get logs etc.

My plan is to ensure that `sonobuoy retrieve` gives output about the files it created so that you can also easily get the tarball and run `sonobuoy e2e` to see the results. Once that is all done then I'll update the readme with the easier CLI flow.

PR #570, opened 2019-01-10
added path in stderr

rajibmitra:
This PR adds the path in stderr message.
- Fixes #
N/A
I thought of adding the path in stderr message.
**Release note**:
```
release-note
```

### DIFF
--- a/cmd/sonobuoy/app/retrieve.go
+++ b/cmd/sonobuoy/app/retrieve.go
@@ -85,7 +85,7 @@ func retrieveResults(cmd *cobra.Command, args []string) {
 
 	err = eg.Wait()
 	if _, ok := err.(exec.CodeExitError); ok {
-		fmt.Fprintln(os.Stderr, "Results not ready yet. Check `sonobuoy status` for status.")
+		fmt.Fprintln(os.Stderr, "Results not ready yet at /tmp/sonobuoy/. Check `sonobuoy status` for status.")
 		os.Exit(1)
 
 	} else if err != nil {


cmd/sonobuoy/app/retrieve.go
```diff
@@ -85,7 +85,7 @@ func retrieveResults(cmd *cobra.Command, args []string) {
 
 	err = eg.Wait()
 	if _, ok := err.(exec.CodeExitError); ok {
-		fmt.Fprintln(os.Stderr, "Results not ready yet. Check `sonobuoy status` for status.")
+		fmt.Fprintln(os.Stderr, "Results not ready yet at /tmp/sonobuoy/. Check `sonobuoy status` for status.")
```
timothysc: Well... that might be confusing, and it's customizable.

rajibmitra: Can I add parameters there ? the customized path you show there in stderr. what do you suggest ?

chuckha:
@rajibmitra Thank you for your contribution!

The updated wording adds a bit of confusion because it implies `/tmp/sonobuoy` will exist on the local machine when we are actually looking at the sonobuoy container. Also, like @timothysc mentioned, the path is customizable so we really should be using the customized attribute in this message. You might consider a message like "Check $somepath on the sonobuoy container" where $somepath is the customized variable. Lastly, you will need to add a sign off to your commit before we can merge it. Please see [sign-off docs](https://github.com/heptio/sonobuoy/blob/master/CONTRIBUTING.md#dco-sign-off). 

If you make those changes I think this would be an improvement in the user experience as it gives them more information to use while debugging.

cmd/sonobuoy/app/retrieve.go
timothysc: There is a config variable that is used to indicate output location.  

https://github.com/heptio/sonobuoy/blob/master/pkg/config/config.go#L276

johnSchnake:
Closing due to age; welcome to reopen/resubmit if you think this tweak will clarify things.

PR #583, opened 2019-02-07
Fix front matter issue

johnSchnake:
Last commit added front matter to the readme page which
broke the preprocessor. This change removes that.

It also moves the build status line below the header so that
the title in the browser does not include that markdown (a
jekyll plugin is creating the title from the first header).

So this gets the site up and fixes the title with the
cost of having the github pages build status shown in a
slightly different place than on the github repo page.

### DIFF
--- a/latest/README.md
+++ b/latest/README.md
@@ -1,10 +1,8 @@
----
-title: Sonobuoy
----
-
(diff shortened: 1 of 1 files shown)


asenchi:
Thanks for tracking this down @johnSchnake!

PR #584, opened 2019-02-10
Acknowledge state between plugins complete and done

### LINKED ISSUE #572: untarring `sonobuoy retrieve` tarball sometimes results in EOF in archive error
I have a script used in CI that:

* runs sonobuoy
* waits for its status to become `complete`
* retrieves the results tarball into an output directory and lets the user know where it is
* dumps the e2e test summary to stdout, ie
```Ran 136 of 996 Specs in 385.069 seconds

SUCCESS! -- 136 Passed | 0 Failed | 0 Pending | 860 Skipped 
```

Occasionally the last step fails with the following output:
```
gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
```

Which suggests to me that `sonobuoy retrieve` is not waiting for the tarball to be complete before pulling it locally.

I expect to be able to untar the tarball and grep files of interest to display results.

Seems potentially related to the following two issues:
* #498
* #407 

**Environment:**

- Sonobuoy version: v0.13.0
```Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.3", GitCommit:"a4529464e4629c21224b3d52edfe0ea91b072862", GitTreeState:"archive", BuildDate:"1970-01-01T00:00:01Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.5", GitCommit:"753b2dbc622f5cc417845f0ff8a77f539a4213ea", GitTreeState:"clean", BuildDate:"2018-11-26T14:31:35Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```
- Kubernetes installer & version: kubespray master branch
- Cloud provider or hardware configuration: openstack private cloud
```System Info:
 Machine ID:                 d4a9d15df9ad4f12993ffaff95800d0c
 System UUID:                D4A9D15D-F9AD-4F12-993F-FAFF95800D0C
 Boot ID:                    b09c3841-1bba-479e-9fcd-4112497bb6ae
 Kernel Version:             4.14.88-coreos
 OS Image:                   Container Linux by CoreOS 1967.3.0 (Rhyolite)
 Operating System:           linux
 Architecture:               amd64
 Container Runtime Version:  docker://18.6.1
 Kubelet Version:            v1.11.5
 Kube-Proxy Version:         v1.11.5
```

johnSchnake:
If your plugin artifacts are large then it will take
a few seconds for the tarball to be created. However,
the aggregator marks the pod status as done and
therefore there can be a few seconds where the
status reports that results are ready for download
but you will wither get a "file does not exist" error
or a tarball locally that is only partially complete
and most likely corrupted.

Fixes #572

In order to repro this you need to have large artifacts (as if you had a multi-node cluster running tests for numerous hours. This can easily be repro'd by using a custom kube-conformance image I created which took a dryrun result and then repeated it 2^15 times (image is available publicly at `schnake/kube-conformance:big`). The file ends up being ~1G before being tar'd. I scripted a little test where I watched for each status change and saw that in this case I had around 8s of time where I could get errors after the `sonobuoy status` output would report that results were available. Part of that time I'd get "no file" errors and the other times I was getting partial tar files.

**Release note**:
```
Fixed a bug which could cause the tarball from sonobuoy retrieve to not exist or be corrupted.
```

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -26,12 +26,14 @@ import (
 
 	"github.com/heptio/sonobuoy/pkg/config"
 	"github.com/heptio/sonobuoy/pkg/errlog"
-	pluginaggregation "github.com/heptio/sonobuoy/pkg/plugin/aggregation"
 	"github.com/pkg/errors"
 	"github.com/rifflock/lfshook"
 	"github.com/sirupsen/logrus"
 	"github.com/viniciuschiele/tarx"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
+
+	pluginaggregation "github.com/heptio/sonobuoy/pkg/plugin/aggregation"
 )
 
 // Run is the main entrypoint for discovery
@@ -124,7 +126,46 @@ func Run(kubeClient kubernetes.Interface, cfg *config.Config) (errCount int) {
 		defer os.RemoveAll(outpath)
 	}
 	trackErrorsFor("assembling results tarball")(err)
+
+	// 9. Mark final annotation stating the results are available and status is completed.
+	trackErrorsFor("updating pod status")(
+		updateStatus(kubeClient, cfg.Namespace, pluginaggregation.CompleteStatus),
+	)
+
 	logrus.Infof("Results available at %v", tb)
 
 	return errCount
+}
+
+// updateStatus changes the summary status of the sonobuoy pod in order to
+// effect the finalized status the user sees. This does not change the status
+// of individual plugins.
+func updateStatus(client kubernetes.Interface, namespace string, status string) error {
+	podStatus, err := pluginaggregation.GetStatus(client, namespace)
+	if err != nil {
+		return errors.Wrap(err, "failed to get the existing status")
+	}
+
+	// Update status
+	logrus.Infof("podstatus at first schnake %v", podStatus)
+
+	podStatus.Status = status
+
+	// Marshal back into json, inject into the patch, then serialize again.
+
+	logrus.Infof("podstatus at myupdate schnake %v", podStatus)
+	statusBytes, err := json.Marshal(podStatus)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal the status")
+	}
+
+	logrus.Infof("podstatus at statusbytes schnake %v", string(statusBytes))
+	patch := pluginaggregation.GetPatch(string(statusBytes))
+	patchBytes, err := json.Marshal(patch)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal the patch")
+	}
+
+	_, err = client.CoreV1().Pods(namespace).Patch(pluginaggregation.StatusPodName, types.MergePatchType, patchBytes)
+	return err
 }
--- a/pkg/plugin/aggregation/status.go
+++ b/pkg/plugin/aggregation/status.go
@@ -16,13 +16,26 @@ limitations under the License.
 
 package aggregation
 
-import "fmt"
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"k8s.io/client-go/kubernetes"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
 
 const (
 	// RunningStatus means the sonobuoy run is still in progress.
 	RunningStatus string = "running"
 	// CompleteStatus means the sonobuoy run is complete.
 	CompleteStatus string = "complete"
+	// PostProcessingStatus means the plugins are complete. The state is not
+	// put in the more finalized, complete, status until any postprocessing is
+	// done.
+	PostProcessingStatus string = "post-processing"
 	// FailedStatus means one or more plugins has failed and the run will not complete successfully.
 	FailedStatus string = "failed"
 )
@@ -42,7 +55,7 @@ type Status struct {
 }
 
 func (s *Status) updateStatus() error {
-	status := CompleteStatus
+	status := PostProcessingStatus
 	for _, plugin := range s.Plugins {
 		switch plugin.Status {
 		case CompleteStatus:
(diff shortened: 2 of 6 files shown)


pkg/discovery/discovery.go
```diff
@@ -124,7 +128,55 @@ func Run(kubeClient kubernetes.Interface, cfg *config.Config) (errCount int) {
...
+
 	logrus.Infof("Results available at %v", tb)
 
 	return errCount
 }
+
+// updateStatus changes the summary status of the sonobuoy pod in order to
+// effect the finalized status the user sees. This does not change the status
+// of individual plugins.
+func updateStatus(client kubernetes.Interface, namespace string, s string) error {
```
johnSchnake: This was the main crux of the work here and I tried a few different things.
1) I thought about moving some of the status/update code up out of the aggregator but as I started doing that I found that it ends up hitting the main interface that was exported to external consumers so I thought I should just leave it where it is.
2) I thought about referencing the `pkg/client` code to get the status (which looks very similar to this) but I wasn't sure if we wanted to tie the `discovery` and `client` packages together in the first place. Secondly, the client uses a `SonobuoyClient` and we just have the `kubernetes.Interface` available here.
3) I tried just patching the one status field that I want to edit, but unfortunately the entire annotation is just a json encoded string and not an object which can be patched. As a result I currently have to get the whole thing, unmarshal it, modify it, and repackage it. This means pulling in those new imports into this package.

I'd be happy to swap this out somehow for something simpler, but as of now it seems like that patch isn't possible without reading and unmarshaling the data first which requires all this extra logic.

pkg/plugin/aggregation/status.go
```diff
@@ -42,7 +46,7 @@ type Status struct {
 }
 
 func (s *Status) updateStatus() error {
-	status := CompleteStatus
+	status := PostProcessingStatus
```
johnSchnake: The aggregator itself now doesn't ever set the CompleteStatus, since it is just responsible for aggregating the data from the plugins. As a whole, sonobuoy has to signoff that everything is ready for download before we report that.

pkg/plugin/aggregation/status_test.go
```diff
@@ -25,14 +25,14 @@ func TestUpdateStatus(t *testing.T) {
 		expectedStatus string
 	}{
 		{
-			name:           "empty is complete",
+			name:           "empty is post-processing",
 			pluginStatuses: []string{},
-			expectedStatus: "complete",
+			expectedStatus: "post-processing",
```
johnSchnake: I did have to modify a few tests, but it genuinely seems to me like the model was just _slightly_ broken here so they really do need updated. No one should assume that anything is available for download or that sonobuoy is completely done while in this post-processing phase.

pkg/plugin/aggregation/update.go
```diff
@@ -146,7 +146,9 @@ func (u *updater) ReceiveAll(results map[string]*plugin.Result) {
 	}
 }
 
-func getPatch(annotation string) map[string]interface{} {
+// GetPatch takes a json encoded string and creates a map which can be used as
+// a patch to indicate the Sonobuoy status.
+func GetPatch(annotation string) map[string]interface{} {
```
johnSchnake: I did decide to export this; otherwise I was 100% duplicating this method in the discovery package. It didn't seem like there was really anything to gain from that seeing as how I did genuinely have a use case for this method outside of this pkg.

johnSchnake:
@chuckha Each plugin state still ends up in `complete`, its just that the _overall_ terminal status, as far as the aggregator is concerned, is `post-processing`. It is now the job of the main `discovery` code, which launches the aggregator, queries, etc to report the results are ready and `sonobuoy` is `complete`.

pkg/discovery/discovery.go
stevesloka: nit: `s` is not that descriptive to understand what it's being used for.

pkg/discovery/discovery.go
```diff
@@ -124,7 +128,55 @@ func Run(kubeClient kubernetes.Interface, cfg *config.Config) (errCount int) {
...
+	pod, err := client.CoreV1().Pods(namespace).Get(pluginaggregation.StatusPodName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "could not retrieve sonobuoy pod")
+	}
+
+	if pod.Status.Phase != corev1.PodRunning {
```
stevesloka: If the pod is running, why return an error?

pkg/discovery/discovery.go
```diff
@@ -124,7 +128,55 @@ func Run(kubeClient kubernetes.Interface, cfg *config.Config) (errCount int) {
...
+		return fmt.Errorf("pod has status %q", pod.Status.Phase)
+	}
+
+	statusJSON, ok := pod.Annotations[pluginaggregation.StatusAnnotationName]
```
stevesloka: nit: It might make this method a bit cleaner to break out the `getting the status from annotations` and `parsing status` from the `updateStatus` method. They might be re-usable elsewhere if we're getting/setting status in other places.

pkg/discovery/discovery.go
johnSchnake: It's that it is _not_ running, so we don't want to check the annotations on the pod and want to give a more clear error.

johnSchnake:
@stevesloka Awesome, thanks for the feedback. Yea I agree that the code needs a bit cleaned up; I didn't want to worry too much about those before I got buy-in about the approach in general (of using the `client` code to grab the status and such.

I'll clean up those nits and see what tests I can add. Because this fixes a timing bug it won't be easy to test in CI without doing a full run like my example but some of the smaller, helper, sections of code could have tests.

pkg/discovery/discovery.go
johnSchnake: 👍 In the next tweak I moved all the code actually out of the `pkg/client` and into the `plugins/aggregation` since all the types and constants in question were defined there. Used in both the `client` and `discovery` packages now.

johnSchnake:
@stevesloka If you take a look now its clear that there isn't much new code here. The new `discovery.updateStatus` method isn't really much new code. It just calls the `GetStatus` method which I moved from the `client` to the `aggregation` package. I then update a string, and create the patch from it (also in the aggregation package), and then patches the job.

Let me know if you think there is some test you'd like, but it doesn't seem like much of a new test case exists.

Otherwise I think this is ready to merge if I can get a LGTM. Happy to do a little demo though if it would help.

pkg/discovery/discovery.go
stevesloka: Ahh, you're right, sorry for the confusion there, I misread that.

PR #586, opened 2019-02-11
Update version command to print GitSHA, Min/Max K8S Versions

### LINKED ISSUE #518: Sonobuoy version and artifact checks.
Sonobuoy version should yield much more detailed information including: 
github sha
md5sum 
client version 
if (kubeconfig) {
server version 
apis detected
}

Part of the build and release testing should be to verify sums and shas

xref: #517

stevesloka:
Fixes #518 by adding additional version information:

- MinimumKubeVersion supported by this version of Sonobuoy
- MaximumKubeVersion supported by this version of Sonobuoy
- GitSHA
- K8S version (if kubeconfig is passed)

### DIFF
--- a/cmd/sonobuoy/app/version.go
+++ b/cmd/sonobuoy/app/version.go
@@ -20,20 +20,76 @@ import (
 	"fmt"
 
 	"github.com/heptio/sonobuoy/pkg/buildinfo"
+	"github.com/heptio/sonobuoy/pkg/errlog"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
-func init() {
-	RootCmd.AddCommand(versionCmd)
+type versionFlags struct {
+	kubecfg Kubeconfig
 }
 
-var versionCmd = &cobra.Command{
-	Use:   "version",
-	Short: "Print sonobuoy version",
-	Run:   runVersion,
-	Args:  cobra.ExactArgs(0),
+var versionflags versionFlags
+
+func init() {
+	cmd := &cobra.Command{
+		Use:   "version",
+		Short: "Print sonobuoy version",
+		Run:   runVersion,
+		Args:  cobra.ExactArgs(0),
+	}
+
+	AddKubeconfigFlag(&versionflags.kubecfg, cmd.Flags())
+
+	RootCmd.AddCommand(cmd)
 }
 
 func runVersion(cmd *cobra.Command, args []string) {
-	fmt.Println(buildinfo.Version)
+
+	fmt.Println(fmt.Sprintf("Sonobuoy Version: %s", buildinfo.Version))
+	fmt.Println(fmt.Sprintf("MinimumKubeVersion: %s", buildinfo.MinimumKubeVersion))
+	fmt.Println(fmt.Sprintf("MaximumKubeVersion: %s", buildinfo.MaximumKubeVersion))
+	fmt.Println(fmt.Sprintf("GitSHA: %s", buildinfo.GitSHA))
+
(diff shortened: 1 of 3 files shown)


cmd/sonobuoy/app/version.go
```diff
@@ -20,20 +20,76 @@ import (
...
+
+	AddKubeconfigFlag(&versionflags.kubecfg, cmd.Flags())
+
+	RootCmd.AddCommand(cmd)
 }
 
 func runVersion(cmd *cobra.Command, args []string) {
-	fmt.Println(buildinfo.Version)
+
+	fmt.Println(fmt.Sprintf("Sonobuoy Version: %s", buildinfo.Version))
```
johnSchnake: nit: Use `fmt.Printf("...%s\n")` rather than combining `Println` and `Sprintf`. Same for the lines below.

cmd/sonobuoy/app/version.go
```diff
@@ -20,20 +20,76 @@ import (
...
+	fmt.Println(fmt.Sprintf("MaximumKubeVersion: %s", buildinfo.MaximumKubeVersion))
+	fmt.Println(fmt.Sprintf("GitSHA: %s", buildinfo.GitSHA))
+
+	// Get Kubernetes version, this is last so that the regular version information
+	// will be shown even if the API server cannot be contacted and throws an error
+	apiVersion, skipk8sCheck := getK8Sversion()
+	if !skipk8sCheck {
+		fmt.Println("API Version: ", apiVersion)
+	} else {
+		fmt.Println("API Version check skipped due to missing `--kubeconfig` or other error")
```
johnSchnake: I'd much rather expose the issue: if I provided a config but still got this message I'd be pretty frustrated that there wasn't more to go off of. Any reason you wanted to hide them?

cmd/sonobuoy/app/version.go
```diff
@@ -20,20 +20,76 @@ import (
...
+	}
+}
+
+func getK8Sversion() (string, bool) {
```
johnSchnake: `(string,err)` would be a more clear signature. Then the caller can decide to expose the error or not.

cmd/sonobuoy/app/version.go
```diff
@@ -20,20 +20,76 @@ import (
...
+		sbc, err := getSonobuoyClient(cfg)
+		if err != nil {
+			errlog.LogError(errors.Wrap(err, "could not create sonobuoy client"))
+			return "", true
+		}
+
+		client, err := sbc.Client()
+		if err != nil {
+			errlog.LogError(err)
+			return "", true
```
johnSchnake: What exactly is the case here where you can get the sonobuoy client w/o error but not the client from that? Again, it would be more clear to the user to explain the issue via a new error rather than just aborting.

cmd/sonobuoy/app/version.go
```diff
@@ -20,20 +20,76 @@ import (
...
+		}
+
+		apiVersion, err := client.Discovery().ServerVersion()
+		if err != nil {
+			errlog.LogError(err)
+			return "", true
```
johnSchnake: Same comment w.r.t. explainations; I'd wrap the error and bubble it up.

cmd/sonobuoy/app/version.go
stevesloka: Yeah 100%, I refactored this out of the other method and didn't come back and update this.

stevesloka:
@johnSchnake thanks for the review! I've addressed all the comments.

PR #588, opened 2019-02-13
Run E2E tests inside a kind cluster

### LINKED ISSUE #250: Use a DIND test jig for automation
As part of the upstream testing commons work we should eventually be able to pull in a integration test framework that spins a DIND cluster that allows us to exercise larger integration pieces that are manual today.  

This is a long term effort.

stevesloka:

This PR now builds a kind cluster inside travis, runs the E2E tests and checks for a valid result. We'll need to see if this should be a 2nd job or have a bot only run occasionally since it will take some time to run. 

- Fixes #250 

**Release note**:
```
release-note
```

### DIFF
--- a/kind-config.yaml
+++ b/kind-config.yaml
@@ -1,0 +1,6 @@
+kind: Config
+apiVersion: kind.sigs.k8s.io/v1alpha2
+nodes:
+- role: control-plane
+- role: worker
+  replicas: 3
--- a/travis-ci.sh
+++ b/travis-ci.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Don't fail silently when a step doesn't succeed
+set -e
+
+make container deploy_kind
+
+echo "|---- Creating new Sonobuoy run/waiting for results..."
+./sonobuoy run --image-pull-policy=IfNotPresent -m Quick --wait
+
+outFile=$(./sonobuoy retrieve)
+results=$(./sonobuoy e2e $outFile)
+
+echo $results
+
+if echo $results | grep --quiet "failed tests: 0"; then
+    echo "|---- Success!"
+else
+    echo "|---- Failure: E2E tests failed"
+    exit 1
+fi
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@ examples/dev.yaml
 
 # Unencrypted credentials
 heptio-images-ee4b0474b93e.json
+
+# Dynamic Dockerfiles
+Dockerfile-amd64
+Dockerfile-arm64
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,25 +4,35 @@ go:
     - "1.10"
 
 services:
-- docker
+  - docker
 
 install: true
 
-script: VERBOSE=true make test int
+script: 
+  - VERBOSE=true make test int 
+  - ./travis-ci.sh 
 
 before_install:
-# Install gcloud cli here so it gets cached
-- if [ ! -d ${HOME}/google-cloud-sdk ]; then
-    curl https://sdk.cloud.google.com | bash /dev/stdin --disable-prompts;
-  fi
+  # Download and install Kind
+  - go get sigs.k8s.io/kind
+  - kind create cluster --config kind-config.yaml
+  - export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"
+  - curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
+
+
+before_deploy:
+  # Install gcloud cli here so it gets cached
+  - if [ ! -d ${HOME}/google-cloud-sdk ]; then
+      curl https://sdk.cloud.google.com | bash /dev/stdin --disable-prompts;
+    fi 
 
 deploy:
-- provider: script
-  skip_cleanup: true
-  script: ./travis-deploy.sh
-  on:
-    repo: heptio/sonobuoy
-    all_branches: true
+  - provider: script
+    skip_cleanup: true
+    script: ./travis-deploy.sh
+    on:
+      repo: heptio/sonobuoy
+      all_branches: true
   
 env:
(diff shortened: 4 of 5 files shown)


timothysc:
This makes me so happy!  Thx @stevesloka!

travis-ci.sh
```diff
@@ -0,0 +1,75 @@
...
+
+  echo "results not ready..."
+  false
+}
+
+echo "|---- Creating new Sonobuoy run..."
+$SONOBUOY_CMD run
+waitForDeployment
+
+while [ !checkForResults ]
```
johnSchnake: We'd want to be careful that this wouldn't end in an infinite loop if a deploy failed.

travis-ci.sh
```diff
@@ -0,0 +1,75 @@
...
+if [ -z "$TRAVIS" ]; then
+    echo "this script is intended to be run only on travis" >&2
+    exit 1
+fi
+
+KUBECTL_WAIT_CMD="kubectl wait --timeout=5m --for=condition=Ready pod/sonobuoy -n heptio-sonobuoy"
+SONOBUOY_CMD=./sonobuoy
+SONOBUOY_SUB_CMD=run -m Quick
+
+function waitForDeployment() {
```
johnSchnake: I have a PR nearly ready (maybe putting it up tomorrow) which would introduce the `--wait` options to run and delete commands. This makes scripting this much easier.

travis-ci.sh
```diff
@@ -0,0 +1,75 @@
...
+      $KUBECTL_WAIT_CMD
+    fi
+  done
+}
+
+function checkForResults() {
+  echo "|---- Looking for results..."
+
+  # Attempt to get status
+  $SONOBUOY_CMD retrieve
```
johnSchnake: So the fear as of now is that you can get a `retrieve` that generates a .tar.gz but that is corrupted because it was grabbed before the writer was done.

That gets solved if https://github.com/heptio/sonobuoy/pull/584 merges.

travis-ci.sh
```diff
@@ -0,0 +1,75 @@
...
+  $SONOBUOY_CMD status
+  sleep 30
+done
+
+echo "|------ Results found, validating output..."
+
+for i in *.gz; do
+  mkdir -p results
+  tar -xvf $i -C results
```
johnSchnake: Does the environment on travis get cleaned up properly behind the scenes, or do we need to handle that too?

travis-ci.sh
```diff
@@ -0,0 +1,75 @@
...
+  grep -q "Test Suite Passed" results/plugins/e2e/results/e2e.log
+  
+  if [ $? -eq 0 ]
+  then
+    echo "Success!"
+    exit 0
+  else
+    echo "Failure: E2E tests failed" >&2
```
johnSchnake: Not sure the file retention via jenkins here, but if we can't grab the tar then we need to be sure to tail/echo the end of the file where all the failed tests are listed otherwise we'll be perpetually frustrated by not knowing which tests failed.

stevesloka:
I think this is good now @johnSchnake! We'll need to come up with a way to make the full suite customizable. Maybe for each merge to master?

kind-config.yaml
```diff
@@ -0,0 +1,7 @@
+# three node (two workers) cluster config
+kind: Config
+apiVersion: kind.sigs.k8s.io/v1alpha2
+nodes:
+- role: control-plane
+- role: worker
+  replicas: 3
```
johnSchnake: Out of sync from the comment; do we want 3 workers or 2?

travis-ci.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# Don't fail silently when a step doesn't succeed
+set -e
+
+# if [ -z "$TRAVIS" ]; then
```
johnSchnake: Do we want this uncommented or removed?

travis-ci.sh
```diff
@@ -0,0 +1,38 @@
...
+# fi
+
+make container deploy_kind
+
+echo "|---- Creating new Sonobuoy run/waiting for results..."
+./sonobuoy run --image-pull-policy=IfNotPresent -m Quick --wait & PID=$!
+
+# While process is running...
+while kill -0 $PID 2> /dev/null; do 
+    printf  "."
```
johnSchnake: You started off the conversation about this in slack and the idea of just outputting a stream of `.` seems the most simple answer. I opened #607 regarding this.

travis-ci.sh
```diff
@@ -0,0 +1,38 @@
...
+done
+
+echo "|---- Retrieving results..."
+./sonobuoy retrieve
+
+echo "|---- Results found, validating output..."
+for i in *.gz; do
+mkdir -p results
+tar -xvf $i -C results
+grep -q "Test Suite Passed" results/plugins/e2e/results/e2e.log
```
johnSchnake: This section for getting and validating output can be reduced to something like:
```
outFile=$(./sonobuoy retrieve)
results=$(./sonobuoy e2e $outFile)
echo $results
exitCode=$($results | grep 0)
exit $exitCode
```

This is of course based on the idea that https://github.com/heptio/sonobuoy/pull/598 merges (any minute after tests go green).

That has the benefit that we also are 
1) dogfooding our own command
2) listing the failed tests, not just reporting failure or not (I know right now this is quick mode so its just one test, but you get the point)
3) less coupled to the text output by ginkgo for passed/failed
4) avoid some of the file creation/manipulation that can sometimes (in my experience) have odd edge cases

kind-config.yaml
stevesloka: 3 is a typical cluster that gives us enough workers I think for a test.

travis-ci.sh
stevesloka: I'll remove, I find it handy to kick it all off locally.

travis-ci.sh
stevesloka: I didn't want to wait for that so this is the easy fix. Showing output I think in the end game is best done with showing the total number of tests and which one is currently getting executed.

travis-ci.sh
stevesloka: We need to look at `sonobuoy e2e outfile`, it's not super helpful which is why I avoided (and your PR was still in flight).

travis-ci.sh
johnSchnake: Yea absolutely; I should have specified I didnt want that to be a blocker but just wanted to start the ticket since we had talked about it.

kind-config.yaml
johnSchnake: 👍 thats fine; I just didnt want the comment there saying 2 if we were setting 3.

stevesloka:
@johnSchnake all cleaned up with comments. Please take a look, thanks! =)

travis-ci.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo "|---- Creating new Sonobuoy run/waiting for results..."
+./sonobuoy run --image-pull-policy=IfNotPresent -m Quick --wait
+
+outFile=$(./sonobuoy retrieve)
+results=$(./sonobuoy e2e $outFile)
+
+echo $results
+
+if echo $results | grep --quiet "failed tests: 0"; then
```
aojea: This is not enough condition to declare the test OK.
I have the same result in a failed deployment 
````
{"level":"info","msg":"Collecting Pod Logs...","time":"2019-03-05T12:35:40Z"}
{"level":"error","msg":"the server rejected our request for an unknown reason (get pods sonobuoy-e2e-job-27d5ca4105034c28)","
time":"2019-03-05T12:35:40Z"}
```

and failed tests :0 but the problem that it didn't run any test
```
sonobuoy e2e 201903051232_sonobuoy_7cd48d20-e974-42dd-8a4b-1dbcf0ae3bd2.tar.gz
failed tests: 0
```

PR #591, opened 2019-02-15
Fix problem for building locally on non-Linux OS

### LINKED ISSUE #590: Cross compile for Mac via `make`
I'm on a Mac and it seems like the Makefile just isnt set up to build_sonobuoy for the darwin GOOS.

It seems like the fix is trivial (track and use the GOOS env var in the makefile) so that we can get `GOOS=darwin make build_sonobuoy` to create Mac compatible binaries.

The reason this is needed (and it seems near bug-level) is that on a fork of sonobuoy you cant use `go build` because of all the import paths referencing the heptio repo.

This fix in the makefile is what I've been doing locally and it works fine. I wanted an issue to track/mention though.

johnSchnake:
This change will keep the existing make settings in place
but will fix a problem where `make build_sonobuoy` wouldn't
apply the correct GOOS/GOARCH for the host.

This means that anyone can run `make build_sonobuoy` and get
just their sonobuoy binary created and that `make sonobuoy`
creates the 2 linux binaries (always generated) and the
correct sonobuoy binary (in the cwd).

Fixes #590

**Release note**:
```
Fixed host build chain for non-linux/amd64 GOOS/GOARCH
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,12 @@ LINUX_ARCH := amd64 arm64
 DOCKERFILE :=
 PLATFORMS := $(subst $(SPACE),$(COMMA),$(foreach arch,$(LINUX_ARCH),linux/$(arch)))
 
+# Not used for pushing images, just for local building on other GOOS. Defaults to
+# grabbing from the local go env but can be set manually to avoid that requirement.
+HOST_GOOS ?= $(shell go env GOOS)
+HOST_GOARCH ?= $(shell go env GOARCH)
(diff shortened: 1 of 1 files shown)


johnSchnake:
Discovered via chat, but for clarity:
the problem is if you have a fork of this repo that you maintain locally in a different location than the original. For instance, I actually clone and work from my fork `johnSchnake/sonobuoy` so the import paths won't match whats in the code.

So `make sonobuoy` (or more specifically `make build_sonobuoy`) will say its building for my host, but in reality it just builds for a linux GOOS and the binary it generates will not work on my local machine.

Issue #482, opened 2018-07-11
Make sonobuoy timeout configurable from the CLI


ceridwen:
**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Add a command-line option to make the time out configurable to `sonobuoy run` and `sonobuoy gen`.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

I'm using sonobuoy for to run the conformance tests in an environment where it's typical for it to take between 5 and 6 hours to finish the tests.  A significant chunk of this time is waiting for tests to hit their time outs when they fail, but I also have environments where cluster operations simply take a long time.  In v.0.10, I increased the timeout by creating a private branch of the repo; in v0.11.3 I'm having to rewrite the JSON in the ConfigMap.

**Environment:**

- Sonobuoy version: v0.11.3
- Kubernetes version: (use `kubectl version`): 1.10, OpenShift 3.9
- Kubernetes installer & version: kubespray and OpenShift
- Hardware configuration: VMs or bare metal
- OS (e.g. from `/etc/os-release`):  Ubuntu or RHEL

PR #592, opened 2019-02-15
Add wait flags to run and delete

### LINKED ISSUE #437: Support new parameter --wait-until-finish

### LINKED ISSUE #568: Feature Request: sonobuoy delete --wait

It would be nice if `sonobuoy delete` surfaced a `--wait` flag similar to `kubectl delete`.

johnSchnake:
Currently running a sonobuoy run takes significant time and
the user is left to poll the status on their own. In a
CI environment this is especially true.

Deleting, likewise, takes some time since cleaning up
namespaces in k8s is not always fast.

This PR introduces wait options to run/delete and allows
customizable timeouts to the wait behavior.

Fixes #568
Fixes #437

**Release note**:
```
Adds --wait flag to run and delete commands.
```

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -189,6 +189,24 @@ func AddDeleteAllFlag(flag *bool, flags *pflag.FlagSet) {
 	)
 }
 
+// AddDeleteWaitFlag adds a boolean flag for waiting for the delete process to complete.
+func AddDeleteWaitFlag(flag *int, flags *pflag.FlagSet) {
+	flags.IntVar(
+		flag, "wait", 0,
+		"Wait for resources to be deleted before completing. 0 indicates do not wait. By providing --wait the default is to wait for 1 hour.",
+	)
+	flags.Lookup("wait").NoOptDefVal = "60"
+}
+
+// AddRunWaitFlag adds an int flag for waiting for the entire run to finish.
+func AddRunWaitFlag(flag *int, flags *pflag.FlagSet) {
+	flags.IntVar(
+		flag, "wait", 0,
+		"Wait for sonobuoy run to be completed (or fail). 0 indicates do not wait. By providing --wait the default is to wait for 1 day.",
+	)
+	flags.Lookup("wait").NoOptDefVal = "1440"
+}
+
 // AddImagePullPolicyFlag adds a boolean flag for deleting everything (including E2E tests).
 func AddImagePullPolicyFlag(policy *ImagePullPolicy, flags *pflag.FlagSet) {
 	*policy = ImagePullPolicy(v1.PullAlways) //default
--- a/pkg/client/delete.go
+++ b/pkg/client/delete.go
@@ -18,11 +18,13 @@ package client
 
 import (
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	kubeerror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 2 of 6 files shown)


cmd/sonobuoy/app/args.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/heptio/sonobuoy/pkg/config"
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
johnSchnake: Not sure why, but goimports insists on this change

cmd/sonobuoy/app/args.go
```diff
@@ -183,6 +183,30 @@ func AddDeleteAllFlag(flag *bool, flags *pflag.FlagSet) {
 	)
 }
 
+// AddDeleteWaitFlag adds a boolean flag for waiting for the delete process to complete.
+func AddDeleteWaitFlag(flag *bool, flags *pflag.FlagSet) {
```
johnSchnake: A little unsure if I should just try and combine the flags, maybe even into their own flagset so they get added together automatically.

Also open to debate about how much to allow customized without flag proliferation:
 --wait
 --wait-timeout

Unsure if we should allow these to be customized; defaults seem sufficient for now:
- polling interval (makes sense to poll delete every 2-5s but run should probably be more like 20-30s default since runs take so much longer)
- grace period (when starting a run, we need to give some time within which to allow errors/allow the pods to come up, but how long do we allow the pending state without erroring?)

pkg/client/delete.go
```diff
@@ -33,46 +35,85 @@ const (
 	e2eNamespacePrefix = "e2e-"
 )
 
+var (
+	pollFreq    = 5 * time.Second
+	pollTimeout = 5 * time.Minute
```
johnSchnake: Not customizable right now; will the values be a problem ever? Not sure the upper bound on namespace cleanup times. Maybe we should just bump that to something pretty big like an hour to ensure CI systems dont get hung but long enough it shouldn't get hit.

pkg/client/interfaces.go
```diff
@@ -60,13 +61,16 @@ type E2EConfig struct {
...
+	Wait        bool
+	WaitTimeout time.Duration
 }
 
 // DeleteConfig are the input options for cleaning up a Sonobuoy run.
 type DeleteConfig struct {
 	Namespace  string
 	EnableRBAC bool
 	DeleteAll  bool
+	Wait       bool
```
johnSchnake: WaitTimeout should have been added here; in which case my previous comment about it not being customizable is false. That gets plumbed down; just the polling interval is not.

pkg/client/run.go
```diff
@@ -18,19 +18,27 @@ package client
...
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
 const bufferSize = 4096
 
+var (
+	pollInterval = 10 * time.Second
```
johnSchnake: Too aggressive? 20-30s?

pkg/client/delete.go
```diff
@@ -33,46 +35,84 @@ const (
...
+// namespaces.
 func (c *SonobuoyClient) Delete(cfg *DeleteConfig) error {
 	client, err := c.Client()
 	if err != nil {
 		return err
 	}
 
-	if err := cleanupNamespace(cfg.Namespace, client); err != nil {
+	conditions := []wait.ConditionFunc{}
+	nsCondition, err := cleanupNamespace(cfg.Namespace, client)
```
johnSchnake: I wrote the logic here so that the method signatures of each of these cleanup steps is responsible for creating (and returning) their own "wait" function. This is so that the logic for the action is right next to the logic regarding what needs to be waited on.

Maybe it would be better to define these as unexported functions though (outside of the functions) so that its a bit easier to test? I do wish these had more test coverage; I think that is the main todo before merge. Good to test manually a.t.m. but needs some automated testing.

cmd/sonobuoy/app/args.go
stevesloka: I'm wondering if we should use just `--timeout` for the flag. Are there any other timeouts that would apply to each action?

pkg/client/delete.go
stevesloka: Were you going to add tests? I don't see any added.

cmd/sonobuoy/app/args.go
stevesloka: ...that would then we'd rename everything from `waittimeout` to just `timeout` in the code for consistency.

johnSchnake: So there is another timeout which is within the sonobuoy config; it is the timeout for how long the aggregator will wait before forcing plugins to stop. If you sync those two up then you are being racy and will always short some output since there is actions the server takes after that timeout. Plus, keeping them separate allows a little more flexibility in case you want to wait (but not force plugin failures) then check logs or something like that.

johnSchnake: As of now that timeout is just in the sonobuoy config (passed via the --config flag) but there is an issue up for making that customizable from the flags. #482

pkg/client/delete.go
johnSchnake: So when we did a brief code review in person I pointed out how, once you extract the boilerplate of getting clients and making API calls, the actual logic of these wait methods is akin to `if empty` or something similarly trivial. At the time I thought you were saying you were OK without adding tests; you want me to do that though?

I could refactor it to try and include tests for each of the wait conditions and then for the "all conditions" condition.

cmd/sonobuoy/app/args.go
stevesloka: I see, thanks for clarifying. I'm not in love with the `wait-timeout` piece. Would there be a way to have just `--wait` which would wait and apply the default, then you could use `--wait=60` to specify the timeout? 

If not, it is what it is. =)

pkg/client/delete.go
stevesloka: Ok cool, was just going off of your comments inline. No problem.

cmd/sonobuoy/app/args.go
johnSchnake: Yes I can do that with cobra. sgtm.

johnSchnake: Modified flags so that it is just one value being tracked (an int) and not two (int and bool).

 - I take use of the 0 value (which I default the flags to) meaning that we should not wait.
 - If the flag is provided but not specified a value: `--wait` then the default value is provided (60m for delete, 24h for run)
 - If the flag is specified a value, it is used as expected

PR #603, opened 2019-02-25
Restructured cmd package

### LINKED ISSUE #462: Restructure cmd package
The command package uses init a lot. Every command requires an init function to register itself.

This makes testing hard as we have no way to mock anything. We run the test, the init runs and the command is built.

I propose we follow a pattern that is a variation on upstream. We start with a [root command](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/kubeadm.go) and from the root we import our sonobuoy [top level command](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/cmd/cmd.go#L32) and call Execute. It will look a lot like [this file](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/kubeadm.go#L31).

## What this would enable

* proper dependency injection
* a single spot that clearly defines all available subcommands
* a clear place to parse flags, i suspect we are doing it slightly wrong because of glog
* ability to run a higher level test
* better package structure

Globals are the norm in sonobuoy CLI and they should not be.

ekj1711:
* Changed init functions in cmd pkg to individual register functions
* Created a root command to import sonobouy top level commands
* Added Execute() in main

- Fixes #462


cmd/sonobuoy/app/gen_plugin.go
```diff
@@ -55,12 +56,13 @@ func init() {
...
 	}
 
-	GenCommand.PersistentFlags().StringArrayVarP(
+	cmd.PersistentFlags().StringArrayVarP(
 		&genPluginOpts.Paths, "paths", "p", []string{".", "./examples/plugins.d/"},
 		"the paths to search for the plugins in. Defaults to . and ./plugins.d/",
 	)
 	// TODO: Other options?
-	GenCommand.AddCommand(cmd)
+	cmd.AddCommand(cmd)
```
johnSchnake: You're adding it here to itself? Assuming this was just leftover from tinkering and needs removed.

That causes the help to panic:
```
panic: Command can't be a child of itself

goroutine 1 [running]:
github.com/heptio/sonobuoy/vendor/github.com/spf13/cobra.(*Command).AddCommand(0xc000148500, 0xc0000ede80, 0x1, 0x1)
	/Users/jschnake/go/src/github.com/heptio/sonobuoy/vendor/github.com/spf13/cobra/command.go:999 +0x222
github.com/heptio/sonobuoy/cmd/sonobuoy/app.NewCmdGenPlugin(0x2063be0, 0xc0000d4008, 0x1)
	/Users/jschnake/go/src/github.com/heptio/sonobuoy/cmd/sonobuoy/app/gen_plugin.go:64 +0x1bf
github.com/heptio/sonobuoy/cmd/sonobuoy/app.NewSonobuoyCommand(0x2063be0, 0xc0000d4008, 0xc0000aa058)
	/Users/jschnake/go/src/github.com/heptio/sonobuoy/cmd/sonobuoy/app/root.go:41 +0x2ad
main.main()
	/Users/jschnake/go/src/github.com/heptio/sonobuoy/main.go:27 +0x39
```

cmd/sonobuoy/app/delete.go
```diff
@@ -34,7 +35,7 @@ var deleteFlags struct {
 	wait       int
 }
 
-func init() {
+func NewCmdDelete(out io.Writer) *cobra.Command {
```
johnSchnake: You added the io.Writer to all of these functions, why? I absolutely get that at some point down the line you may want to mock out `io` for the commands but this doesn't plumb it through any farther so the variable is just unused everywhere.

Seems like for now this io stuff should be out of scope for this change. Let me know if there something I wasn't seeing w.r.t. to it though.

ekj1711:
Thanks @johnSchnake for reviewing the change. I have addressed the issues with AddCommand() and io.Writer

johnSchnake:
lgtm; just needs a rebase now and you should be fine. Thanks again.

PR #612, opened 2019-02-27
Add support for KUBE_TEST_REPO_LIST and custom registries

### LINKED ISSUE #160: Config option to reference private registry

johnSchnake:
K8s 1.13 organized and allowed overrides for where to pull
e2e images from. The env var needed is KUBE_TEST_REPO_LIST.

To facilitate users providing this, the following changes
have been made:

 - the user can specify a file path on the command line with
the proper contents
 - the file is read
 - the contents are used to create a configmap of the values
 - the configmap is mounted as a volume for the kube-conformance
container
 - the env var KUBE_TEST_REPO_LIST is set to point to this file

Fixes #160

Personally to test this I just pulled from `docker.io/library` (the nginx:1.14-alpine which is used by the quick mode test), tagged it as `schnake/nginx:1.14-alpine`, pushed to the public dockerhub and ran the test again (setting the yaml/flag as necessary). You could test with any test/image this one was just fast. It has the downside of tearing down the container though so you just look through logs and such to see what it was based on.

Otherwise we need to set up another public docker registry, secured with TLS.

**Release note**:
```release-note 
Adds support for custom registries during e2e tests via --e2e-repo-config flag.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,35 @@ To clean up Kubernetes objects created by Sonobuoy, run:
 sonobuoy delete
 ```
 
+### Custom registries and air-gapped testing
+
+In air-gapped deployments where there is no access to the public Docker registries
+Sonobuoy supports running end-to-end tests with custom registries. This enables
+you to test your air-gapped deployment once you've loaded the necessary images
+into a registry that is reachable by your cluster.
+
+Just provide the `--e2e-repo-config` parameter and pass it the path to a local
+yaml file pointing to the registries you'd like to use. This will instruct the
+Kubernetes end-to-end suite to use your registries instead of the default ones.
+
+```
+sonobuoy run --e2e-repo-config custom-repos.yaml
+```
+
+The registry list is a yaml document specifying a few different registry
+categories and their values:
+
+```
+dockerLibraryRegistry: docker.io/library
+e2eRegistry: gcr.io/kubernetes-e2e-test-images
+gcRegistry: k8s.gcr.io
+privateRegistry: gcr.io/k8s-authenticated-test
+sampleRegistry: gcr.io/google-samples
+```
+
+The keys in that file are specified in the Kubernetes test framework itself. You
+may provide a subset of those and the defaults will be used for the others.
+
 ### Run on Google Cloud Platform (GCP)
 
 Note that if you run Sonobuoy on a Google Kubernetes Engine (GKE) cluster, you
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -18,7 +18,10 @@ package app
 
 import (
 	"fmt"
+	"io/ioutil"
 	"strings"
+
+	"gopkg.in/yaml.v2"
 
 	ops "github.com/heptio/sonobuoy/pkg/client"
 	"github.com/heptio/sonobuoy/pkg/config"
@@ -100,9 +103,10 @@ func AddSonobuoyConfigFlag(cfg *SonobuoyConfig, flags *pflag.FlagSet) {
 }
 
 const (
-	e2eFocusFlag    = "e2e-focus"
-	e2eSkipFlag     = "e2e-skip"
-	e2eParallelFlag = "e2e-parallel"
+	e2eFocusFlag          = "e2e-focus"
+	e2eSkipFlag           = "e2e-skip"
+	e2eParallelFlag       = "e2e-parallel"
+	e2eRegistryConfigFlag = "e2e-repo-config"
 )
 
 // AddE2EConfigFlags adds three arguments: --e2e-focus, --e2e-skip and
@@ -127,6 +131,10 @@ func AddE2EConfigFlags(flags *pflag.FlagSet) *pflag.FlagSet {
 	e2eFlags.String(
 		e2eParallelFlag, defaultMode.E2EConfig.Parallel,
 		"Specify the E2E_PARALLEL flag to the conformance tests. Overrides --mode.",
+	)
+	e2eFlags.String(
+		e2eRegistryConfigFlag, "",
+		"Specify a yaml file acting as KUBE_TEST_REPO_LIST, overriding registries for test images.",
 	)
 	e2eFlags.MarkHidden(e2eParallelFlag)
 	flags.AddFlagSet(e2eFlags)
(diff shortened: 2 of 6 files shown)


pkg/templates/manifest.go
```diff
@@ -138,6 +146,12 @@ data:
       secret:
         secretName: ssh-key
         defaultMode: 256
+{{- end}}
+{{- if .CustomRegistries }}
+    extra-volumes:
```
johnSchnake: Both this and the ssh logic above add this line if needed. I'll need to make sure it doesnt get duplicated.

cmd/sonobuoy/app/args.go
```diff
@@ -128,6 +130,10 @@ func AddE2EConfigFlags(flags *pflag.FlagSet) *pflag.FlagSet {
 		e2eParallelFlag, defaultMode.E2EConfig.Parallel,
 		"Specify the E2E_PARALLEL flag to the conformance tests. Overrides --mode.",
 	)
+	e2eFlags.String(
+		e2eCustomRegistryFlag, "",
+		"Specify a yaml file acting as KUBE_TEST_REPO_LIST, overriding registries for test images.",
```
johnSchnake: TODO: Add something in the readme about this functionality and what the file looks like.

johnSchnake:
So far I've just tested gen; after getting this done today I hadn't had a chance to spin up a test using it.

Just wanted to open this draft sooner rather than later for visibility and comments.

johnSchnake:
- Fixed the `extra-volumes` issue (it could have gotten duplicated before)
 - Added namespace metadata to the new configmap

Tested with a single image and confirmed it pulled to my own repo. I'm going to add some stuff the readme, but then other than that it is ready for some more testing and review.

cmd/sonobuoy/app/args.go
```diff
@@ -128,6 +130,10 @@ func AddE2EConfigFlags(flags *pflag.FlagSet) *pflag.FlagSet {
 		e2eParallelFlag, defaultMode.E2EConfig.Parallel,
 		"Specify the E2E_PARALLEL flag to the conformance tests. Overrides --mode.",
 	)
+	e2eFlags.String(
+		e2eCustomRegistryFlag, "",
```
stevesloka: nit: Should we make this something like `e2eCustomRegistryConfigFlag`? Just to make clear it's expecting a config file and not `gcr.io/heptio-imags` or something?

cmd/sonobuoy/app/args.go
```diff
@@ -160,6 +166,18 @@ func GetE2EConfig(mode ops.Mode, flags *pflag.FlagSet) (*ops.E2EConfig, error) {
...
+	if flags.Changed(e2eCustomRegistryFlag) {
+		repoFile, err := flags.GetString(e2eCustomRegistryFlag)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't retrieve registry list flag")
+		}
+		contents, err := ioutil.ReadFile(repoFile)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't read registry list")
+		}
+		cfg.CustomRegistries = string(contents)
```
stevesloka: Are all values required or can I just override one of the params and the defaults are used for the others? I wonder if we should add some sort of validation to make sure users don't pass a bad config?

README.md
```diff
@@ -120,6 +120,34 @@ To clean up Kubernetes objects created by Sonobuoy, run:
...
 ```
 
+### Custom registries and air-gapped testing
+
+In air-gapped deployments where there is no access to the public Docker registries
+Sonobuoy supports running end-to-end tests with custom registries. This enables
+you to test your air-gapped deployment once you've loaded the necessary images
+into a registry that is reachable by your cluster.
+
+Just provide the `--e2e-repo-list` parameter and pass it the path to a local list
```
stevesloka: nit: "pass it the path to a local list of registries". This sounds like I can pass `--e2e-repo-list gcr.io/heptio-images,docker.io`.

cmd/sonobuoy/app/args.go
```diff
@@ -100,9 +101,10 @@ func AddSonobuoyConfigFlag(cfg *SonobuoyConfig, flags *pflag.FlagSet) {
 }
 
 const (
-	e2eFocusFlag    = "e2e-focus"
-	e2eSkipFlag     = "e2e-skip"
-	e2eParallelFlag = "e2e-parallel"
+	e2eFocusFlag          = "e2e-focus"
+	e2eSkipFlag           = "e2e-skip"
+	e2eParallelFlag       = "e2e-parallel"
+	e2eCustomRegistryFlag = "e2e-repo-list"
```
stevesloka: Same as the comment for the docs, should this be `e2e-repo-config` or something implying a config file?

cmd/sonobuoy/app/args.go
johnSchnake: Sure. I struggled with the naming a bit since I didnt want to make it too long/awkward. ala `e2eCustomRegistryFilepathRegistryListFlag` :-p Yours sounds fine to me 👍

stevesloka: haha. :1st_place_medal:

cmd/sonobuoy/app/args.go
johnSchnake: That sounds good; `repo-config` is a good balance between saying `repo-file` which wasnt as clear to me and `repo` which definitely suggested a single repo.

cmd/sonobuoy/app/args.go
johnSchnake: I wondered about the validation too but thought that passing the plain file itself just had the benefit of ensuring the implementations were too coupled. We could add a simple yaml unmarshal to ensure its valid yaml at least.

And the code for its consumption is here: https://github.com/kubernetes/kubernetes/blob/master/test/utils/image/manifest.go#L58-L81

It just unmarshals into a map; so if settings aren't provided in the custom file it should be using the defaults. I'll add that to the readme as I'm sure that will be a common question.

stevesloka: I think ideally we do the validation upstream, so what you have is good.

cmd/sonobuoy/app/args.go
johnSchnake: `e2eRegistryConfigFlag`; by it being a flag its clear enough its custom. The name was getting a bit long :)

cmd/sonobuoy/app/args.go
johnSchnake: TL;DR; I'm going to add it :)

Longer version:
eh I'm just split. I just had another issue #615 come up which having a bit of client validation would be nice. Otherwise for users of sonobuoy that means having to 
 - run the tests
 - see the failure
 - pull the logs
 - untar
 - know where to look
 - find the error about parsing
 - loop back to see what was being parsed and hopefully figure out the small repo config was the problem

In both the regexp and yaml cases, we just need to do a yaml/regexp parse and error check. Since we already document they should be yaml/regexp we are only slightly increasing the coupling by doing minimal validation clientside.

README.md
johnSchnake:
```
Just provide the `--e2e-repo-config` parameter and pass it the path to a local
yaml file pointing to the registries you'd like to use.
```

PR #622, opened 2019-03-05
Wait should emit some output

### LINKED ISSUE #607: Wait should emit some output
The `--wait` functionality of run/delete is great but run, since it can take so long, can really be difficult for a user to know if the process is still running as expected.

I think we should have a reasonable default that looks good to users sitting in front of the machine but allow an ovverride via a flag so that people in other situations/preferences can override for their use case.

Suggestion:
Default: A simple spinner that might also show the statement like "sonobuoy state is xyz, this command will return after it completes or errors"

Alternative options via flag:
 - just lazily output a stream of dots every few min so it doesnt read as a dead process. Can be ugly if its a long run but may really be the preference of some people who just want an 'alive' signal.

 - no output, just wait quietly as we do today

The reason I say the default should be the good looking but (maybe) slightly less robust one is that people writing CI scripts or with strong preferences dont mind adding the flag. People typing quickly just getting it working will want/like the nicer output IMO.

ekj1711:
Add a spinner so that during long runs its easy for users to know that
process is still running. It should also update the intermediate state
of execution for sonobouy run and delete command.

Fixes #607

Signed-off-by: ekj1711 <ejohn@vmware.com>

- Fixes #607


johnSchnake:
Thanks @ekj1711 ! I'll take a look as soon as I can. FYI tests are failing because the build can't find the vendored code; we save the vendored code in the repo as well so you'll need to run `dep ensure` and commit the new vendor code too.

ekj1711:
Thanks @johnSchnake 
I have committed the vendor package also along with the changes.

cmd/sonobuoy/app/delete.go
```diff
@@ -52,35 +53,59 @@ func NewCmdDelete() *cobra.Command {
 }
 
 func deleteSonobuoyRun(cmd *cobra.Command, args []string) {
+	s := spinner.New(spinner.CharSets[0], 100*time.Millisecond)
```
johnSchnake: We should const the charset, timing, and color. I'd make a helper method to create those too so that other uses (if we have them) just reference `getSpinner` for instance.

Also, nit: I prefer #14 from https://github.com/briandowns/spinner#available-character-sets :)

cmd/sonobuoy/app/delete.go
```diff
@@ -52,35 +53,59 @@ func NewCmdDelete() *cobra.Command {
...
+			s.Stop()
+		}
 		errlog.LogError(errors.Wrap(err, "couldn't get kubernetes config"))
 		os.Exit(1)
 	}
 
 	sbc, err := getSonobuoyClient(cfg)
 	if err != nil {
+		if deleteFlags.wait > 0 {
+			s.Suffix = "Status: Could not create Sonobouy Client!"
```
johnSchnake: Why repeat the status line here when the error will be wrapped/returned? I don't think that extra UI element is really necessary.

cmd/sonobuoy/app/delete.go
```diff
@@ -52,35 +53,59 @@ func NewCmdDelete() *cobra.Command {
...
+			s.Stop()
+		}
 		errlog.LogError(errors.Wrap(err, "could not create sonobuoy client"))
 		os.Exit(1)
 	}
 
 	client, err := sbc.Client()
 	if err != nil {
+		if deleteFlags.wait > 0 {
+			s.Stop()
```
johnSchnake: This pattern is repeated elsewhere, where on each path you check/stop the spinner.

The intuitive thing is after starting the spinner to just say `defer s.Stop()`, does that interact poorly with the rest of the UI?

cmd/sonobuoy/app/delete.go
```diff
@@ -52,35 +53,59 @@ func NewCmdDelete() *cobra.Command {
...
 	if err := sbc.Delete(&deleteopts); err != nil {
+		s.Suffix = "Status: Failed to delete sonobuoy resources!"
+		s.Stop()
 		errlog.LogError(errors.Wrap(err, "failed to delete sonobuoy resources"))
 		os.Exit(1)
 	}
-
-}
+	if deleteFlags.wait > 0 {
+		s.FinalMSG = "Status: Success!"
```
johnSchnake: Again, I'd remove this. This one is especially misleading since users will think this means that their tests were successful when this just means that the run completed.

ekj1711:
Thanks @johnSchnake for the review comments.

cmd/sonobuoy/app/run.go
```diff
@@ -71,6 +75,15 @@ func NewCmdRun() *cobra.Command {
 }
 
 func submitSonobuoyRun(cmd *cobra.Command, args []string) {
+
+	var s *spinner.Spinner
+
+	if (runflags.wait > 0 && runflags.waitOutput == "spinner") {
```
johnSchnake: I'd modify this and the flag definition in a few ways:
 - const values like `spinner`
 - create a type with custom `Set` method so that pflag calls it when set. Validate the user provided only one of the allowable values. See `AddRBACModeFlags` for an example.

cmd/sonobuoy/app/run.go
```diff
@@ -71,6 +75,15 @@ func NewCmdRun() *cobra.Command {
...
+		s = getSpinnerInstance()
+		s.Start()
```
johnSchnake: I'm on the fence about the desire to just put this snippet in a goroutine and sleep for 2-3s just to avoid the first line clash between the spinner and the log output.

Maybe we just leave it this way for now (again it defaults to silent) and see how it goes. We can always do that followup, especially if we ever want to make it the default.

cmd/sonobuoy/app/common.go
```diff
@@ -1,13 +1,19 @@
...
 	sonodynamic "github.com/heptio/sonobuoy/pkg/dynamic"
+	"time"
 
 	"github.com/pkg/errors"
 	"k8s.io/client-go/rest"
 )
 
+var spinnerType int  = 14
+var spinnerDuration time.Duration = 2000*time.Millisecond
+var spinnerColor = "red"
```
johnSchnake: nit: use the grouped var like
```
var (
  spinnerType...
  spinnerDuration...
  etc
)
```

cmd/sonobuoy/app/delete.go
```diff
@@ -47,11 +49,19 @@ func NewCmdDelete() *cobra.Command {
...
 	AddDeleteAllFlag(&deleteopts.DeleteAll, cmd.Flags())
 	AddDeleteWaitFlag(&deleteFlags.wait, cmd.Flags())
+	AddWaitOutputFlag(&deleteFlags.waitOutput, cmd.Flags())
 
 	return cmd
 }
 
 func deleteSonobuoyRun(cmd *cobra.Command, args []string) {
+
+	var s *spinner.Spinner
```
johnSchnake: Dont need to define this here, just do `:=` below when you get the instance. Then we also wont need to add the import line at the top of this file.

cmd/sonobuoy/app/run.go
```diff
@@ -93,7 +106,7 @@ func submitSonobuoyRun(cmd *cobra.Command, args []string) {
 		plugins[i] = plugin.Name
 	}
 
-	if len(plugins) > 0 {
+	if len(plugins) > 0 && runflags.wait == 0 {
```
johnSchnake: What was this added for? Seems unrelated to this spinner change, maybe just a leftover of tinkering?

pkg/config/config.go
```diff
@@ -47,6 +47,7 @@ const (
 	MasterContainerName = "kube-sonobuoy"
 	// MasterResultsPath is the location in the main container of the master pod where results will be archived.
 	MasterResultsPath = "/tmp/sonobuoy"
+	DefaultWaitOutput = "silent"
```
johnSchnake: Once a type is in place so we can properly validate this flag whenever set, this would be changed to referring to a specific, typed value not the string.

cmd/sonobuoy/app/common.go
ekj1711: Done.

cmd/sonobuoy/app/run.go
ekj1711: Done.

johnSchnake:
ping; are you still planning to fix this up and get it merged?

ekj1711:
> ping; are you still planning to fix this up and get it merged?

Yes, sorry for delay. I will update new patch in next 2 days.

johnSchnake:
Based on my quick look it seems good; some nits but nothing I'll block it on.

I'll try and build/run it manually tomorrow just to double check things but otherwise I think this is a great start to try this feature out. Once we get some sort of status capability, this may be really useful to have, and, as the original problem scope stated, it already solves the problem of leaving a terminal hanging in CI (which can cause fatal errors)

johnSchnake:
Builds and runs great! Thanks @ekj1711 !

PR #625, opened 2019-03-06
WIP: Update ci builds

### LINKED ISSUE #613: Fix image pushing in travis
Something has happened such that the automated image push at the end of the automated travis runs of master is broken. It fails and this causes our travis banner to show that we are in an error'd state.

We should be passing one way or the other. If that codepath is broken we need to evaluate if we should be fixing that or if requirements have changed and we need to remove it.

stevesloka:
Updates how images are built to enable `master` branch builds as well as `tagged` releases. 

- Fixes #613

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,31 +8,31 @@ services:
 
 install: true
 
-script: 
-  - VERBOSE=true make test int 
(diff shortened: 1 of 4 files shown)


stevesloka:
superseded by #642

PR #626, opened 2019-03-11
Show list of images used for upstream Kubernetes E2E tests

### LINKED ISSUE #604: Show list of tests in upstream image
Would like a way to show a list of images that are used in the upstream e2e tests to have an understanding of what is required and it's current location. 

The command could look something like this:
```
$ sonobuoy e2e images --list
```

stevesloka:
Allows users to see what images are used in upstream e2e tests by adding a new command: `images e2e`.

- Fixes #604

**Release note**:
```
release-note: Adds a new command `images e2e` which shows all upstream images used for E2E plugin tests. 

```

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -21,6 +21,8 @@ import (
 	"io/ioutil"
 	"strings"
 
+	"github.com/heptio/sonobuoy/pkg/image"
+
 	"gopkg.in/yaml.v2"
 
 	ops "github.com/heptio/sonobuoy/pkg/client"
@@ -34,6 +36,7 @@ const (
 	namespaceFlag       = "namespace"
 	sonobuoyImageFlag   = "sonobuoy-image"
 	imagePullPolicyFlag = "image-pull-policy"
+	pluginFlag          = "plugin"
 )
 
 // AddNamespaceFlag initialises a namespace flag.
@@ -73,11 +76,11 @@ func AddKubeConformanceImage(image *string, flags *pflag.FlagSet) {
 }
 
 // AddKubeConformanceImageVersion initialises an image version flag.
-func AddKubeConformanceImageVersion(imageVersion *ConformanceImageVersion, flags *pflag.FlagSet) {
+func AddKubeConformanceImageVersion(imageVersion *image.ConformanceImageVersion, flags *pflag.FlagSet) {
 	help := "Use default Conformance image, but override the version. "
 	help += fmt.Sprintf("Default is 'auto', which will be set to your cluster's version if detected, erroring otherwise.")
 
-	*imageVersion = ConformanceImageVersionAuto
+	*imageVersion = image.ConformanceImageVersionAuto
 	flags.Var(imageVersion, "kube-conformance-image-version", help)
 }
 
@@ -85,6 +88,12 @@ func AddKubeConformanceImageVersion(imageVersion *ConformanceImageVersion, flags
 func AddKubeconfigFlag(cfg *Kubeconfig, flags *pflag.FlagSet) {
 	// The default is the empty string (look in the environment)
 	flags.Var(cfg, "kubeconfig", "Path to explicit kubeconfig file.")
+}
+
+// AddPluginFlag describes which plugin's images to interact with
+func AddPluginFlag(cfg *string, flags *pflag.FlagSet) {
+	// The default is 'e2e' since it's the only plugin enabled at the moment
+	flags.StringVarP(cfg, pluginFlag, "p", "e2e", "Describe which plugin's images to interact (Valid plugins are 'e2e').")
 }
 
 // AddSonobuoyConfigFlag adds a SonobuoyConfig flag to the provided command.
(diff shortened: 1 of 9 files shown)


pkg/image/imageversion.go
```diff
@@ -106,7 +106,7 @@ func (c *ConformanceImageVersion) Get(client discovery.ServerVersionInterface) (
 	return string(*c), nil
 }
 
-func validateVersion(v string) (*version.Version, error) {
+func ValidateVersion(v string) (*version.Version, error) {
```
johnSchnake: Doesn't seem like this needs exported.

Gopkg.lock
```diff
@@ -688,6 +688,14 @@
   pruneopts = "UT"
   revision = "e5bc2a7bbbdf78f2a331b8c23838d1bbb1c6b40a"
 
+[[projects]]
+  branch = "master"
+  digest = "1:d6a310dcbc14bd1fa3b2bf336252ccf1f49143d511cfb8bdb8d1e6e8bb97fe25"
+  name = "k8s.io/kubernetes"
+  packages = ["test/utils/image"]
```
johnSchnake: Why are we adding this to the vendor if its not used? Maybe leftover from dev?

cmd/sonobuoy/app/images.go
```diff
@@ -0,0 +1,106 @@
...
+	return flagset
+}
+
+func NewCmdImages() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:       "images",
+		Short:     "Manage images used in a plugin. Supported plugins are: 'e2e'",
+		Run:       getImages,
+		ValidArgs: []string{"e2e"},
+		Args:      matchAll(cobra.MinimumNArgs(1), cobra.OnlyValidArgs),
```
johnSchnake: Would it be better to make `--plugin` a flag that defaults to `e2e`?

Then the user could provide 
```
sonobuoy images
sonobuoy images --plugin e2e // Same as the above
sonobuoy images --plugin foo // one day if another plugin supports this
```

The description of `--plugin` and which are supported values would then go onto the flag.

I tried it out and the cobra generated help is pretty confusing because of this. It doesnt show the `short` description during the error so it doesnt tell you at all what you really need to put for the arg (which to a user is odd because only one value is allowed anyways)

image

stevesloka: Ok we discussed this previously: https://github.com/heptio/sonobuoy/issues/604#issuecomment-467031910

I can use `--plugin` will fit better with `download`, `upload`, etc.

stevesloka:
All updated @johnSchnake, thanks for the feedback! =)

cmd/sonobuoy/app/images.go
johnSchnake: :-D Well what does that other guy know!

I was running an idea there but def didnt think of the implications with cobra; that works out worse than I would have thought.

The other way the original logic could work OK with cobra is to make it a subcommand, not an argument; that way cobra would suggest it as a subcommand, give additional help, etc. I'm surprised cobra doesn't make the valid args more clear tbh. It would have looked the same but played nicer with all the doc gen stuff.

stevesloka: Nah I think this works in the end.

cmd/sonobuoy/app/args.go
```diff
@@ -87,6 +90,12 @@ func AddKubeconfigFlag(cfg *Kubeconfig, flags *pflag.FlagSet) {
 	flags.Var(cfg, "kubeconfig", "Path to explicit kubeconfig file.")
 }
 
+// AddPluginFlag describes which plugin's images to interact
+func AddPluginFlag(cfg *string, flags *pflag.FlagSet) {
+	// The default is the empty string (look in the environment)
+	flags.StringVarP(cfg, pluginFlag, "p", "", "Describe which plugin's images to interact.")
```
johnSchnake: nit: `interact with`

johnSchnake: I'd also list the valid values here on the flag since that will be shown when/if they dont provide it.

cmd/sonobuoy/app/args.go
```diff
@@ -87,6 +90,12 @@ func AddKubeconfigFlag(cfg *Kubeconfig, flags *pflag.FlagSet) {
 	flags.Var(cfg, "kubeconfig", "Path to explicit kubeconfig file.")
 }
 
+// AddPluginFlag describes which plugin's images to interact
+func AddPluginFlag(cfg *string, flags *pflag.FlagSet) {
+	// The default is the empty string (look in the environment)
```
johnSchnake: Not sure I follow this comment; I dont see any code looking in the environment for a default value or anything.  Seems like the default should just be `e2e` since it is the only value for now and our main use case?

Gopkg.lock
```diff
@@ -738,6 +738,7 @@
     "k8s.io/client-go/tools/clientcmd",
     "k8s.io/client-go/tools/remotecommand",
     "k8s.io/client-go/util/exec",
+    "k8s.io/kubernetes/test/utils/image",
```
johnSchnake: No reason to have this right? I dont know why the solver would need it since you dont reference it and the list is just manually maintained for now?

PR #627, opened 2019-03-11
Allow ordered operations via plugin order

### LINKED ISSUE #39: Allow for user ordered operations in aggregator
Allow for collections to occur in a specific order A->B->C possibly by making everything a plugin.

johnSchnake:
This change introduces the concept of plugin order. It
is backwards compatible since the default value is 0
which can execute with without precondition.

A plugin can only be executed after all plugins of lower
order have reported results.

Fixes #39

**Release note**:
```
Add support for ordered operations via plugin "order" field.
```

### DIFF
--- a/pkg/plugin/aggregation/aggregator.go
+++ b/pkg/plugin/aggregation/aggregator.go
@@ -100,6 +100,22 @@ func (a *Aggregator) isComplete() bool {
 	return true
 }
 
+// currentOrder returns the order of lowest order plugin which doesn't have a result yet. If all have results then -1 is returned.
+func (a *Aggregator) currentOrder() int {
+	lowestWithoutResults := -1
+	set := false
+	for _, result := range a.ExpectedResults {
+		if _, ok := a.Results[result.ID()]; !ok {
+			if result.Order < lowestWithoutResults || !set {
+				lowestWithoutResults = result.Order
+				set = true
+			}
+		}
+	}
+
+	return lowestWithoutResults
+}
+
 func (a *Aggregator) isResultExpected(result *plugin.Result) bool {
 	_, ok := a.ExpectedResults[result.ExpectedResultID()]
 	return ok
--- a/pkg/plugin/aggregation/run.go
+++ b/pkg/plugin/aggregation/run.go
@@ -150,18 +150,46 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 	}()
 
 	// 4. Launch each plugin, to dispatch workers which submit the results back
-	for _, p := range plugins {
-		cert, err := auth.ClientKeyPair(p.GetName())
-		if err != nil {
-			return errors.Wrapf(err, "couldn't make certificate for plugin %v", p.GetName())
+	go func() {
+		pluginsRun := map[int]bool{}
+		for {
+			for index, p := range plugins {
+				// Plugin already run.
+				if pluginsRun[index] {
+					continue
+				}
+
+				// If unable to run plugin, wait.
+				if !canStartPlugin(aggr, p) {
+					logrus.Info("schnake: cant start plugin, continue")
+					continue
+				}
+
+				cert, err := auth.ClientKeyPair(p.GetName())
+				if err != nil {
+					logrus.Errorf("Couldnt make certificate for plugin %v: %v", p.GetName(), err)
+					return
+				}
+				logrus.WithField("plugin", p.GetName()).Info("Running plugin")
+				if err = p.Run(client, cfg.AdvertiseAddress, cert); err != nil {
+					logrus.Errorf("Error running plugin %v: %v", p.GetName(), err)
+					return
+				}
+				pluginsRun[index] = true
+				// Have the plugin monitor for errors
+				go p.Monitor(client, nodes.Items, monitorCh)
+			}
+
+			// Break when all plugins have been run.
+			if len(pluginsRun) == len(plugins) {
+				break
+			}
+
+			// Sleep for a shorter period before rechecking plugins.
+			time.Sleep(10 * time.Second)
 		}
-		logrus.WithField("plugin", p.GetName()).Info("Running plugin")
-		if err = p.Run(client, cfg.AdvertiseAddress, cert); err != nil {
-			return errors.Wrapf(err, "error running plugin %v", p.GetName())
-		}
-		// Have the plugin monitor for errors
-		go p.Monitor(client, nodes.Items, monitorCh)
-	}
+	}()
+
 	// 5. Have the aggregator plumb results from each plugins' monitor function
 	go aggr.IngestResults(monitorCh)
 
@@ -187,6 +215,11 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 			return nil
 		}
 	}
+}
+
+func canStartPlugin(a *Aggregator, p plugin.Interface) bool {
+	logrus.Infof("schnake: p.order %v current order: %v", p.GetOrder(), a.currentOrder())
+	return p.GetOrder() <= a.currentOrder()
 }
 
 // Cleanup calls cleanup on all plugins
--- a/pkg/plugin/driver/base.go
+++ b/pkg/plugin/driver/base.go
@@ -66,6 +66,11 @@ func (b *Base) GetName() string {
 	return b.Definition.Name
 }
 
+// GetIOrder returns the order of this Job plugin.
+func (b *Base) GetOrder() int {
+	return b.Definition.Order
+}
+
 // GetSecretName gets a name for a secret based on the plugin name and session ID.
 func (b *Base) GetSecretName() string {
 	return fmt.Sprintf("sonobuoy-plugin-%s-%s", b.GetName(), b.GetSessionID())
--- a/pkg/plugin/driver/job/job.go
+++ b/pkg/plugin/driver/job/job.go
@@ -23,7 +23,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kuberuntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
@@ -63,7 +63,10 @@ func NewPlugin(dfn plugin.Definition, namespace, sonobuoyImage, imagePullPolicy 
 // a Job only launches one pod, only one result type is expected.
 func (p *Plugin) ExpectedResults(nodes []v1.Node) []plugin.ExpectedResult {
 	return []plugin.ExpectedResult{
-		plugin.ExpectedResult{ResultType: p.GetResultType()},
+		plugin.ExpectedResult{
+			ResultType: p.GetResultType(),
+			Order:      p.Definition.Order,
+		},
 	}
 }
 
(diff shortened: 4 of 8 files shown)


pkg/plugin/aggregation/run.go
```diff
@@ -150,18 +150,46 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
...
-		if err != nil {
-			return errors.Wrapf(err, "couldn't make certificate for plugin %v", p.GetName())
-		}
-		logrus.WithField("plugin", p.GetName()).Info("Running plugin")
-		if err = p.Run(client, cfg.AdvertiseAddress, cert); err != nil {
-			return errors.Wrapf(err, "error running plugin %v", p.GetName())
+	go func() {
+		pluginsRun := map[int]bool{}
+		for {
+			for index, p := range plugins {
```
johnSchnake: Previously we just synchronously started each plugin and now we just iterate over them checking if we can start them or not.
 - uses a map to check if we started each plugin (just using index of the plugin as a key)
 - didnt want to alter the slice of plugins since that is used elsewhere

pkg/plugin/aggregation/run.go
```diff
@@ -150,18 +150,46 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
...
+				go p.Monitor(client, nodes.Items, monitorCh)
+			}
+
+			// Break when all plugins have been run.
+			if len(pluginsRun) == len(plugins) {
+				break
+			}
+
+			// Sleep for a shorter period before rechecking plugins.
+			time.Sleep(10 * time.Second)
```
johnSchnake: Arbitrary; just created as poc. Would make a constant but honestly 10s seems fine to me since its just doing some iterations on a map its not an expensive operation to do every few sec. Especially since the # of plugins is usually very small

pkg/plugin/aggregation/run.go
```diff
@@ -189,6 +217,11 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 	}
 }
 
+func canStartPlugin(a *Aggregator, p plugin.Interface) bool {
+	logrus.Infof("schnake: p.order %v current order: %v", p.GetOrder(), a.currentOrder())
```
johnSchnake: debug line

pkg/plugin/aggregation/run.go
```diff
@@ -189,6 +217,11 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 	}
 }
 
+func canStartPlugin(a *Aggregator, p plugin.Interface) bool {
```
johnSchnake: Even though the logic is small I wanted to put it here so that it would be easier to test/manipulate in the future.

johnSchnake:
@akutz what do you think about simple support of ordered operations like this?

See the ticket for the reasoning I had on doing a small version like this; tldr is that I think this unlocks a lot of options for users for minimal effort. I think implementing a full/robust dependency logic and linking runs together will be about 50x harder and (as of now) I've not heard of someone saying they need it (as opposed to just stating they needed ordered ops)

pkg/plugin/aggregation/run.go
stevesloka: Where do we keep status of each plugin? Could you create a struct to store all the bits about the plugin (i.e. current status, start time, etc).

pkg/plugin/aggregation/run.go
```diff
@@ -150,18 +150,46 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
...
-	for _, p := range plugins {
-		cert, err := auth.ClientKeyPair(p.GetName())
-		if err != nil {
-			return errors.Wrapf(err, "couldn't make certificate for plugin %v", p.GetName())
-		}
-		logrus.WithField("plugin", p.GetName()).Info("Running plugin")
-		if err = p.Run(client, cfg.AdvertiseAddress, cert); err != nil {
-			return errors.Wrapf(err, "error running plugin %v", p.GetName())
+	go func() {
+		pluginsRun := map[int]bool{}
```
akutz: Hi @johnSchnake,

Suggesting the following:

```go
pluginsRun := map[int]struct{}
```

Anonymous (or even empty) structs do not consume any memory in Go. If you're using a map as an addressable flag set, anonymous structs should be used as the value type. Dave Cheney wrote a great piece on this a [few years back](https://dave.cheney.net/2014/03/25/the-empty-struct).

pkg/plugin/aggregation/run.go
```diff
@@ -150,18 +150,46 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
...
+			for index, p := range plugins {
+				// Plugin already run.
+				if pluginsRun[index] {
+					continue
+				}
+
+				// If unable to run plugin, wait.
+				if !canStartPlugin(aggr, p) {
+					logrus.Info("schnake: cant start plugin, continue")
+					continue
```
akutz: Hi @johnSchnake,

This `continue` has no sleep, and neither does the loop's beginning. Thus if the plug-in is unable to start, and there's only one plug-in, it will immediately try again. Is this desired?

pkg/plugin/aggregation/run.go
```diff
@@ -150,18 +150,46 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
...
-		cert, err := auth.ClientKeyPair(p.GetName())
-		if err != nil {
-			return errors.Wrapf(err, "couldn't make certificate for plugin %v", p.GetName())
-		}
-		logrus.WithField("plugin", p.GetName()).Info("Running plugin")
-		if err = p.Run(client, cfg.AdvertiseAddress, cert); err != nil {
-			return errors.Wrapf(err, "error running plugin %v", p.GetName())
+	go func() {
+		pluginsRun := map[int]bool{}
+		for {
```
akutz: Hi @johnSchnake,

You may want to consider using a Go context and channels to select on a channel buffered to the size of the plug-ins and `ctx.Done` in case the operation is cancelled.

pkg/plugin/aggregation/run.go
akutz: Hi @johnSchnake,

Just curious, why isn't this a private function on `*Aggregator`?

pkg/plugin/driver/base.go
```diff
@@ -66,6 +66,11 @@ func (b *Base) GetName() string {
 	return b.Definition.Name
 }
 
+// GetIOrder returns the order of this Job plugin.
```
akutz: Hi @johnSchnake,

`// GetIOrder` -> `// GetOrder`

Also, consider rewriting the function signature as `func (b Base) GetOrder() int` since there is no reason for the function to be able to write any values to the address of `b`. A read on a copy will suffice and offers better protection.

pkg/plugin/driver/base.go
```diff
@@ -66,6 +66,11 @@ func (b *Base) GetName() string {
...
+func (b *Base) GetOrder() int {
+	return b.Definition.Order
+}
+
 // GetSecretName gets a name for a secret based on the plugin name and session ID.
 func (b *Base) GetSecretName() string {
```
akutz: Hi @johnSchnake,

Consider rewriting the function signature as func `(b Base) GetSecretName() string` since there is no reason for the function to be able to write any values to the address of `b`. A read on a copy will suffice and offers better protection.

pkg/plugin/driver/job/job.go
```diff
@@ -63,7 +63,10 @@ func NewPlugin(dfn plugin.Definition, namespace, sonobuoyImage, imagePullPolicy
 // a Job only launches one pod, only one result type is expected.
 func (p *Plugin) ExpectedResults(nodes []v1.Node) []plugin.ExpectedResult {
```
akutz: Hi @johnSchnake,

Consider rewriting the function signature as func `func (p Plugin) ExpectedResults(nodes []v1.Node) []plugin.ExpectedResult` since there is no reason for the function to be able to write any values to the address of `p`. A read on a copy will suffice and offers better protection.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -100,6 +100,22 @@ func (a *Aggregator) isComplete() bool {
 	return true
 }
 
+// currentOrder returns the order of lowest order plugin which doesn't have a result yet. If all have results then -1 is returned.
+func (a *Aggregator) currentOrder() int {
```
akutz: Hi @johnSchnake,

Please consider rewriting the function signature as `func (a Aggregator) currentOrder()` since there is no reason for the function to be able to write any values to the address of `a`. A read on a copy will suffice and offers better protection.

I do not see where the result Order is set. It seems to be related to the plug-in order. If plug-ins have a defined order, and the number of plug-ins is known, does that also mean the number of expected results are known? If so, then you could also maintain an analogue slice of addresses of the expected results that is sorted by order and use a `Search` function from the stdlib `sort` package. It may be overkill to do so though. I don't know whether or not the list is ever sufficiently large that keeping a second copy of addresses in memory is worth a faster search result.

pkg/plugin/aggregation/run.go
johnSchnake: @akutz , thanks. Will do. I'm aware of the pattern, I think this is just a bit left over from iterations on the theme.

pkg/plugin/aggregation/run.go
johnSchnake: 👍 no, thanks for the find. When fixing up the PR I'll be sure to move the sleep (or something) around appropriately.

pkg/plugin/aggregation/run.go
johnSchnake: No reason; 6 on one hand half a dozen on the other IMO. Under the hood its all the same so it just came down to my random train of thought at the moment of writing it.

pkg/plugin/driver/base.go
johnSchnake: 👍 on the rename.

As far as the pointer, I agree with the unencessity since it is a read. However, go best practice states that you shouldn't have a [mixture](https://tour.golang.org/methods/8) of pointer/value receivers for a given type. Since other pointer receiver methods exist (and I wasn't planning on changing them all, nor do I think I could with all the implications because this is used by all the other plugin types). (also fwiw I hate linking to the tour as a 'best practice' reference but I can't find the original link I'm thinking of. We can talk more about this if it is a serious concern to you)

pkg/plugin/aggregation/aggregator.go
johnSchnake: The list is usually very small; default is 2 and I could realistically imagine 5-10 being common for someone who is trying to run all their own custom, small plugins on each run. Even in a weird application I can't see it being an orders of magnitude higher to where we are concerned with the performance of a sort.

pkg/plugin/aggregation/run.go
johnSchnake: Hi @akutz , that is an interesting idea but would that actually be necessary here since all this is running in containers in the cluster? A "cancel" in that context would mean the user is stopping the run and we are just blowing up the container entirely, right? Or is there another way you can see that being used?

pkg/plugin/driver/base.go
akutz: Hi @johnSchnake,

I'm not overly concerned about this, but as long as there's a clear pattern to when things read and when things write, I think it's fine. All of GoVmomi does it. Read functions use byCopy receivers, and write functions use byAddress receivers.

I don't feel strongly either way. Just noticed it.

pkg/plugin/aggregation/run.go
akutz: Hi @johnSchnake,

I didn't have any clear idea on how this was being used. When I see the `for {` pattern I usually ask myself if it should be cancellable, and if so, does it make sense to use a Go context. You know the call stack better than I do, so please do whatever you think makes sense.

johnSchnake:
@akutz Thank you so much for both the high and low-level review. I'm going to consolidate what I've learned and done into a design doc to clarify some design details before finalizing things in this PR (or a related one).

A little odd to do a POC, then design doc for what is such a small feature in terms of LOC but I think something like this is pretty important to get right the first time.

 💯 for your time and detail.

PR #632, opened 2019-03-13
Download e2e & push to private registry

### LINKED ISSUE #606: Download and load conformance images to private registry
To support air-gapped environments would like to be able to download all of the images required so that I can push to a private registry and run an e2e scan. 

The command could look something like this:
```
$ sonobuoy e2e images https://registry --upload 
```

//cc #605

stevesloka:
New commands allows the following:
 - "pull": Docker pull all E2E images based upon the cluster version (v1.13 or greater) to the local docker client
- "download": Once "pull" has happened, export all images as a tar file
- "push": Tag all images with private registry then and push to a registry
- "delete": Deletes all images that were pulled from local docker client

- Fixes #606 #604

This supersedes #626 

**Release note**:
```
New command allows users to download all E2E images based upon the cluster version (v1.13 or greater), then tag them and push to a local registry. 
```


cmd/sonobuoy/app/images.go
```diff
@@ -0,0 +1,293 @@
...
+	// Sub command
+	subCmd := &cobra.Command{
+		Use:   "download",
+		Short: "Download images to local Docker client for a specific plugin",
+		Run:   downloadImages,
+		Args:  cobra.ExactArgs(0),
+	}
+	subCmd.Flags().AddFlagSet(ImagesFlagSet(&imagesflags))
+	subCmd.MarkFlagRequired(pluginFlag)
+	subCmd.MarkFlagRequired(e2eRegistryConfigFlag)
```
johnSchnake: For downloading they shouldn't need this; for the upload it makes sense though.

cmd/sonobuoy/app/images.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// Sub command
+	subCmd := &cobra.Command{
+		Use:   "download",
+		Short: "Download images to local Docker client for a specific plugin",
+		Run:   downloadImages,
+		Args:  cobra.ExactArgs(0),
+	}
+	subCmd.Flags().AddFlagSet(ImagesFlagSet(&imagesflags))
+	subCmd.MarkFlagRequired(pluginFlag)
```
johnSchnake: Why is this marked required? This causes it to error even when trying to just use the default value.

cmd/sonobuoy/app/images.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+
+		privateImages, err := privateRegistry.GetImageConfigs()
+		if err != nil {
+			errlog.LogError(errors.Wrap(err, "couldn't get images for version"))
+			os.Exit(1)
+		}
+
+		ctx := context.Background()
+		cli, err := client.NewClientWithOpts(client.WithVersion("1.37"))
```
johnSchnake: 1.37? I'm sure its some sane default but we should const that.

cmd/sonobuoy/app/images.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+}
+
+func pushImage(ctx context.Context, cli *client.Client, img image.Config) {
+	auth := types.AuthConfig{
+		Username: imagesflags.username,
+		Password: imagesflags.password,
+	}
+
+	fmt.Println("pass: ", imagesflags.password)
```
johnSchnake: debug?

cmd/sonobuoy/app/images.go
```diff
@@ -0,0 +1,293 @@
...
+
+	authBytes, _ := json.Marshal(auth)
```
johnSchnake: Should handle err

cmd/sonobuoy/app/images.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// Show status
+	err = streamDockerMessages(out)
+	if err != nil {
+		errlog.LogError(errors.Wrapf(err, "error downloading image: %v", img.GetE2EImage()))
+	}
+	// Wait for all data to complete
+	_, err = ioutil.ReadAll(out)
+	if err != nil {
+		errlog.LogError(errors.Wrapf(err, "error downloading image: %v", img.GetE2EImage()))
```
johnSchnake: This pattern is used a few times in the code; could you elaborate a bit on it?

An error on this `ioutil.ReadAll` would indicate the read failed, not the image failed to download. It seems like we'd be doing ourselves a disservice to err the download (or tag or push) if this read fails. And if we don't even want the data that we are reading, why are we doing the read at all?

cmd/sonobuoy/app/images.go
```diff
@@ -0,0 +1,293 @@
...
+	// Check if the e2e file exists
+	if _, err := os.Stat(imagesflags.e2eRegistryConfig); os.IsNotExist(err) {
+		errlog.LogError(errors.Errorf("file does not exist or cannot be opened: %v", imagesflags.e2eRegistryConfig))
+		os.Exit(1)
+	}
+
+	// Check if username if specified for Registry Auth
+	if len(imagesflags.username) > 0 {
+		fmt.Print("Registry password: ")
+		bytePassword, err := terminal.ReadPassword(int(syscall.Stdin))
```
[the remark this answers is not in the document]
stevesloka: ENV isn't really a secure place to put stuff which is why I had it prompt, it's never written anywhere. Could add though if you feel strongly we need it as ENV.

cmd/sonobuoy/app/images.go
stevesloka: Depends on if we change the functionality. I don't think most air-gapped environments are truly air-gapped, they just don't have internet access.

cmd/sonobuoy/app/images.go
stevesloka: It could come out, I don't like how much we integrate a specific plugin into Sonobuoy, one plugin is 1st class, and any other's don't have any cli integration. I can make it optional.

cmd/sonobuoy/app/images.go
stevesloka: The read is reading the file stream. If this fails then the download failed. This blocks to wait for the image to finish, otherwise, they all go in parallel and we don't want 20 images downloading concurrently.

stevesloka:
@johnSchnake I'm going to collapse all these into a single PR. Since they are all related I don't want to run across 3 rebases on the same code. I'll integrate the feedback, heads up shortly.

cmd/sonobuoy/app/images.go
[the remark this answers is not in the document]
johnSchnake: I was under the impression (as a non-security expert) that the env was a _good_ place to put creds. Not that "other people doing it" makes it right, but:

 - [AWS](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) does it for their cli and other tools. They do cite that it is particularly useful for scripting and other use cases though.
 - [SauceLabs](https://wiki.saucelabs.com/display/DOCS/Best+Practice%3A+Use+Environment+Variables+for+Authentication+Credentials) lists is as a best practice

ENV isn't providing security per-se since it is unencrypted but the alternative for a lot of workflows (e.g. CI or Docker) would be setting the values as env values and then (if you require them on stdin) they'd have to just echo them through a pipe and hope that is handled correctly (I've seen password libraries including the go stdlib have problems on different platforms w.r.t. pipes getting broken or not being handled at all or causing other issues).

I guess in some situations you're right though: if I'm a user standing in front of the keyboard, _having_ to write the password to the env would be less secure since it would be visible. They _could_ write it to a file and source it or something like that, but its non-ideal for any user.

Thoughts on the dual support or do you think that is still actively a bad idea to support at all?

cmd/sonobuoy/app/images.go
johnSchnake: But if you can' reach the internet you _cant_ do the download/upload at the same time and still have to walk it to the other network? In some cases you do have one particular machine that can reach both, but I guess I dont have the experience to know which topology is more common.

pkg/image/manifest.go
```diff
@@ -0,0 +1,97 @@
...
+		PrivateRegistry:       "gcr.io/k8s-authenticated-test",
+		SampleRegistry:        "gcr.io/google-samples",
+	}
+
+	// Load in a config file
+	if repoConfig != "" {
+
+		fileContent, err := ioutil.ReadFile(repoConfig)
+		if err != nil {
+			panic(fmt.Errorf("Error reading '%v' file contents: %v", repoConfig, err))
```
johnSchnake: I know this was copied from the upstream, but since we are using this logic within Sonobuoy I think we should avoid the panic (which upstream in testing is OK, on a CLI we shouldn't).

I'd change these panics to print the errors and exit 1.

cmd/sonobuoy/app/images.go
```diff
@@ -19,45 +19,108 @@ package app
...
 var imagesflags imagesFlags
 
-type imagesFlags struct {
-	e2eRegistryConfig string
-	plugin            string
-	kubeconfig        Kubeconfig
-
-	imagesflags *pflag.FlagSet
-}
+const dockerClientVersion = "1.37"
```
johnSchnake: Do you know how much support above/below this there is?

cmd/sonobuoy/app/images.go
```diff
@@ -101,3 +164,249 @@ func getImages(cmd *cobra.Command, args []string) {
...
+				bytePassword, err := terminal.ReadPassword(int(syscall.Stdin))
+				if err != nil {
+					errlog.LogError(errors.Wrap(err, "couldn't get password from user"))
+					os.Exit(1)
+				}
+
+				imagesflags.password = string(bytePassword)
+				fmt.Print("\n")
+			} else {
+				imagesflags.password = envPassword
```
johnSchnake: nit: just move this up to the line grabbing it with `viper` so you can remove the entire `else` block/path.

stevesloka: What if it's not passed via env and I want to prompt the user?

cmd/sonobuoy/app/images.go
stevesloka: Honestly no, I'm spinning up a windows machine to validate this works there as well.

pkg/image/image.go
```diff
@@ -0,0 +1,159 @@
...
+func SaveToTar(ctx context.Context, cli *client.Client, images []string, filepath string) error {
+	file, err := os.Create(filepath)
+	if err != nil {
+		return errors.Wrapf(err, "Could not create tarball file '%s'", filepath)
+	}
+	defer file.Close()
+
+	out, err := cli.ImageSave(ctx, images)
+	if err != nil {
+		return errors.Wrap(err, "error saving images to tar")
```
johnSchnake: I hit this error when trying to run it. I am on version 1.39 of the docker api though locally.

```
./sonobuoy images download
ERRO[0000] couldn't save images to tar: error saving images to tar: Error response from daemon: reference does not exist
```

cmd/sonobuoy/app/images.go
johnSchnake: I commented at the place the err hit me; but this does seem like it will cause at least some problems. I couldnt download images even though a pull worked. On version 1.39 locally.

johnSchnake: I think we need to think about this particular point. Maybe there is a good way to grab the local version and it would generally work if they aren't really breaking their api much. Hopefully we dont need this to be manually fed in; maybe we can even try shelling out and checking it? That would also ensure the docker daemon is running.

pkg/image/image.go
stevesloka: Seems like you don't have all the images downloaded?

stevesloka: Worked on linux as well without changes. Hmmm

cmd/sonobuoy/app/images.go
stevesloka: If you could pull then seems like the version issue is not the problem. Are you sure you got all the images downloaded? It took me a good 10mins to gen the tar file.

johnSchnake: I also confirmed that when modified it to 1.39 the download worked for me without other code modifications.

stevesloka: Take out the version all together, does it work?

johnSchnake: Surprisingly yes it seems to. It is taking a while to run but in the error case it failed right away. Just takes a while to download 7g. *heads to look at the docker docs

stevesloka: It worked for a while for me without. Then at some point, it started to error. From my understanding, we're specifying the version of the api that we want.

johnSchnake: Looks like we should just be using `cli, err := client.NewClientWithOpts(client.FromEnv)` which will handle defaults and also allow overrides for users to specify their versions if there are issues.

pkg/image/image.go
johnSchnake: Just resolving this conversation; the thread above about hardcoding the version # has the relevant info. It isnt an OS issue but a docker api version one.

cmd/sonobuoy/app/images.go
stevesloka: I double checked how kind did it and they shell out to the local client. May switch over to that which also removes the need for any type of auth: https://github.com/kubernetes-sigs/kind/blob/master/pkg/container/docker/pull.go#L30:6

stevesloka:
@johnSchnake I refactored everything modeled after kind. PTAL! =)

cmd/sonobuoy/app/images.go
```diff
@@ -22,46 +22,102 @@ import (
...
-	flagset := pflag.NewFlagSet("images", pflag.ExitOnError)
-	AddPluginFlag(&cfg.plugin, flagset)
-	AddKubeconfigFlag(&cfg.kubeconfig, flagset)
-	cfg.imagesflags = flagset
-	return flagset
+type imagesFlags struct {
+	e2eRegistryConfig  string
+	plugin             string
+	username           string
+	password           string
```
johnSchnake: No more need for username/password, right?

cmd/sonobuoy/app/images.go
```diff
@@ -22,46 +22,102 @@ import (
...
+	imagesSaveFileName string
+	downloadTar        bool
```
johnSchnake: This isn't used

cmd/sonobuoy/app/images.go
```diff
@@ -22,46 +22,102 @@ import (
...
-	AddPluginFlag(&cfg.plugin, flagset)
-	AddKubeconfigFlag(&cfg.kubeconfig, flagset)
-	cfg.imagesflags = flagset
-	return flagset
+type imagesFlags struct {
+	e2eRegistryConfig  string
+	plugin             string
+	username           string
+	password           string
+	imagesSaveFileName string
```
johnSchnake: This isnt used either; I'm guessing you thought about having it as a flag and cut it out?

I do think that would be a reasonable thing to do so they dont have to find what file it was after the fact (same small UX issue that retrieve used to have and we just decided to echo it out to stdout so it could be grabbed)

cmd/sonobuoy/app/images.go
```diff
@@ -22,46 +22,102 @@ import (
...
 
-func getImages(cmd *cobra.Command, args []string) {
+func listImages(cmd *cobra.Command, args []string) {
 
 	switch imagesflags.plugin {
 	case "e2e":
 
+		if len(imagesflags.e2eRegistryConfig) > 0 {
+			// Check if the e2e file exists
+			if _, err := os.Stat(imagesflags.e2eRegistryConfig); os.IsNotExist(err) {
```
johnSchnake: Based on the err message you should just check if `err != nil` here; otherwise we're not catching any of the "cannot be opened" cases.

pkg/image/docker/tag.go
```diff
@@ -0,0 +1,47 @@
...
+// Tag tags an image, retrying up to retries times
+func Tag(src, dest string, retries int) error {
+	log.Infof("Tagging image: %s as %s ...", src, dest)
+	err := exec.Command("docker", "tag", src, dest).Run()
+	// retry tagging up to retries times if necessary
+	if err != nil {
+		for i := 0; i < retries; i++ {
+			time.Sleep(time.Second * time.Duration(i+1))
+			log.WithError(err).Infof("Trying again to tag image: %s as %s ...", src, dest)
+			// TODO(stevesloka): add some backoff / sleep?
```
johnSchnake: Outdated; looks like you opted for a linear backoff? Only one retry though so its not that big a deal either way.

pkg/image/docker/save.go
```diff
@@ -0,0 +1,39 @@
...
+	log.Info("Saving images: ...")
+
+	// Build out docker command
+	args := append([]string{"save"}, images...)
+	args = append(args, "--output", filename)
+
+	err := exec.Command("docker", args...).Run()
+	// retry pushing up to retries times if necessary
+	if err != nil {
+		log.WithError(err).Infof("Failed to save images.")
```
johnSchnake: nit: I think the pattern should probably be that we just return the wrapped error rather than logging it _and_ returning it?

That way the caller can decide how to handle it (log or not) and we don't end up with repetitive logs where each level logs and returns.

pkg/image/docker/pull.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	// otherwise try to pull it
+	return Pull(image, retries)
+}
+
+// Pull pulls an image, retrying up to retries times
+func Pull(image string, retries int) error {
+	log.Infof("Pulling image: %s ...", image)
+	err := exec.Command("docker", "pull", image).Run()
+	// retry pulling up to retries times if necessary
```
johnSchnake: You have this cmder + loop logic a number of times. I'd suggest extracting the "loop a cmd with retry" logic to a single place and just call each time with the cmder (or args and make the cmder in that func)

cmd/sonobuoy/app/images.go
```diff
@@ -22,46 +22,102 @@ import (
...
 
-type imagesFlags struct {
-	e2eRegistryConfig string
-	plugin            string
-	kubeconfig        Kubeconfig
-
-	imagesflags *pflag.FlagSet
-}
+const numDockerRetries = 1
+)
```
johnSchnake: left over; prevents build.

cmd/sonobuoy/app/images.go
```diff
@@ -101,3 +157,195 @@ func getImages(cmd *cobra.Command, args []string) {
...
+		for k, v := range upstreamImages {
+			privateImg := privateImages[k]
+			err = docker.Tag(v.GetE2EImage(), privateImg.GetE2EImage(), numDockerRetries)
+			if err != nil {
+				errlog.LogError(errors.Wrapf(err, "couldn't tag image: %v", v.GetE2EImage()))
+			}
+
+			err = docker.Push(privateImg.GetE2EImage(), numDockerRetries)
+			if err != nil {
+				errlog.LogError(errors.Wrapf(err, "couldn't push image: %v", v.GetE2EImage()))
```
johnSchnake: For these errors we may want to look at how we log them. When shelling out to other commands you get a certain type of error which doesnt print useful info, just the exit status:
```
ERRO[0005] couldn't push image: gcr.io/google-samples/gb-redisslave:v3: exit status 1
```
Stderr has the useful info but we may inspect the error and see if there is another piece of useful data there instead.

pkg/image/docker/tag.go
```diff
@@ -0,0 +1,47 @@
...
+	"time"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/heptio/sonobuoy/pkg/image/exec"
+)
+
+// Tag tags an image, retrying up to retries times
+func Tag(src, dest string, retries int) error {
+	log.Infof("Tagging image: %s as %s ...", src, dest)
```
johnSchnake: An optimization but if the src==dest we should early return. Those are effectively images they dont want to retag/push.

If it makes more sense then maybe it should be an error (or a special error/type) so that `push` knows not to try and push that since it didnt get retagged?

```
INFO[0043] Tagging image: gcr.io/google-samples/gb-frontend:v6 as gcr.io/google-samples/gb-frontend:v6 ...
```

pkg/image/docker/save.go
johnSchnake: Example here (reminder this is just a nit and maybe we actually prefer this?):
```
INFO[0043] Pushing image: gcr.io/google-samples/gb-frontend:v6 ...
INFO[0047] Trying again to push image: gcr.io/google-samples/gb-frontend:v6 ...  error="exit status 1"
INFO[0050] Failed to push image: gcr.io/google-samples/gb-frontend:v6  error="exit status 1"
ERRO[0050] couldn't push image: gcr.io/google-samples/gb-frontend:v6: exit status 1
```

It is a little verbose but it does I suppose make it very clear what was happening. Again, the error message itself is useless since its just the code.

cmd/sonobuoy/app/images.go
johnSchnake: I think the solution to this is to run using `exec.RunLoggingOutputOnFailed`. I got the following output which is much more clear:

```
INFO[0026] Trying again to push image: docker.io/schnake/volume/nfs:1.0 ...  error="exit status 1"
ERRO[0027] failed with:
ERRO[0027] The push refers to repository [docker.io/schnake/volume/nfs]
ERRO[0027] 37970ad80d97: Preparing
ERRO[0027] 37970ad80d97: Preparing
ERRO[0027] 1448ce2d0d0c: Preparing
ERRO[0027] 2cedbfe57c8f: Preparing
ERRO[0027] 4d03698a8437: Preparing
ERRO[0027] 1d31b5806ba4: Preparing
ERRO[0027] denied: requested access to the resource is denied
```

cmd/sonobuoy/app/images.go
```diff
@@ -22,46 +22,102 @@ import (
...
+	}
+	AddKubeconfigFlag(&imagesflags.kubeconfig, downloadCmd.Flags())
+	AddPluginFlag(&imagesflags.plugin, downloadCmd.Flags())
+
+	// Push command
+	pushCmd := &cobra.Command{
+		Use:   "push",
+		Short: "Pushes images to docker registry for a specific plugin",
+		Run:   pushImages,
+		Args:  cobra.ExactArgs(0),
```
johnSchnake: Just an idea (and maybe this will just be a future ticket) but maybe in the future we allow N args to these commands so that users can push/pull just the list of tests they want? Either just the ones from a given registry list (e2e repo vs sample repo) or explicit image names or something?

cmd/sonobuoy/app/images.go
```diff
@@ -101,3 +157,195 @@ func getImages(cmd *cobra.Command, args []string) {
...
+
+func pushImages(cmd *cobra.Command, args []string) {
+
+	switch imagesflags.plugin {
+	case "e2e":
+
+		if len(imagesflags.e2eRegistryConfig) > 0 {
+			// Check if the e2e file exists
+			if _, err := os.Stat(imagesflags.e2eRegistryConfig); os.IsNotExist(err) {
+				errlog.LogError(errors.Errorf("file does not exist or cannot be opened: %v", imagesflags.e2eRegistryConfig))
```
johnSchnake: same nit as the other line like this; by checking the `isNotExist` we aren't really checking the `cannot be opened` case. Maybe just remove that wording from the error in each case?

stevesloka: The os.Stat check if we can open I think. We're creating the file and overwriting each time as well.

cmd/sonobuoy/app/images.go
stevesloka: I think future, it's out of scope for this issue.

pkg/image/docker/tag.go
stevesloka: I worry about erroring out here too much. There are so many images you end up having to do alot of manual effort to get past the errors. If you retag the same image, it's essentially a noop on the local client.

pkg/image/docker/pull.go
stevesloka: It is interfaced out in the `exec` package. You think we need a further abstraction on top of that interface?

pkg/image/docker/save.go
stevesloka: I think the retry is good to show that it's trying again. Let me see if I can get the real error returned like you suggested in another comment.

pkg/image/docker/tag.go
stevesloka: There's no exponential backoff, it just tries once, then retry again.  A real exponential would try, fail, try again and wait a minute, then fail, wait 5 mins, etc.

cmd/sonobuoy/app/images.go
stevesloka: This should check if the file doesn't exist or if it cannot be opened.

pkg/image/docker/tag.go
johnSchnake: I know its not an exponential backoff; you have a linear back off 2 lines above:
```
time.Sleep(time.Second * time.Duration(i+1))
```

pkg/image/docker/pull.go
johnSchnake: I'm just saying that we're just repeating ourselves if in push/pull/tag are writing the same loop/retry/backoff logic and you could DRY that up a bit if you put the loop into a single method and just call `execWithRetries` with your args (or something like that)

cmd/sonobuoy/app/images.go
johnSchnake: Sorry, I'm not sure I understand your response.

I agree it should check if the file exists or we cant open it. But if you check `os.IsNotExist` then you're only checking whether or not the file exists. Other sorts of access issues wont hit this (and the other similar) blocks. Right?

pkg/image/docker/tag.go
johnSchnake: Sorry if I wasnt clear; what I was saying that this _function_ could return a sentinel error (or error type/interface) so that it exits early and doesnt shell out to docker. Then the calling function would handle it appropriately.

E.g.
```
func tagAll(images []string){
  for _,i:=range images{
    err := tag(src,dest,retries)
    switch err{
    case err==sameTagErr:
      log.Infof("Source and destination tags match; skipping...")
...
...
```

Shelling out to the local docker agent and retagging the image is a noop so thats not that big a deal; the messaging to the user "Tagging X as X" is just a little weird. 

The bigger issue is when you are then trying to push right after this: in 99.9% of all cases the user doesn't have authority to push to the default registry and it will fail. If we _know_ they aren't retagging image X, why go through the process of tagging X as X, pushing X then failing, retrying to push X, then erroring when all our retries fail.

In that flow we end up with 5 log lines for something we _knew_ was a failure (noop at best)

cmd/sonobuoy/app/images.go
johnSchnake: Agreed; I was just mentioning it as a point of interest. Wasn't sure if you had thought about that too. It will become slightly less of an issue if we can get those upstream images minimized.

stevesloka:
@johnSchnake I think I got everything resolved now. Please take a look. Thanks! =)

cmd/sonobuoy/app/images.go
```diff
@@ -101,3 +153,205 @@ func getImages(cmd *cobra.Command, args []string) {
...
+		}
+
+		fileName := image.GetTarFileName(version)
+		err = docker.Save(images, fileName)
+		if err != nil {
+			errlog.LogError(errors.Wrap(err, "couldn't save images to tar"))
+			os.Exit(1)
+		}
+
+		fmt.Println("Archive is: ", fileName)
```
johnSchnake: Although the prefix makes it easy to read I think the benefit of outputting this at all is being able to easily script around it. 

Without the prefix you just need to know the output is the file:
```
output=$(sonobuoy images download)
```

With the prefix you have to know the format of the output and use another tool like awk or cut:
```
output=$(sonobuoy images download | awk '{print $NF}'
```

Removing the prefix would also make it consistent with `sonobuoy retrieve`.

cmd/sonobuoy/app/images.go
```diff
@@ -101,3 +153,205 @@ func getImages(cmd *cobra.Command, args []string) {
...
+			os.Exit(1)
+		}
+
+		version, err := sbc.Version()
+		if err != nil {
+			errlog.LogError(errors.Wrap(err, "couldn't get Sonobuoy client"))
+			os.Exit(1)
+		}
+
+		upstreamImages, err := image.GetImages("", version)
```
johnSchnake: nit: Probably should const the empty string here so it has better meaning in the code. As a reader of the code I'd understand it is some default value, but I dont really know what. Maybe just a name like `const defaultE2ERegistries = ""`?

stevesloka:
@johnSchnake I did a bunch of refactoring. I added a Docker interface so we can stub out what happens there and then tested all the image functions (e.g. push, pull, etc) along with all the other comments.

johnSchnake:
@stevesloka Looks good to me! Just needs a rebase.  I really liked where it is now in terms of clarity of messages to the user 👍

PR #635, opened 2019-03-18
Replace alpine with debian-slim

### LINKED ISSUE #609: Update base images
Every iteration we evaluate an update to the base images.  Per requirements, we should update the base images to use to some form of debian-slim.  

Please sync with @rosskukulinski on any details or requirements.  

/assign @johnSchnake @stevesloka

johnSchnake:
Updates the images we use when building Sonobuoy.
For the build image this also updates go from
1.12.0 to 1.12.1.

Due to the fact that golang doesnt offer a
stretch-slim image, the 1.12.1 stretch image
is used.

Stretch slim seems to come with what it needs
so the dockerfile CMD1 can just be removed. For
ease of future changes I left the CMD1 in place and,
just like on arm64, we just remove the line rather
than replace it.

Fixes #609

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,9 @@ ifneq ($(VERBOSE),)
 VERBOSE_FLAG = -v
 endif
 BUILDMNT = /go/src/$(GOTARGET)
-BUILD_IMAGE ?= golang:1.12.0-alpine3.9
+BUILD_IMAGE ?= golang:1.12.1-stretch
+AMD_IMAGE ?= debian:stretch-slim
+ARM_IMAGE ?= arm64v8/ubuntu:16.04
 
 TESTARGS ?= $(VERBOSE_FLAG) -timeout 60s
 TEST_PKGS ?= $(GOTARGET)/cmd/... $(GOTARGET)/pkg/...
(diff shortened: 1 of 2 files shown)


johnSchnake:
@emuetzel for building against the arm architecture is the Ubuntu image still appropriate/OK?

johnSchnake:
Right now trying to diagnose the failure; it does seem to be a legitimate hang right now.

When you try and retrieve the results you get the results as expected, but then it hangs as if it doesnt realize the stream is done. Presumably, something is different between the two versions of `tar` used on the different images.

In stretch-slim it is GNU tar 1.29 and on the older alpine image it is `tar (busybox) 1.29`

In both cases when I ruun the tar command I get some messages on stderr but we just ignore those at this time. On both OSs, the tar command finishes. Seems like the stream doesnt get closed or some terminal character isnt found now.

johnSchnake:
Swapped back and forth between new/old images with a local client. The alpine image worked fine even with the new client so the problem is on the container side. Adding debug messages showed that the problem has to do with stdout not being closed ever; it is just keeps the stream open for some reason. More investigation required.

johnSchnake:
I'm fairly boggled right now. Kubectl exec works just as well as it did before (or seems to) with the new image. However there is still the problem that the stdout stream doesnt ever get closed for some reason so we hang at copying stdout, effectively waiting for the command to finish.

Once the command would finish, the stdout reader should have an eof/break that causes the copy to end. Since updating to debian that doesnt happen. Will keep working on it. Can I get confirmation that, assuming I can get this fixed, the images are appropriate now?

stevesloka:
I'm not 100% sure on the Ubuntu one (double checking with @timh). My other issue is, the arm e2e images don't work (or exist), so I'd actually recommend removing the arm builds until they are fixed.

johnSchnake:
@stevesloka Not sure we're on the same page on the arm one: it exists here https://hub.docker.com/r/arm64v8/ubuntu/ and I was able to do a `docker pull arm64v8/ubuntu:16.04` without problem. We have talked in the past and agree that we need to either test it in a meaningful way if we're going to build it though. I can make a ticket for that (either test or remove and let the discussion play out there)

johnSchnake:
*Solved*

So the issue is in the way that the different tar implementations work in the different platforms.

The reason the command hung had nothing to do with the connection being left open, but instead was the fact that in one build the tar library seemed to realize that it was seeing an end of archive marker (lots of 0s to indicate the end of the archive) and stopped reading. The fact that it stopped reading (in one platform and not the other) caused the `io.Pipe` to stop being able to write to the other side (it had more 0s to write) and so the `io.Copy` hung.

This is all remedied by adding some draining function to ensure the pipe is cleared. If any non-zero data is read we can throw an error/warning otherwise the 0s are harmless. Going to look upstream and see if this is a problem too (we got this code from upstream originally)

stevesloka:
@johnSchnake My point on arm is that if you do a `sonobuoy run` on an arm system, it will fail because upstream images don't exist. So we're essentially making a build which cannot run an E2E plugin. It would still work for other custom plugins, but not sure if anyone is actually doing that.

johnSchnake:
@stevesloka you're right, I misread your comment regarding the e2e images. Sorry about that. I'm with you otherwise about arm. I made #636 earlier today.

stevesloka:
No worries, I wasn't all that clear in my other comment. =)

vielmetti:
@johnSchnake - if you need GNU tar for this to work, you should install GNU tar everywhere - the `busybox` tar is no where near as complete.

johnSchnake:
It was actually the opposite problem; we went from alpine (which used the busybox tar) to the gnu tar. Something about the way the tar library is built and/or the end-of-archive markers meant that the tail of the archive (even though it was just 0s) was getting left in a reader which we expected to have been drained.

It didnt really have anything to do with tar itself, but I think much more likely something about the build of the stdlib tar reader leaving the 0s unconsumed on that platform.

PR #642, opened 2019-03-28
Push images in Jenkins

johnSchnake:
Our new build/release process goes through Jenkins
so the agent/auth requirements are a bit different.


stevesloka:
@johnSchnake just needs a quick rebase! =)

stevesloka:
Also, the only other change we could do is to do it all now in a container. But it's not a blocker to getting this out.

Issue #643, opened 2019-03-28
Gen uses different image for worker vs aggregator by default


johnSchnake:
**What steps did you take and what happened:**
sonobuoy gen

**What did you expect to happen:**
The WorkerImage in the config.json to match the image used for the aggregator. The former still defaults to latest (it was hardcoded) and the former was updated to refer to the build version of sonobuoy itself.

**Environment:**

- Sonobuoy version: master

PR #644, opened 2019-03-28
Use default image for worker

johnSchnake:
The worker's default image was using a hardcoded
value and not referencing the const that was
defined for that purpose leading to the
aggregator and worker using different images.

* Notes for reviewer
A seemingly minor problem, I worry about what some of the edge cases might be if we have different defaults there. It could lead to some awkward behavior if we have a few point releases.

I _know_ we need to also add more tests but this is a tiny bug with an obvious cause. We need to spike eventually on improving testing of gen input/output.

Fixes: #643

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -47,6 +47,8 @@ const (
 	MasterContainerName = "kube-sonobuoy"
 	// MasterResultsPath is the location in the main container of the master pod where results will be archived.
 	MasterResultsPath = "/tmp/sonobuoy"
+	// DefaultSonobuoyPullPolicy is the default pull policy used in the Sonobuoy config.
+	DefaultSonobuoyPullPolicy = "IfNotPresent"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -297,8 +299,8 @@ func New() *Config {
 	}
 
 	// TODO (timothysc) reference the other consts
-	cfg.WorkerImage = "gcr.io/heptio-images/sonobuoy:latest"
-	cfg.ImagePullPolicy = "Always"
+	cfg.WorkerImage = DefaultImage
+	cfg.ImagePullPolicy = DefaultSonobuoyPullPolicy
```
stevesloka: Since the const `DefaultImage` uses `gcr.io/heptio-images/sonobuoy:buildinfo.Version`, I think the pull policy should change to `IfNotPresent` so that if it's already there, it won't always attempt to pull (especially in air-gapped envs).

johnSchnake: Done.

PR #649, opened 2019-03-29
Skip goreleaser validation

johnSchnake:
Testing that we've added leaves a few artifacts around
which mess up the porcelain state goreleaser wants.

For the time being we're just going to remove the
validation.


johnSchnake:
Not going to wait on this to test since it is a one line change that wont even be hit since we aren't tagging this commit.

PR #654, opened 2019-04-02
Add additional documentation about testing in airgap

johnSchnake:
In addition to the test images the users need
the conformance and sonobuoy images. Created a section
showing this and also an example script for how to pull,
tag, and push the images.

### DIFF
--- a/docs/airgap.md
+++ b/docs/airgap.md
@@ -5,6 +5,8 @@ Sonobuoy supports running end-to-end tests with custom registries. This enables
 you to test your air-gapped deployment once you've loaded the necessary images
 into a registry that is reachable by your cluster.
 
+## Test images
+
 Just provide the `--e2e-repo-config` parameter and pass it the path to a local
 YAML file pointing to the registries you'd like to use. This will instruct the
 Kubernetes end-to-end suite to use your registries instead of the default ones.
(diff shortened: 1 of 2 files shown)


johnSchnake:
Added a section about systemd_logs since that plugin is a bit of an oddball as is. When we get the `--plugin` options we should make it easier to skip that from the CLI

docs/airgap.md
```diff
@@ -26,4 +28,31 @@ sampleRegistry: gcr.io/google-samples
...
+docker pull gcr.io/heptio-images/sonobuoy:$SONO_VERSION
+
+docker tag gcr.io/google-containers/conformance:$CLUSTER_VERSION $PRIVATE_REG/conformance:$CLUSTER_VERSION
+docker tag gcr.io/heptio-images/sonobuoy:$SONO_VERSION $PRIVATE_REG/sonobuoy:$SONO_VERSION
+
+docker push $PRIVATE_REG/conformance:$CLUSTER_VERSION
+docker push $PRIVATE_REG/sonobuoy:$SONO_VERSION
+```
+
+If you want to run the systemd_logs plugin you'll need to pull/tag/push it as well. In addition, you'll have to manually specify the image you want to use via `sonobuoy gen` -> `kubectl apply` since that image is not overridable on the CLI.
```
stevesloka: Maybe specify the value to replace here for the plugin so folks know what to look for? (e.g. `gcr.io/heptio-images/sonobuoy-plugin-systemd-logs:latest`)

Also might be nice to next add docs section on how to edit plugins (add / remove) and can link to that from here.

docs/airgap.md
```diff
@@ -26,4 +28,31 @@ sampleRegistry: gcr.io/google-samples
...
+
+If you do not wish to run it in your air-gapped cluster, just remove it from the list of plugins to be run (again, using `sonobuoy gen` -> `kubectl apply`).
```
stevesloka: might be nice to next add docs section on how to edit plugins (add / remove) and can link to that from here.

johnSchnake: I think that exists and can be linked to; I agree we should.

Also 🤞 that that will get simplified in the next release or two.

johnSchnake:
Well, I did find that our plugin documentation was indeed lacking. I added a small section about the plugin selection list and linked to `gen` for illustration about the general `gen -> kubectl` flow.

Also added the default image name for systemd_logs plugin.

Ready for another look.

PR #656, opened 2019-04-04
Fix CRD query

### LINKED ISSUE #640: CRDs not queried correctly
`sonobuoy run`, get the tarball and inspect. In the pod logs for the aggregator it is clear that the CRD query fails:
```
time="2019-03-14T20:02:28Z" level=error msg="error querying CustomResourceDefinitions: the server could not find the requested resource"
```
It is not due to lack of CRDs, but because the query is executed incorrectly. I created a test image to see what URL it was trying to hit and it is using:
```
https://10.96.0.1:443/api/v1/customresourcedefinitions
```

It should be using `/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions`

I found this when investigating the new k8s 1.14 release but it seems to have been a longstanding issue that was never reported. I consider this low priority and not a blocking bug.

Here is the section in question: https://github.com/heptio/sonobuoy/blame/master/pkg/discovery/queries.go#L181-L187

You can see that it uses the corev1 client which maps to that url instead of the one we want.

**Environment:**

- Sonobuoy version: master
- Kubernetes version: (use `kubectl version`): 1.14.0 (but it existed before)
- Kubernetes installer & version: kind 1.14.0 cluster

johnSchnake:
The CRD query did not have a properly formed
rest client leading to the wrong URLs being
used to query for CRDs. This resulted in an
error every Sonobuoy run on the aggregator.

This change fixes the issue in a backwards
compatible way by introducing a wrapper function
which properly handles the crdClient. The legacy
method is kept for 2 k8s release cycles just
in case someone is integrating against it. The
legacy method acts as if they had just passed
nil, or no crdClient, which will just err as
it did before.

Fixes #640

** Special note for reviewer **
No test added here since it is just a matter of integrating correctly with the k8s API.

To manually test this you can do the following (assuming you have a cluster up):

```
sonobuoy run --mode quick --wait
```

You can either just tail the logs or retrieve the results and look in the aggregator logs. You will see an error about failing to query CRDs.

If you do that again but with an image based on this fix:

```
sonobuoy run --mode quick --wait --sonobuoy-image=--sonobuoy-image=schnake/sonobuoy:v0.14.0-4-g7775d40-dirty
```

You will see the error goes away. If you add a CRD to your cluster you will also see it in the `resources/cluster/customresourcedefinitions` file. For an example CRD, just use `kubectl apply -f crd.yaml`

where `crd.yaml` is:
```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: crontabs.stable.example.com
spec:
  group: stable.example.com
  version: v1
  scope: Namespaced
  names:
    plural: crontabs
    singular: crontab
    kind: CronTab
    shortNames:
    - ct
```

**Release note**:
```
Fixed a bug which caused CRDs to be improperly queried.
```

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -30,22 +30,36 @@ import (
 	"github.com/rifflock/lfshook"
 	"github.com/sirupsen/logrus"
 	"github.com/viniciuschiele/tarx"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 
 	pluginaggregation "github.com/heptio/sonobuoy/pkg/plugin/aggregation"
 )
 
-// Run is the main entrypoint for discovery
-func Run(kubeClient kubernetes.Interface, cfg *config.Config) (errCount int) {
+// Run is the main entrypoint for discovery.
+func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 	t := time.Now()
+
+	kubeClient, err := kubernetes.NewForConfig(restConf)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "could not create kubernetes client"))
+		return errCount + 1
+	}
+
+	crdClient, err := apiextensionsclient.NewForConfig(restConf)
+	if err != nil {
+		errlog.LogError(errors.Wrap(err, "could not create apiextensions client"))
+		return errCount + 1
+	}
 
 	// 1. Create the directory which will store the results, including the
 	// `meta` directory inside it (which we always need regardless of
 	// config)
 	outpath := path.Join(cfg.ResultsDir, cfg.UUID)
 	metapath := path.Join(outpath, MetaLocation)
-	err := os.MkdirAll(metapath, 0755)
+	err = os.MkdirAll(metapath, 0755)
 	if err != nil {
 		errlog.LogError(errors.Wrap(err, "could not create directory to store results"))
 		return errCount + 1
@@ -102,7 +116,7 @@ func Run(kubeClient kubernetes.Interface, cfg *config.Config) (errCount int) {
 	// 5. Run the queries
 	recorder := NewQueryRecorder()
 	trackErrorsFor("querying cluster resources")(
-		QueryClusterResources(kubeClient, recorder, cfg),
+		QueryClusterResources(kubeClient, crdClient, recorder, cfg),
 	)
 
 	for _, ns := range nslist {
--- a/cmd/sonobuoy/app/master.go
+++ b/cmd/sonobuoy/app/master.go
@@ -25,8 +25,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-
-	"k8s.io/client-go/kubernetes"
 )
 
 var noExit bool
@@ -63,14 +61,8 @@ func runMaster(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	clientset, err := kubernetes.NewForConfig(kcfg)
-	if err != nil {
-		errlog.LogError(err)
-		os.Exit(1)
-	}
-
 	// Run Discovery (gather API data, run plugins)
-	errcount := discovery.Run(clientset, cfg)
+	errcount := discovery.Run(kcfg, cfg)
 
 	if noExit {
 		logrus.Info("no-exit was specified, sonobuoy is now blocking")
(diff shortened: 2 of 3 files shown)


pkg/discovery/discovery.go
```diff
@@ -30,14 +30,21 @@ import (
...
 
-// Run is the main entrypoint for discovery
+// Deprecated during 0.14.0; should be removed by 0.16.0. Previously the main entrypoint
+// for discovery but without the crdClient. Querying for CRDs is skipped.
 func Run(kubeClient kubernetes.Interface, cfg *config.Config) (errCount int) {
+	return RunConfig(kubeClient, nil, cfg)
+}
+
+// RunConfig is the main entrypoint for discovery.
+func RunConfig(kubeClient kubernetes.Interface, crdClient apiextensionsclient.Interface, cfg *config.Config) (errCount int) {
```
johnSchnake: Perhaps instead of passing the 2 clients we should be just passing the `*rest.Config`.

I have a POC locally which uses the dynamic client for all the queries and it needs the `*rest.Config` in order to make any clients it needs for any of the resources.

johnSchnake: It is at least a misnomer because we aren't passing a config here so we shouldn't call it `RunConfig`.

I think the reason I landed on this was that otherwise we can't easily support the older `Run` method because we can't get the `rest.Config` back from the `kubernetes.Interface`

And moving to the dynamic client, although my preference, isn't so perfect because it does slow down queries significantly as of now.

So my opinion is we just patch this bug like above (fixing the misnamed method) then if/when we move to the dynamic client we'll just duplicate the code and deprecate these methods (or remove them based on how you feel about deprecation)

pkg/discovery/discovery.go
```diff
@@ -30,14 +30,21 @@ import (
...
 	"github.com/viniciuschiele/tarx"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 
 	pluginaggregation "github.com/heptio/sonobuoy/pkg/plugin/aggregation"
 )
 
-// Run is the main entrypoint for discovery
+// Deprecated during 0.14.0; should be removed by 0.16.0. Previously the main entrypoint
```
johnSchnake: @stevesloka thoughts on deprecation vs removal?

These are exported methods so people could consume them in code, but if thats the case they can also pin the version they want/need.

These are not intrinsically part of the sonobuoy client interface and is really just consumed by the `sonobuoy master` command itself.

pkg/discovery/discovery.go
johnSchnake: Updated name already now `RunWithCRDs`; a little awkward but accurate and, again, isnt a commonly used method.

pkg/discovery/discovery.go
stevesloka: Why are you deprecating? If they aren't used anywhere else I'd just throw away and remove the confusion.

stevesloka:
Something doesn't seem right. I did a scan, but don't get the `customresourcedefinitions` file in resources --> cluster output. 

```
$ kubectl get crd                                                                        
NAME                                           CREATED AT
ingressroutes.contour.heptio.com               2019-04-05T14:04:38Z
tlscertificatedelegations.contour.heptio.com   2019-04-05T14:04:38Z
```

Screen Shot 2019-04-05 at 3 15 19 PM

johnSchnake:
@stevesloka you mentioned things aren't seeming right but did you follow all the repro steps? You actually have to push/reference your custom sonobuoy image so that the workers use the new code.

I just want to make sure you didnt just update your CLI before I try to repro.

johnSchnake:
Rereading my PR it may have been misleading; I can't attest that the `schnake/sonobuoy:v0.14.0` has this fix anymore; it was just one that I iterated with at the time of that writing. I can push one that should be known to be good and double check that works as expected so you dont have to push your own. Will post back here when I'm done.

stevesloka:
Let me rebuild my image and retest.

johnSchnake:
So I used our heptio kube-conformance image because on my first run to validate this I got an unusual but scary message in the logs:

```
/run_e2e.sh: line 65: wait: `': not a pid or valid job spec
```

so that means it, for whatever reason, hit the bug with the upstream conformance image and the container died leaving Sonobuoy hanging.

Confirmed that things worked for me again. My image is `schnake/sonobuoy:crdQuery` and my command was:
```
$ go build
$ ./sonobuoy run --mode=quick --sonobuoy-image schnake/sonobuoy:crdQuery --kube-conformance-image=gcr.io/heptio-images/kube-conformance:v0.14.0 --wait
$ rm -rf results
$ outfile=$(sonobuoy retrieve)
$ mkdir results && tar -xf $outfile -C results && code results
```

Issue #638, opened 2019-03-25
Frontload template usage


johnSchnake:
**Describe the solution you'd like**
I think that once #405 is completed it unblocks my preferred solution for this (see design doc below).

TL;DR; is that right now we use a template locally to make the aggregator (among other things) but only upload a small plugin definition for each of the plugins. Although that eases the writing of plugins, it added another step on the aggregator which executed other templates to generate the full YAML for each plugin.

As a user I would much rather all the templating happens locally so that I can make any modifications to the YAML after the fact.

>Note: I still want to leverage the simple plugin model that has been created and dont want the user to have to specify the full podspec. I just want the CLI to be the consumer of that simple plugin definition and the aggregator not generate the YAML server side.

**Anything else you would like to add:**
https://docs.google.com/document/d/1P7wpB3XR4-0u6I3rdsezw_E5oZ8uPhEFQqqfpm-F3tE/edit#

@stevesloka

johnSchnake:
So doing something like this makes effectively everything configurable. Is that _only_ a good thing or does it open security vulnerabilities too? Do we explicitly _not_ want to let people modify pod settings?

johnSchnake:
FWIW we dont use the templates for the plugins now, but instead use the actual k8s objects. 

In addition, @zubron just added the ability to modify other podSpec options in a general way so there are becoming less and less things that we _cant_ manipulate using the plugin specs and initial yaml produced by `sonobuoy gen`.

johnSchnake:
After reviewing the problem a bit more and looking at the design doc, I think that the work to expose the podspec options hits 95+%  of the use cases and we can even add an objectMetadata field if we need as well. In either case, each plugin can specify everything they want/need there and there is no reason to be doing more work clientside and complicating the output of sonobuoy gen.

PR #660, opened 2019-04-05
Support imagePullSecrets for sonobuoy workers

### LINKED ISSUE #424: Allow 'plugin' specifications to reference 'imagePullSecrets'.
**FEATURE REQUEST**: 

Allow 'plugin' specifications to reference 'imagePullSecrets' to allow test container to be pulled from secured repositories.

**What happened**: 

We were using the sonobuoy plugin mechanism to run our own test images alongside the E2E tests against our K8s clusters. The host registry had to be made private, so, now requires docker credentials in order to pull down the image.

We should be able to specify 'imagePullSecrets' in our test plugin. If the required secret is available in the cluster, then it should be able to use them to pull down the test container image.

Take any plugin test image and store it in a repository that requires credentials to pull it. The test pod will report an image pull error when trying to run the tests.

**Environment**:
v0.11.1
```
Server Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.5", GitCommit:"f01a2bf98249a4db383560443a59bed0c13575df", GitTreeState:"clean", BuildDate:"2018-03-19T15:50:45Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```
n/a
n/a
- Kernel (e.g. `uname -a`):
n/a

johnSchnake:
In the situation where the sonobuoy image itself was
in a private registry there was no recourse for users
because imagePullSecrets was not defined in the
config object used to help populate the templates
and these templates were generated on the server
side (so the user couldn't use sonobuoy gen and
manually manipulate them).

This change does the following:
 - adds imagePullSecrets to the config type
 - adds sections in the templates to honor it

It does not modify the original template to
add the secret or apply it to the aggregator
pod, those steps must be done manually after
sonobuoy gen.

This at least unblocks people who may be in
this situation.

Fixes #424

A little hard to test in a unit test besides testing the gen output but I've ensured this works by doing the following:

 - create a private repo on docker hub
 - save output of `sonobuoy gen --mode quick --sonobuoy-image <my-private-image>`

 - create a secret in my default namespace [using my docker hub credentials](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/)
 - use kubectl to grab that secret's yaml and trim out the metadata
 - paste that yaml into the output from sonobuoy gen so that we create the secret in the heptio-sonobuoy namespace
 - add `imagePullSecrets` field to the aggregator pod
 - change the `ImagePullPolicy` in the config.json to 'Always'

If you run it at that stage you should get the aggregator up (because you set its imagePullSecrets) but the e2e test pod fails to load because it can't pull the sonobuoy worker image.

 - delete sonobuoy so you can retry with the fix
 - now set the ImagePullPolicy field appropriately in the config.json (in the main yaml again)

You should now be able to run that yaml and have everything work.

It is not a particularly elegant fix, but we were a hard blocker before so this at least adds that support.

**Release note**:
```
ImagePullSecrets added as a customizable field in the SonobuoyConfig so that workers can pull Sonobuoy images from private registries if needed. Requires the user to manually create the secret in the appropriate namespace.
```


cmd/sonobuoy/app/gen_plugin.go
```diff
@@ -59,6 +60,10 @@ func NewCmdGenPlugin() *cobra.Command {
 		&genPluginOpts.Paths, "paths", "p", []string{".", "./examples/plugins.d/"},
 		"the paths to search for the plugins in. Defaults to . and ./plugins.d/",
 	)
+	cmd.Flags().StringVar(
```
johnSchnake: This is just on the hidden, dev command which executes the plugin templates. This normally happens on the aggregator server and is not locally easy to do. I think this command though should either (1) go away (2) become the actual path by which the aggr server generates the yaml (3) go away due to frontloading the yaml generation via #638

pkg/client/gen.go
```diff
@@ -66,6 +68,12 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
 		return nil, errors.Wrap(err, "couldn't marshall selector")
 	}
 
+	// Allow nil cfg.Config but avoid dereference errors.
+	conf := &config.Config{}
+	if cfg.Config != nil {
+		conf = cfg.Config
+	}
```
johnSchnake: We didn't ever use the nested values in the config; instead, there are redundant fields on the genConfig. I created a ticket (https://github.com/heptio/sonobuoy/issues/691) to solve that. Otherwise, the behavior is completely unclear.

johnSchnake:
Tests are green and I manually confirmed things work as expected as well.

Created https://github.com/heptio/sonobuoy/issues/692 to document this process. It is non-trivial, but again the point was to stop being a blocker.

PR #663, opened 2019-04-08
Makefile build

stevesloka:
When doing a build locally (e.g. `make container`) the makefile would error out if you didn't have the `docker-credential-gcr` binary locally. This moves that to the spot where it's needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,9 +28,6 @@ LINUX_ARCH := amd64 arm64
 DOCKERFILE :=
 PLATFORMS := $(subst $(SPACE),$(COMMA),$(foreach arch,$(LINUX_ARCH),linux/$(arch)))
 
-# SECRET is used when pushing the manifest.
-SECRET := $(shell echo "https://gcr.io" | docker-credential-gcr get | jq '.Secret')
-
 # Not used for pushing images, just for local building on other GOOS. Defaults to
 # grabbing from the local go env but can be set manually to avoid that requirement.
 HOST_GOOS ?= $(shell go env GOOS)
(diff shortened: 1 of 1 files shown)


johnSchnake:
I dont think this happens how you think it happens:
image

I'm trying to find the make references that clarify this, but it has to do with the order of operations within the makefile. This is what took me so long to figure out when fixing the Jenkins build because my make-fu is not strong.

Also, the makefile throws an error but I think its inaccurate to say that it `errs out`, it still finishes and makes the images necessary. It is a misleading message though so if we can get the message to go away (without messing up the build) then I'm all for it.

johnSchnake:
I think what we are looking for is `eval`, as in `$(eval SECRET=foo)`

https://www.gnu.org/software/make/manual/html_node/Reading-Makefiles.html partially clears it up, though I couldn't fully explain it. I think we can move it down if we use the `eval` statement correctly to ensure that the secret definition and substitution happen with the proper value.

johnSchnake:
Ha, so apparently they show how to do something similar to what we want here: http://make.mad-scientist.net/deferred-simple-variable-expansion/ with some wonky syntax.

It does also mention what I think would work for us too:
```
OUTPUT = $(shell some-command)
```
> This has the pleasant feature that some-command is not actually invoked until the variable reference $(OUTPUT) is expanded somewhere in the makefile. If $(OUTPUT) is never expanded, then some-command is never invoked. 

So then if the recipe we call doesnt use it, it won't be called and shouldnt error. Going to try it out...

johnSchnake:
Confirmed locally. Doing make clean with SECRET := ... throws the error (but continues). If you just remove the colon : then make clean doesnt throw the error. A non-trivial, one character change I think gets a 🏆 too

Makefile
```diff
@@ -148,6 +145,8 @@ push_images:
 	fi
 
 push_manifest:
+	# SECRET is used when pushing the manifest.
+	SECRET := $(shell echo "https://gcr.io" | docker-credential-gcr get | jq '.Secret')
```
johnSchnake: Comments left on the PR as a whole; changing the existing `:=` to `=` should fix our issue. (github made me add a comment here too...)

johnSchnake:
Superseded by #674

PR #664, opened 2019-04-08
Use heptio kube-conformance image for any cluster < v1.14.1

### LINKED ISSUE #655: Use heptio image for <= 1.14.0
ref https://github.com/kubernetes/kubernetes/pull/76101

Upstream conformance image hangs on failure because the container exits due to `errexit` being set.

Results should be saved and the sonobuoy worker should pick them up/report complete.

**Environment:**

- Sonobuoy version: 0.14.0
- Kubernetes version: (use `kubectl version`): 0.13.0 and above

stevesloka:
Any cluster earlier than v1.14.1 that has failing e2e tests would crash and not give feedback to the user. The heptio kube-conformance image is used by default for these clusters. Starting with any cluster v1.14.1 the upstream conformance image will be used by default. 

- Fixes #655 

**Release note**:
```
release-note: Any cluster earlier than v1.14.1 that has failing e2e tests would crash and not give feedback to the user. The heptio kube-conformance image is used by default for these clusters. Starting with any cluster v1.14.1 the upstream conformance image will be used by default. 
```

### DIFF
--- a/cmd/sonobuoy/app/gen.go
+++ b/cmd/sonobuoy/app/gen.go
@@ -139,19 +139,19 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
 	}, nil
 }
 
-// resolveConformanceImage maps versions before 1.13 to Heptio's image and otherwise
+// resolveConformanceImage maps versions before 1.14.0 to Heptio's image and otherwise
 // to the upstream cnoformance image. Latest is always mapped to the upstream
-// regardless. The comparison is just lexical, e.g. "foo" < "v1.13" and "zip" >
-// "v1.13". These are a-typical and not given more support at this time.
+// regardless. The comparison is just lexical, e.g. "foo" <= "v1.14.0" and "zip" >
+// "v1.14.0". These are a-typical and not given more support at this time.
 func resolveConformanceImage(imageVersion string) string {
 	// TODO(johnschnake): This logic should be temporary and is only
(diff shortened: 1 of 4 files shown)


johnSchnake:
Tests will need updated; currently ensures the change in image we use happens at 12.99 vs 13.0 but tests will need bumped to 1.14.0 and 1.14.1

stevesloka:
@johnSchnake tests updated.

johnSchnake:
Confirmed locally. Doing `make clean` with `SECRET := ...` throws the error (but continues). If you just remove the colon `:` then `make clean` doesnt throw the error. A non-trivial, one character change I think gets a  🏆 too

johnSchnake:
Going to assume this is failing because along with the window of which image to use, you need to shift:
```
func (c *ConformanceImageVersion) Get(client discovery.ServerVersionInterface) (string, error) {
```
It was sync'ed with that upstream window so that for upstream images we took `x.y.z` but for the heptio ones we just used `x.y`. That needs to use the `x.y` for all < 1.14.0 and less instead of "< 1.13" which is uses now.

johnSchnake:
Going to be glad when we can phase this out; those two separate things (image and version) were not meant to be so tightly bound in this wonky way.

stevesloka:
@johnSchnake I don't get why it was failing, but the tests are passing now.

PR #670, opened 2019-04-11
Add `gen plugin` command to generate boilerplate for plugins

### LINKED ISSUE #665: Plugin definition generation
Currently if you are developing local Sonobuoy plugins you have to write the plugin definition yourself. It is a simple document (a container spec along with some Sonobuoy data) but you either have to copy the e2e spec (via sonobuoy gen) and modify it or write it from scratch (confusing and no documentation).

I want to make that simple via a new sonobuoy command which generates the output for most cases.

Design doc here: https://docs.google.com/document/d/15RNWJRo8QLTp6bWBPrtos_Z32VD_zIvCWiVeoBPTs5E/edit?usp=sharing

Relates to #405

johnSchnake:
When creating a plugin there are two main steps:
 - generating the plugin manifest
 - putting that manifest into the `sonobuoy gen` YAML to run it

This addresses the first point. Using this command we can generate
the boilerplate YAML for most standard plugins. It allows
customizing the plugin type, name, image, command, and env vars.

Fixes #665

**Release note**:
```
Added `gen plugin` command to generate plugin manifest for custom plugins.
```

### DIFF
--- a/cmd/sonobuoy/app/gen_plugin_def.go
+++ b/cmd/sonobuoy/app/gen_plugin_def.go
@@ -1,0 +1,146 @@
+/*
+Copyright the Sonobuoy contributors 2019
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package app
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/heptio/sonobuoy/pkg/errlog"
+	"github.com/heptio/sonobuoy/pkg/plugin/manifest"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	v1 "k8s.io/api/core/v1"
+	kuberuntime "k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	defaultPluginName    = "plugin"
+	defaultPluginDriver  = "Job"
+	defaultMountPath     = "/tmp/results"
+	defaultMountName     = "results"
+	defaultMountReadOnly = false
+)
+
+// GenPluginDefConfig are the input options for running
+type GenPluginDefConfig struct {
+	def manifest.Manifest
+
+	// env holds the values that will be parsed into the manifest env vars.
+	env EnvVars
+
+	// driver holds the values that will be parsed into the plugin driver.
+	// Allows validation during flag parsing by having a custom type.
+	driver pluginDriver
+}
+
+// NewCmdGenPluginDef ...
+func NewCmdGenPluginDef() *cobra.Command {
+	genPluginOpts := GenPluginDefConfig{
+		def: defaultManifest(),
+		env: map[string]string{},
+	}
+
+	cmd := &cobra.Command{
+		Use:   "plugin",
+		Short: "Generates the manifest Sonobuoy uses to define a plugin",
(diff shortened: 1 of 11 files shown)


cmd/sonobuoy/app/plugindriver.go
```diff
@@ -0,0 +1,44 @@
...
+func (d *pluginDriver) Type() string   { return "pluginDriver" }
+
+func (d *pluginDriver) Set(str string) error {
+	lcase := strings.ToLower(str)
+	driver, ok := driverMap[lcase]
+	if !ok {
+		return fmt.Errorf("unknown plugin driver %q", str)
+	}
+	fmt.Println("schnake driver map", driverMap)
+	fmt.Println("schnake str", str, "driver", driver, ok)
```
johnSchnake: Remove debug lines...

pkg/plugin/manifest/manifest.go
```diff
@@ -44,7 +44,7 @@ func (s *SonobuoyConfig) DeepCopy() *SonobuoyConfig {
 type Manifest struct {
 	SonobuoyConfig SonobuoyConfig `json:"sonobuoy-config"`
 	Spec           Container      `json:"spec"`
-	ExtraVolumes   []Volume       `json:"extra-volumes"`
+	ExtraVolumes   []Volume       `json:"extra-volumes,omitempty"`
```
johnSchnake: This was because otherwise, we got a slightly ugly (and unnecessary) `extra-volumes: []` which we didn't normally see in our manually written YAML.

cmd/sonobuoy/app/gen_plugin_def.go
```diff
@@ -0,0 +1,144 @@
...
+			MountPath: "/tmp/results",
+			Name:      "results",
+			ReadOnly:  false,
+		},
+	}
+	return m
+}
+
+// toYAML serializes the manifest to YAML.
+func toYAML(m manifest.Manifest) ([]byte, error) {
```
johnSchnake: No point to having this method, its completely trivial (only found that out after creating it).

cmd/sonobuoy/app/gen_plugin_def.go
```diff
@@ -0,0 +1,144 @@
...
+	)
+
+	return cmd
+}
+
+// defaultManifest returns the basic manifest the user's options will be placed
+// on top of.
+func defaultManifest() manifest.Manifest {
+	m := manifest.Manifest{}
+	m.Spec.Name = "plugin"
```
johnSchnake: Use consts/vars for these magic strings/values.

johnSchnake:
Good to see the test stats. Looking at the lines covered/not it is not surprising: the CLI based code (setting up commands, handling input, etc) is not tested well while the actual logic is covered fine.

Some refactoring will need to be done to try and easily test more of the codebase (right now its too awkward to test all the CLI stuff because we have exits throughout the code (among a few other things that could be fixed up).

cmd/sonobuoy/app/gen_plugin_def.go
```diff
@@ -0,0 +1,146 @@
...
+			errlog.LogError(err)
+			os.Exit(1)
+		}
+		fmt.Println(s)
+	}
+}
+
+// genPluginDef returns the YAML for the plugin which Sonobuoy would expect as
+// a configMap in order to run/gen a typical run.
+func genPluginDef(cfg *GenPluginDefConfig) (string, error) {
```
chuckha: I kind of expected this to have a return type of `([]byte, error)`, but I can see why this works as `(string, error)`

johnSchnake:
Changes since approval:
 - changed signature to use ([]byte,error) which is more idiomatic (and fixed references)
 - changed capitalization error I found with plugin driver (`Daemonset` was used instead of `DaemonSet`)
 - because of how annoyed I was at the issue above, I changed the plugin loader to use `ToLower` before comparing the plugindriver. We do not intend for there to be a `Job` and a `job` meaning different things. I could easily see someone *cough* me *cough* making that mistake again

PR #673, opened 2019-04-13
Change logic in template to use a list of plugin manifests

### LINKED ISSUE #672: Template should use objects for plugin yaml
Use objects which serialize themselves to YAML instead of just go templates when writing the config map for all the plugin manifests.

The reason for doing this is in support of the `--plugin`; e.g. loading custom plugins and putting them into the template (see #405 )

The logic is this: we want to, at final template generation time, be able to just iterate through the plugins and serialize them all as needed. We'd don't want to always list the e2e plugin if we aren't running it (same for systemd_logs). There are some extra resources that e2e may require like ssh keys/ repo lists, and if we can make those objects as well (and maybe make them more generic) then we could further extend our capabilities later.

As of now though, we want to get away from just the templated plugin definition part of the template.

We still need all that data to be considered when customizing the plugin so we can't just hardcode the e2e plugin. A reasonable approach seems to be to add a slice of manifests to the template data and then, prior to filling the template, generating the e2e/systemd_logs plugin objects as needed.

johnSchnake:
Currently, the template for the initial YAML (via sonobuoy gen)
has it hardcoded that both the e2e and systemd-logs plugin are
configured. This is 1) misleading to users who might see this and
think both are going to be run and 2) brittle.

With upcoming changes wanting to make the plugin list more
variable, it makes sense to put this change first: go ahead and
track a plugin list and adjust the template to iterate over the
plugins, inserting their YAML each time.

This makes the YAML more clear, flexible, and in some cases shorter.

Fixes #672

**Release note**:
```
`sonobuoy gen` output will now reflect only the set of plugins you are selecting to run instead of always generating the e2e and systemd-logs configurations.
```

### DIFF
--- a/pkg/client/gen.go
+++ b/pkg/client/gen.go
@@ -26,7 +26,18 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/heptio/sonobuoy/pkg/buildinfo"
+	"github.com/heptio/sonobuoy/pkg/plugin"
+	"github.com/heptio/sonobuoy/pkg/plugin/manifest"
 	"github.com/heptio/sonobuoy/pkg/templates"
+
+	corev1 "k8s.io/api/core/v1"
+	kuberuntime "k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	e2ePluginName    = "e2e"
+	systemdLogsName  = "systemd-logs"
+	pluginResultsDir = "/tmp/results"
 )
 
 // templateValues are used for direct template substitution for manifest generation.
@@ -34,6 +45,8 @@ type templateValues struct {
 	E2EFocus    string
 	E2ESkip     string
 	E2EParallel string
+
+	Plugins []string
 
 	SonobuoyConfig       string
 	SonobuoyImage        string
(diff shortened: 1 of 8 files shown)


pkg/client/gen.go
```diff
@@ -67,6 +80,23 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
 		}
 	}
 
+	plugins := []*manifest.Manifest{}
+	if includes(cfg.Config.PluginSelections, e2ePluginName) {
+		plugins = append(plugins, e2eManifest(cfg))
+	}
+	if includes(cfg.Config.PluginSelections, systemdLogsName) {
```
johnSchnake: The 2x check on `includes` is obviously inefficient but:
1) the list of plugins is always short
2) this ensures a standard order: e2e, systemdlogs, other plugins (in followup PR)

pkg/client/gen.go
```diff
@@ -67,6 +80,23 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
...
+		plugins = append(plugins, systemdLogsManifest(cfg))
+	}
+
+	pluginYAML := []string{}
+	for _, v := range plugins {
+		yaml, err := kuberuntime.Encode(manifest.Encoder, v)
+		if err != nil {
+			return nil, errors.Wrapf(err, "serializing plugin %v as YAML", v.SonobuoyConfig.PluginName)
+		}
+		pluginYAML = append(pluginYAML, strings.TrimSpace(string(yaml)))
```
johnSchnake: I was getting an EOL character or something that led to an extra, empty line in the template when rendered. Trimming it seems reasonable and fixed the issue.

pkg/client/testdata/ssh.golden
```diff
@@ -35,72 +35,38 @@ metadata:
...
-      imagePullPolicy: 
       name: e2e
+      resources: {}
       volumeMounts:
       - mountPath: /tmp/results
         name: results
-        readOnly: false
       - mountPath: /root/.ssh
         name: sshkey-vol
-      tolerations:
```
johnSchnake: The tolerations is one thing worth noting: There is no field on the object `tolerations` so this just gets dropped. The YAML we now generate is accurate: serialized by the standard library and types are from the k8s vendored code.

pkg/client/gen.go
chuckha: Why do these two have to come first and in that order?

johnSchnake: No hard reason it is exactly that order; it does make it easier for testing if we know the order, regardless of what it is. I wouldn't be against just alphabetizing them by plugin name though (which actually results in this same order for the default coincidentally)

chuckha: was just thinking it'd be great if the tests didn't rely on order and the e2es and systemdlog plugins were not considered special or blessed plugins

johnSchnake: Yea, that has been a matter of discussion lately and with some of the upcoming plugin changes that gets closer to achievable. The idea of it being alphabetically sorted would satisfy the testing need and also be compatible with that goal.

Issue #651, opened 2019-04-01
Improve test coverage


johnSchnake:
We've known that we needed to increase test coverage for a while, I think that time has come to spike a bit on it.

github.com/heptio/sonobuoy/cmd/sonobuoy/app | 15.00%
github.com/heptio/sonobuoy/pkg/backplane/ca | 91.20%
github.com/heptio/sonobuoy/pkg/backplane/ca/authtest |  
github.com/heptio/sonobuoy/pkg/buildinfo |  
github.com/heptio/sonobuoy/pkg/client | 10.30%
github.com/heptio/sonobuoy/pkg/client/results | 72.20%
github.com/heptio/sonobuoy/pkg/client/results/e2e |  
github.com/heptio/sonobuoy/pkg/config | 64.90%
github.com/heptio/sonobuoy/pkg/discovery |  
github.com/heptio/sonobuoy/pkg/dynamic | 46.90%
github.com/heptio/sonobuoy/pkg/errlog |  
github.com/heptio/sonobuoy/pkg/image | 31.20%
github.com/heptio/sonobuoy/pkg/image/docker |  
github.com/heptio/sonobuoy/pkg/image/exec |  
github.com/heptio/sonobuoy/pkg/plugin |  
github.com/heptio/sonobuoy/pkg/plugin/aggregation | 43.60%
github.com/heptio/sonobuoy/pkg/plugin/driver | 40.00%
github.com/heptio/sonobuoy/pkg/plugin/driver/daemonset | 9.90%
github.com/heptio/sonobuoy/pkg/plugin/driver/job | 14.00%
github.com/heptio/sonobuoy/pkg/plugin/driver/utils |  
github.com/heptio/sonobuoy/pkg/plugin/loader | 84.50%
github.com/heptio/sonobuoy/pkg/plugin/manifest | 26.80%
github.com/heptio/sonobuoy/pkg/tarball | 72.40%
github.com/heptio/sonobuoy/pkg/templates |  
github.com/heptio/sonobuoy/pkg/worker | 60.30%
github.com/heptio/sonobuoy/test/integration |  

We have an average coverage of 45% with the actually CLI stuff (sonobuoy/app) having only 15%.

Need to set a goal for this particular spike but I wanted it considered when doing planning.

johnSchnake:
The first part of improving test coverage is to track it. Upstream k8s uses codecov so I'm looking at integrating it.

johnSchnake:
[...]

PR #675, opened 2019-04-14
Centralize the process of going from kubeconfig to sonobuoy client

johnSchnake:
This logic is repeated throughout nearly every command; this
tries to further centralize it.

The goal is to DRY up the codebase slightly making way for
further refactors to increase testability and test coverage
of the entire project.

Related to #651

Small change but I just wanted to keep it very focused for an easy review. More refactor/testability changes will be coming.

### DIFF
--- a/cmd/sonobuoy/app/common.go
+++ b/cmd/sonobuoy/app/common.go
@@ -19,3 +19,11 @@ func getSonobuoyClient(cfg *rest.Config) (*client.SonobuoyClient, error) {
 	}
 	return client.NewSonobuoyClient(cfg, skc)
 }
+
+func getSonobuoyClientFromKubecfg(kubecfg Kubeconfig) (*client.SonobuoyClient, error) {
+	cfg, err := kubecfg.Get()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get rest config")
+	}
+	return getSonobuoyClient(cfg)
+}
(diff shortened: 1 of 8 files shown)


cmd/sonobuoy/app/common.go
```diff
@@ -19,3 +19,11 @@ func getSonobuoyClient(cfg *rest.Config) (*client.SonobuoyClient, error) {
 	}
 	return client.NewSonobuoyClient(cfg, skc)
 }
+
+func getSonobuoyClientFromKubecfg(kubecfg Kubeconfig) (*client.SonobuoyClient, error) {
```
chuckha: I could see there also being a mustGetSonobuoyClientFromKubecfg function that panics/exits on an error and doesn't return an error since this function generally is used with os.Exit(1).

johnSchnake: I had done that originally but it was counter to my eventual goal. Having looked at the codebase (and having done something like this before) I'd like to switch to `RunE` for the cobra commands and centralize all the error handling there.

That way as much as possible I can centralize the error reporting, os.Exit calls, and make each of the functions as testable as possible. Having the os.Exit calls throughout can make it difficult to test certain paths.

johnSchnake:
gofmt failed to run against a file so I just fixed it and force pushed. FYI

PR #678, opened 2019-04-15
Tighten preflight check to err on existing namespace

### LINKED ISSUE #677: Prevent Sonobuoy from being deployed in an existing namespace (or at least kube-system)
Prevent using `sonobuoy run -n <namespace>` to deploy to an existing namespace (kube-system in particular). Doing so seems unnecessary (unless someone can think of a use case where it would be needed) and makes cleaning up Sonobuoy resources a pain. 

At the very least, maybe prevent it from ever being deployed in kube-system and print a warning with a y/n prompt if attempting to deploy into any other existing namespace.

**Environment:**

- Sonobuoy version: 0.14.1
- Kubernetes version: (use `kubectl version`): 1.14

johnSchnake:
The typical flow for Sonobuoy involves creating a new namespace
and it is uncommon to need to put it into an existing namespace.
When that occurs, cleanup must be manual since Sonobuoy by default
just deletes the namespace.

This tightens the precheck to error on the existing namespace. In
situations where that is still desired, preflight checks can be
skipped.

Refactored the logic some so that tests could be added for preflight
checks.

Fixes #677

**Release note**:
```
Preflight checks will not fail if the namespace Sonobuoy is targeting already exists. This helps avoid a situation where a user must manually cleanup a Sonobuoy run in order to avoid destroying the namespace.
```

### DIFF
--- a/pkg/client/preflight.go
+++ b/pkg/client/preflight.go
@@ -22,16 +22,38 @@ import (
 	version "github.com/hashicorp/go-version"
 	"github.com/heptio/sonobuoy/pkg/buildinfo"
 	"github.com/pkg/errors"
+	apicorev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 )
 
-var preflightChecks = []func(kubernetes.Interface, *PreflightConfig) error{
-	preflightDNSCheck,
-	preflightVersionCheck,
-	preflightExistingSonobuoy,
-}
+const (
+	kubeSystemNamespace = "kube-system"
+	kubeDNSLabelKey     = "k8s-app"
+	kubeDNSLabelValue   = "kube-dns"
+	coreDNSLabelValue   = "coredns"
+)
+
+var (
+	minimumKubeVersion = version.Must(version.NewVersion(buildinfo.MinimumKubeVersion))
+	maximumKubeVersion = version.Must(version.NewVersion(buildinfo.MaximumKubeVersion))
+
+	expectedDNSLabels = []string{
+		kubeDNSLabelValue,
+		coreDNSLabelValue,
+	}
+
+	preflightChecks = []func(kubernetes.Interface, *PreflightConfig) error{
+		preflightDNSCheck,
+		preflightVersionCheck,
+		preflightExistingNamespace,
+	}
+)
+
+type listFunc func(metav1.ListOptions) (*apicorev1.PodList, error)
+type nsGetFunc func(string, metav1.GetOptions) (*apicorev1.Namespace, error)
 
 // PreflightChecks runs all preflight checks in order, returning the first error encountered.
 func (c *SonobuoyClient) PreflightChecks(cfg *PreflightConfig) []error {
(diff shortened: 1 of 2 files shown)


pkg/client/preflight.go
```diff
@@ -41,40 +63,42 @@ func (c *SonobuoyClient) PreflightChecks(cfg *PreflightConfig) []error {
...
-	kubeDNSLabelValue   = "kube-dns"
-	coreDNSLabelValue   = "coredns"
-)
-
 func preflightDNSCheck(client kubernetes.Interface, cfg *PreflightConfig) error {
-	var dnsLabels = []string{
-		kubeDNSLabelValue,
-		coreDNSLabelValue,
+	return dnsCheck(
+		client.CoreV1().Pods(kubeSystemNamespace).List,
```
stevesloka: I think this can be simplified to just run `dnsCheck`.  Passing the dnsLabels & pod list command doesn't add much and just complicates.

pkg/client/preflight.go
```diff
@@ -84,44 +108,51 @@ func preflightDNSCheck(client kubernetes.Interface, cfg *PreflightConfig) error
...
+		return fmt.Errorf("maximum kubernetes version is %s, got %s", max.String(), versionInfo.String())
 	}
 
 	return nil
 }
 
-func preflightExistingSonobuoy(client kubernetes.Interface, cfg *PreflightConfig) error {
-	_, err := client.CoreV1().Pods(cfg.Namespace).Get("sonobuoy", metav1.GetOptions{})
+func preflightExistingNamespace(client kubernetes.Interface, cfg *PreflightConfig) error {
+	return nsCheck(
```
stevesloka: Same comments apply here as dnsCheck

stevesloka:
> The typical flow for Sonobuoy involves creating a new namespace
and it is uncommon to need to put it into an existing namespace.

Any reason to not support this? Why must the user use `heptio-sonobuoy` as the namespace? Or run multiple versions on sonobuoy in the same cluster? (i.e. have one to do e2e tests, another for customer plugin, etc)?

pkg/client/preflight.go
johnSchnake: It allows us (for the first time) to run tests against those functions. 100% of our logic requires a single method (e.g. list pods or whichever method you're talking about) but our method signatures require the entire kubernetes.Interface which makes them effectively untestable.

Then the dnsLabels stuff _could_ be hardcoded into the method but it's just a preference that things like that be injected as well. It also makes it more clear when looking at the method signature what you need to do/what it may be looking for; otherwise, you need to look up which constants actually get used in the method. Again, I think of this part of the code a little more habit/preference though.

johnSchnake:
> Any reason to not support this? 

It is still supported, but you just have to provide the `--skip-preflight` flag, implying that you know there may be consequences. The problem currently is that some users have put Sonobuoy in an existing namespace without realizing that removing the Sonobuoy resources must be a manual process in that case. If they _dont_ realize that and still try `sonobuoy delete -n my-existing-namespace` then we could end up causing damage to their cluster. I'm not against a more _careful_ deletion if lots of people are requesting it later, but as of now Sonobuoy gains nothing from being in an existing namespace, it isn't the norm, and it doesn't work well with the deletion model that has been in place for a long time.

> Why must the user use heptio-sonobuoy as the namespace? 

They don't have to use heptio-sonobuoy as the namespace, there is a `--namespace` flag which allows them to define which to use. The change to the preflight checks is only that we want the namespace to be a new, fresh namespace.

> Or run multiple versions on sonobuoy in the same cluster? (i.e. have one to do e2e tests, another for customer plugin, etc)?

Using the namespace flag you can run multiple copies of sonobuoy in your cluster; I actually do this all the time. In particular when a local k8s namespace is failing to delete I just use a different namespace rather than clean it up for my own velocity. In this situation (running multiple Sonobuoy copies) the current preflight checks already fail if the Sonobuoy pod exists in the chosen namespace so you already can't run multiple Sonobuoy copies in the same namespace. In addition, even if you skip preflight checks, the resource creation will fail because of the name conflict (pods/configmaps with same names in same namespaces)

PR #679, opened 2019-04-16
Fix flaky unit test

johnSchnake:
The ordering of YAML lists is not well defined and so this passed
tests locally and in CI by chance. Even trying to repeat the
flakiness failed at first because I was hitting the testing cache.

Removing the 2nd end var does little to limit the value of the test
and we also have functional tests (e2e integration tests) which
will hit this flow.


johnSchnake:
Fixed a test that I saw fail on master (which passed on branch). Merging without review to get green and move forward with other merges/dev.

PR #680, opened 2019-04-16
Add sonobuoy gen config file and documentation

### LINKED ISSUE #661: Sonobuoy default configuration command
A new command `sonobuoy config` which will simply print out a default sonobuoy configuration file. It is not clear to the user without looking in the code what all the possible values there are and this would help advertise that and make it easy for users to make simple changes:

```
sonbobuoy config > mySonobuoyConfig.json
```

Then they can edit as they need.

Very little overhead and I think it helps if we want users to actually use config files (which is preferable to adding flags for everything)

johnSchnake:
The Sonobuoy config file is a little mysterious to most
users since the values are only really defined in code
and not well documented.

This change adds a command to output the default config
so that users might tweak the values they need more
easily.

In addition, I updated the documentation to make it more
clear for everyone what the configuration file is and
how each of the fields are used.

Fixes #661

**Release note**:
```
Added "sonobuoy gen config" command to see the default config that is passed to the server and added documentation about each field. This helps any user utilize the customizability that already exists.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ sonobuoy logs
  * [plugins][plugins]
  * Testing of [air gapped clusters][airgap].
  * [Customization][gen] of YAML prior to running.
+ * The [Sonobuoy config file][sonobuoyconfig] file and how to edit it.
 
 ## Troubleshooting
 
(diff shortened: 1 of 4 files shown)


johnSchnake:
Had a few debug items I just removed from the build. I originally built it based on the `config.New()` function but found that the CLI adds some extra logic (mainly defaulting the plugin selection).

Build/test should pass now. No extra test added as the field ordering and UUIDs will make testing against a goldenfile/configuration difficult. It is just printing out an object so manual testing at this time is OK IMO.

PR #681, opened 2019-04-16
Add test that will run through command creation

johnSchnake:
The CLI code has very low code coverage which indicates
that it has lots of areas potentially containing bugs.

However, a large chunk of the code in the CLI is just
boilerplate that sets up cobra commands and doesn't
execute much actual logic.

The point of this test is to have a code flow go through
those so that code coverage better reflects actual,
substantive code.

Relates to #651

**Release note**:
```
N/A
```

### DIFF
--- a/cmd/sonobuoy/app/root_test.go
+++ b/cmd/sonobuoy/app/root_test.go
@@ -1,0 +1,29 @@
+/*
+Copyright the Sonobuoy contributors 2019
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+     http://www.apache.org/licenses/LICENSE-2.0
+
+	 Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


johnSchnake:
Merging this without review. For documentation's sake, I thought I'd argue with myself though.

A potential argument against it is that if the test itself doesn't hold much value then we shouldn't merge. However, as the commit message states, if we leave large chunks of code like this "untested" then the code coverage # means less to anyone looking at it and gives a false sense of how well tested the codebase is.

Followup: then why don't we add a more meaningful test?
Answer: in some capacity, the question is well founded; obviously we want good tests. But if the code is just setting up the command structure, then the only "unit test" of that would be to test the exact structure we're coding. It would just be repeating ourselves for the sake of testing, wouldn't add much value, and would make changes to the codebase more difficult because you'd have to update the test every time.

Eventually, the goal would be to have real tests use this to test the actual code flows. A few other structural things have to happen to make that a reality though including better error patterns and dependency injection.

PR #683, opened 2019-04-19
Plugin flag

### LINKED ISSUE #405: Run sonobuoy cli with custom plugins from local filesystem
Feature request

**What happened**:
I'd like to `sonobuoy run` with custom plugins. Right now it's not possible. I need to:
1. run `sonobuoy gen` > file.yaml
2. modify configmap with plugins definitions
3. kubectl apply -f file.yaml

I'd like to us only sonobuoy cli without editing yaml files.
F.e.:
1. I have custom config.json with list of the plugins
2. I have folder with my plugin definitions, each in a separate yaml file
3. I run:
`sonobuoy run --config config.json --pluginsFolder <path_to_folder>`
4. sonobuoy would get files from folder, parse and validate it, and modify configmap 'in fly' before hitting apiserver

**Environment**:
- Kernel (e.g. `uname -a`):

johnSchnake:
The user can now load custom plugin manifests into the
YAML that Sonobuoy will run with the --plugin flag.

For backwards compatibility, if no plugins are specified
in this manner and instead a PluginSelection object is
still specified, we will fallback to adding the necessary
plugins.

When loading plugins using the flag, they should specify
either a filename which points to a plugin manifest or
one of the sentinal values representing our internal plugins:
e2e or systemd-logs.

Fixes #405

Previously listed docs as a todo item on this but I think the docs can be added/reviewed separately.

**Release note**:
```
Added a --plugin flag to gen/run commands which allows you to run custom plugins from your local filesystem without editing the YAML.
```


cmd/sonobuoy/app/gen.go
```diff
@@ -128,14 +133,16 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
...
+		Config:               g.resolveConfig(),
 		Image:                g.sonobuoyImage,
 		Namespace:            g.namespace,
 		EnableRBAC:           rbacEnabled,
 		ImagePullPolicy:      g.imagePullPolicy.String(),
 		KubeConformanceImage: image,
 		SSHKeyPath:           g.sshKeyPath,
 		SSHUser:              g.sshUser,
+		DynamicPlugins:       g.plugins.dynamicPlugins,
+		StaticPlugins:        g.plugins.staticPlugins,
```
johnSchnake: These were put into the `GenConfig` and not the `Config` very consciously.  In the `Config` we used to give the list of plugins to run, but not the actual config. The static plugins in particular, since they are full objects, make the config.json much larger _and_ it is completely unnecessary as the aggregator doesnt need the full manifest in the json; it loads it from the configmap anyways.

cmd/sonobuoy/app/gen.go
```diff
@@ -128,14 +133,16 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
 
 	return &client.GenConfig{
 		E2EConfig:            e2ecfg,
-		Config:               g.getConfig(),
+		Config:               g.resolveConfig(),
```
johnSchnake: Updated the name to try and be a bit more explicit that this mutates things. It was really awkward when I tried to break the function up. The default config, loaded config, e2e flags, mode, and plugin flags are all just intertwined. It would be nice to be more clear in the code, but I'll definitely settle for clarity via good testing.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -205,6 +205,17 @@ func TestGetConfig(t *testing.T) {
...
 			},
+		}, {
+			name:     "Manually specified plugins should result in empty selection",
+			input:    &genFlags{},
+			cliInput: "--plugin e2e",
+			expected: &config.Config{
+				Namespace:        "heptio-sonobuoy",
+				WorkerImage:      "gcr.io/heptio-images/sonobuoy:" + buildinfo.Version,
+				ImagePullPolicy:  "IfNotPresent",
+				PluginSelections: nil,
```
johnSchnake: Empty selection now means "run every plugin in the search path". Did not solve the "run no plugins, only queries" issue.

I could:
 1) not worry about it as that is a _very_ small group if anyone does it and they can always run `--mode quick` without much of a meaningful slowdown.
 2) Maybe the nil slice indicates run all, the empty slice means run nothing? I hate to split hairs there but that is kind of the point of the nil value.

pkg/client/gen.go
```diff
@@ -42,21 +41,15 @@ const (
 
 // templateValues are used for direct template substitution for manifest generation.
 type templateValues struct {
-	E2EFocus    string
-	E2ESkip     string
-	E2EParallel string
-
```
johnSchnake: A previous PR made these values unnecessary in the template data. These values are consumed by a method which creates the e2e plugin object.

pkg/client/gen.go
```diff
@@ -81,43 +74,32 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
...
-	}
-	if includes(cfg.Config.PluginSelections, systemdLogsName) {
-		plugins = append(plugins, systemdLogsManifest(cfg))
+	for _, v := range cfg.DynamicPlugins {
+		switch v {
+		case e2ePluginName:
+			plugins = append(plugins, e2eManifest(cfg))
+		case systemdLogsName:
+			plugins = append(plugins, systemdLogsManifest(cfg))
+		}
```
johnSchnake: This iterates the loop once but the values could end up in either order based on how they were in the initial slice.

Good that it is not preferential there, but maybe bad that we still just put them in listed order and list the dynamic ones first.

I think this needs a sort by name at the end. It would be well defined and also independent of where the plugin came from.

pkg/client/gen.go
```diff
@@ -81,43 +74,32 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
...
 	}
 
-	// Template values that are regexps (`E2EFocus` and `E2ESkip`) are
-	// embedded in YAML files using single quotes to remove the need to
-	// escape characters e.g. `\` as they would be if using double quotes.
-	// As these strings are regexps, it is expected that they will contain,
-	// among other characters, backslashes. Only single quotes need to be
-	// escaped in single quote YAML strings, hence the substitions below.
-	// See http://www.yaml.org/spec/1.2/spec.html#id2788097 for more details
-	// on YAML escaping.
```
johnSchnake: All this is now handled by the yaml library which serializes it.  Should probably try to validate cases like this though; I'm not 100% sure there was sufficient testing here. The proof isnt that _our_ tests pass, but that the e2e tests would select properly. We may have been too opinionated in certain cases and the yaml serializer knows best.

pkg/client/gen_test.go
```diff
@@ -156,34 +152,71 @@ func TestGenerateManifestSSH(t *testing.T) {
...
 			name: "Only systemd_logs",
 			inputcm: &client.GenConfig{
-				E2EConfig: &client.E2EConfig{},
-				Config: fromConfig(func(c *config.Config) *config.Config {
-					c.PluginSelections = []plugin.Selection{plugin.Selection{Name: "systemd-logs"}}
-					return c
-				}),
+				E2EConfig:      &client.E2EConfig{},
+				Config:         newConfigWithoutUUID(),
+				DynamicPlugins: []string{"systemd-logs"},
```
johnSchnake: I updated these tests but I need to think through the implications a bit more. What if someone does just send a plugin selection and no other plugins?

Do we want to treat it like they are saying to run nothing? Error?

Again, the aggregator should be getting the plugin info from the mounted configmaps not from the config itself. But I just need to consider what these tests were really about and how consumers may be affected.

pkg/client/testdata/default.golden
```diff
@@ -16,7 +16,7 @@ metadata:
...
+    {"Description":"DEFAULT","UUID":"","Version":"v0.14.1","ResultsDir":"/tmp/sonobuoy","Resources":["CertificateSigningRequests","ClusterRoleBindings","ClusterRoles","ComponentStatuses","CustomResourceDefinitions","Nodes","PersistentVolumes","PodSecurityPolicies","ServerGroups","ServerVersion","StorageClasses","ConfigMaps","ControllerRevisions","CronJobs","DaemonSets","Deployments","Endpoints","Ingresses","Jobs","LimitRanges","NetworkPolicies","PersistentVolumeClaims","PodDisruptionBudgets","PodTemplates","Pods","ReplicaSets","ReplicationControllers","ResourceQuotas","RoleBindings","Roles","ServiceAccounts","Services","StatefulSets"],"Filters":{"Namespaces":".*","LabelSelector":""},"Limits":{"PodLogs":{"LimitSize":"","LimitTime":""}},"Server":{"bindaddress":"0.0.0.0","bindport":8080,"advertiseaddress":"","timeoutseconds":10800},"PluginSearchPath":["./plugins.d","/etc/sonobuoy/plugins.d","~/sonobuoy/plugins.d"],"Namespace":"heptio-sonobuoy","WorkerImage":"gcr.io/heptio-images/sonobuoy:v0.14.1","ImagePullPolicy":"IfNotPresent"}
```
johnSchnake: Lots of these config.json files were changed because I made certain null values `omitempty`; that clouds up this PR slightly and I should separate those out.

pkg/plugin/loader/loader.go
```diff
@@ -73,7 +75,12 @@ func LoadAllPlugins(namespace, sonobuoyImage, imagePullPolicy string, searchPath
 		pluginDefinitions = append(pluginDefinitions, pluginDefinition)
 	}
 
-	pluginDefinitions = filterPluginDef(pluginDefinitions, selections)
+	// Run every plugin that is loaded unless specifically told otherwise.
+	// This is kept only for some amount of backwards compatibility. We should now only
+	// be loading the plugins we want and may be able to remove this logic soon.
+	if len(selections) > 0 {
```
johnSchnake: Functional change; again I should consider whether or not I want to differentiate between nil and empty here. That way I can support nil -> run all implicitly vs empty -> I really dont want to run any plugins.

pkg/client/gen_test.go
johnSchnake: I think especially in this context, updating the tests are fine. The tests are about whether or not given an appropriate configuration whether or not we generate the yaml we expect.

I do, however, need to define and test the behavior if someone wants to ignore these changes. If they just send the `Plugin` without setting either of the new fields. Also need to consider interaction with #684. TBD

johnSchnake:
Will be ready for review once the 2 PRs I split from this (#684 and #685 merge) and this gets rebased.

cmd/sonobuoy/app/gen_test.go
johnSchnake: So this [PR][https://github.com/heptio/sonobuoy/pull/684] made nil -> run all and empty slice means run nothing.

To order to actually execute that path you'd have to edit the YAML manually. I'm fine with this since it we'd be making the common cases easier without completely preventing the rare use case.

pkg/client/gen.go
```diff
@@ -80,13 +74,28 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
 		}
 	}
 
-	plugins := []*manifest.Manifest{}
-	if includes(cfg.Config.PluginSelections, e2ePluginName) {
-		plugins = append(plugins, e2eManifest(cfg))
+	// Support legacy logic for the time being.
+	if len(cfg.DynamicPlugins) == 0 && len(cfg.StaticPlugins) == 0 &&
+		len(cfg.Config.PluginSelections) > 0 {
```
johnSchnake: Double check this logic and add a test case; if we want to support backwards compat then we need to support the empty array -> run no plugins case. Rare, but that is the intent and we need to make sure it works as expected.

cmd/sonobuoy/app/args.go
```diff
@@ -281,6 +281,12 @@ func AddSSHUserFlag(user *string, flags *pflag.FlagSet) {
 	)
 }
 
+// AddPluginSetFlag adds the flag for gen/run which keeps track of which plugins
+// to run and loads them from local files if necessary.
+func AddPluginSetFlag(p *pluginList, flags *pflag.FlagSet) {
+	flags.VarP(p, "plugin", "p", "Which plugins to run. Can either point to a local file or be one of the known plugins (e2e or systemd-logs). Can be specified multiple times to run multiple plugins.")
```
johnSchnake: Should test that names should be unique. This error case was called out in the design doc.

pkg/client/gen.go
johnSchnake: Should be `PluginSelections!=nil`

johnSchnake:
Added tests for the actual plugin `Set` function which pulled in a test dependency which was useful in figuring out the problem. Clarified a few of the error messages too with quotes and such.

Also added the logic for erroring if the same plugin is set multiple times. The reasoning is that the server's behavior is unpredictable. It will actually try to run both but I think it just overwrites the files because it uses those for a predictable name.

If we have a reason to support running a plugin N times then we need to figure out how we'd want to do the naming so it is predictable and most easily maintained (obviously just numbering them seems to make the most sense, but I'm not sure if there are any other odd impacts or, honestly, if there are any other issues with getting/mixing up the results as they come in)

cmd/sonobuoy/app/args.go
johnSchnake: Done.

johnSchnake:
Added another test to clarify some logic surrounding if someone were to manually send yaml/config data that had the older "PluginSelection" but also specified plugins via the dynamic/static objects. Things work as expected and the YAML has both plugins listed (so the server will load them) but the PluginSelection object is still sent and the aggregation server deliberately filters out the unwanted plugin.

Eventually, I want to remove some of that logic but in the meantime, it would be nice to continue to allow yaml made with older sonobuoy versions (and saved in source control) to still work the same.

I think this is good to go now and I've run it through the automated tests, added numerous tests, and ran more automated tests that the behavior is like I'd expect.

Will merge in a few hours if no further comments need addressing.

PR #684, opened 2019-04-22
Run all loaded plugins by default

johnSchnake:
The aggregator has a plugin filtering mechanism that
seems to actually complicate things for most users. Beyond
adding their plugin to the config map, they also have to
adjust the plugin selection.

This seems to have originated due to a desire for
flexibility (before knowing we needed it or not) and also
ease of testing.

This change just introduces special handling for the nil
selection (not empty, completely nil). In that case, all
plugins are run which are loaded.

ref #405

I copied some plugin manifests in the testing directory because otherwise, you get an error when you try and run the invalid manifest.

**Release note**:
```
The aggregator pod will now run all plugins found in the search paths unless given a non-nil value for `Plugins` in the config.json settings.
```

### DIFF
--- a/pkg/plugin/loader/loader.go
+++ b/pkg/plugin/loader/loader.go
@@ -73,7 +73,11 @@ func LoadAllPlugins(namespace, sonobuoyImage, imagePullPolicy string, searchPath
 		pluginDefinitions = append(pluginDefinitions, pluginDefinition)
 	}
 
-	pluginDefinitions = filterPluginDef(pluginDefinitions, selections)
+	// The zero value for the slice, nil, indicates that all plugins should be run.
+	// If the user wants to run 0 plugins, they should explicitly pass an empty array.
+	if selections != nil {
+		pluginDefinitions = filterPluginDef(pluginDefinitions, selections)
+	}
 
 	plugins := []plugin.Interface{}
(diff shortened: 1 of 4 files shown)


pkg/plugin/loader/loader.go
```diff
@@ -73,7 +73,11 @@ func LoadAllPlugins(namespace, sonobuoyImage, imagePullPolicy string, searchPath
 		pluginDefinitions = append(pluginDefinitions, pluginDefinition)
 	}
 
-	pluginDefinitions = filterPluginDef(pluginDefinitions, selections)
+	// The zero value for the slice, nil, indicates that all plugins should be run.
+	// If the user wants to run 0 plugins, they should explicitly pass an empty array.
```
stevesloka: What's the use-case to not run any plugins? Should that be an error?

johnSchnake: So currently the use case for that, not that I'm sure anyone uses it, is that you want to do either:
 - run/test sonobuoy works at all
 - run sonobuoy to just get the query data

So the reason I want to continue to allow that is more about limiting truly breaking changes when they aren't really necessary. I agree it is not a common flow, but I don't really see a reason to not allow it if someone wants to explicitly request it.

PR #686, opened 2019-04-22
Add CLI flag for aggregation timeout

### LINKED ISSUE #482: Make sonobuoy timeout configurable from the CLI

johnSchnake:
One of the most common config settings that may change between
different environments and test cases is the amount of time
Sonobuoy will wait for results. Currently, this value is configurable
but it is in a nested config structure and not terribly well
documented at this time.

Personally, in the past, I had projects where we had to start
using a sonobuoy config-based flow instead of flags due solely
to this field.

This change adds the flag to make it more easy to set.

Fixes #482

**Release note**:
```
Added --timeout flag to set the number of seconds Sonobuoy will wait for plugins to complete.
```

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -38,6 +38,7 @@ const (
 	sonobuoyImageFlag   = "sonobuoy-image"
 	imagePullPolicyFlag = "image-pull-policy"
 	pluginFlag          = "plugin"
+	timeoutFlag         = "timeout"
 )
 
 // AddNamespaceFlag initialises a namespace flag.
@@ -251,6 +252,14 @@ func AddRunWaitFlag(flag *int, flags *pflag.FlagSet) {
 		"Wait for sonobuoy run to be completed (or fail). 0 indicates do not wait. By providing --wait the default is to wait for 1 day.",
 	)
 	flags.Lookup("wait").NoOptDefVal = "1440"
+}
+
+// AddTimeoutFlag adds an int flag for waiting for the entire run to finish.
+func AddTimeoutFlag(flag *int, flags *pflag.FlagSet) {
+	flags.IntVar(
+		flag, timeoutFlag, config.DefaultAggregationServerTimeoutSeconds,
+		"How long (in seconds) Sonobuoy will wait for plugins to complete before exiting. 0 indicates no timeout.",
+	)
 }
 
 // AddImagePullPolicyFlag adds a boolean flag for deleting everything (including E2E tests).
--- a/cmd/sonobuoy/app/gen.go
+++ b/cmd/sonobuoy/app/gen.go
@@ -46,6 +46,7 @@ type genFlags struct {
 	kubeConformanceImageVersion imagepkg.ConformanceImageVersion
 	imagePullPolicy             ImagePullPolicy
 	e2eRepoList                 string
+	timeoutSeconds              int
 
 	// plugins will keep a list of the plugins we want. Custom type for
 	// flag support.
@@ -68,6 +69,7 @@ func GenFlagSet(cfg *genFlags, rbac RBACMode) *pflag.FlagSet {
 	cfg.e2eflags = AddE2EConfigFlags(genset)
 	AddRBACModeFlags(&cfg.rbacMode, genset, rbac)
 	AddImagePullPolicyFlag(&cfg.imagePullPolicy, genset)
+	AddTimeoutFlag(&cfg.timeoutSeconds, genset)
 
 	AddNamespaceFlag(&cfg.namespace, genset)
 	AddSonobuoyImage(&cfg.sonobuoyImage, genset)
@@ -274,6 +276,10 @@ func (g *genFlags) resolveConfig() *config.Config {
 
 	if g.genflags.Changed(imagePullPolicyFlag) {
 		conf.ImagePullPolicy = g.imagePullPolicy.String()
+	}
+
+	if g.genflags.Changed(timeoutFlag) {
+		conf.Aggregation.TimeoutSeconds = g.timeoutSeconds
 	}
 
 	return conf
--- a/cmd/sonobuoy/app/gen_test.go
+++ b/cmd/sonobuoy/app/gen_test.go
@@ -79,6 +79,7 @@ func TestResolveConformanceImage(t *testing.T) {
 
 func TestResolveConfig(t *testing.T) {
 	defaultPluginSearchPath := config.New().PluginSearchPath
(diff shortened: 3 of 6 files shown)


johnSchnake:
// @ceridwen

stevesloka:
So now there are 2 flags which determine "wait". There's `--wait` and now `--timeout` (added in this PR). Should those do the same thing and be combined?

cmd/sonobuoy/app/args.go
```diff
@@ -253,6 +254,14 @@ func AddRunWaitFlag(flag *int, flags *pflag.FlagSet) {
 	flags.Lookup("wait").NoOptDefVal = "1440"
 }
 
+// AddTimeoutFlag adds an int flag for waiting for the entire run to finish.
+func AddTimeoutFlag(flag *int, flags *pflag.FlagSet) {
+	flags.IntVar(
+		flag, timeoutFlag, config.DefaultAggregationServerTimeoutSeconds,
+		"How long (in seconds) Sonobuoy will wait for plugins to complete before exiting. 0 indicates no timeout.",
```
stevesloka: Might be cool to make this a duration type and let users enter `1m` or 2h`, etc. Otherwise, you've got to do the math to figure out the number of seconds.

stevesloka: nit: Maybe print the default to the help so users know what it is if not defined?

stevesloka:
This could be a future design PR, but is there a way to define the timeout per plugin? For instance, the e2e tests may take a long time, but maybe my systemd_logs only takes 2mins? This flag could be the default of all and then later maybe let plugin developers further define?

cmd/sonobuoy/app/args.go
johnSchnake: cobra adds the default; this was copied from my console just now:
```
... before exiting. 0 indicates no timeout. (default 10800)
```

johnSchnake:
> So now there are 2 flags which determine "wait". There's --wait and now --timeout (added in this PR). Should those do the same thing and be combined?

So there are two flags but they really do indicate different things and could be used in various ways.

The `--timeout` flag indicates the server timeout and when that is hit, everything turns off and explodes (more or less).

The `--wait` flag is a CLI-only flag that just indicates how long you are willing to block in the terminal for. It won't err the run just because you hit that limit.

The reason I don't want to sync the two is that a pretty reasonable use case is the following:
 - run and allow the server to timeout after 2hr
 - you really dont think it'll take that long so you just have your CLI stop waiting and return after an hour.
 - after the hour is up you can check the status; you assume it should be done but if not (and it is still running) you can investigate everything while it is still running.

johnSchnake:
> This could be a future design PR, but is there a way to define the timeout per plugin? For instance, the e2e tests may take a long time, but maybe my systemd_logs only takes 2mins? This flag could be the default of all and then later maybe let plugin developers further define?

So I agree something like that may eventually be implemented. There are definitely impacts of this overall timeout and per-plugin timeout that play a role in #627 My POC implementation of plugin ordering also showed me that it could be very easy to adjust some of these details by adding them into the plugin manifest metadata.

But yes, I agree it is for another day since this is currently coded for just one big timeout and this was just to make it easier to set.

stevesloka:
Yeah that holds up, but I can still `sonobuoy run --wait`, then CTRL-C and then `sonobuoy status` and all works. 

> The --wait flag is a CLI-only flag that just indicates how long you are willing to block in the terminal for. It won't err the run just because you hit that limit.

But should it? If you had CI running and this happened you'd get a false positive wouldn't you? 

I'm just trying to challenge the overall design to make sure we're solving the problem and not just throwing flags at it. Is there a scenario where you'd want the client to wait 5 mins and the server wait 2 hours?

johnSchnake:
> Yeah that holds up, but I can still sonobuoy run --wait, then CTRL-C and then sonobuoy status and all works.

True, though you may want to do something like this even in CI. Before the `--wait` commands existed thats what I had seen done when people were rolling their own polling logic. Wait while it runs for a while, then when we hit our own timeout logic go and gather logs and maybe run some checks while it Sonobuoy is still running, then err out ourselves and clean up.

> But should it? If you had CI running and this happened you'd get a false positive wouldn't you?

As the example above says, I don't think forcing the server to error and stop running because we hit a CLI timeout is always the right thing to do. It prevents a user (including CI systems) from investigating if they want to.

Just because `sonobuoy run --wait` returns doesn't guarantee anything about whether or not anything succeeded; do you mean `false positive` in that you might think it is done? That is partially why the default time for `--wait` is much longer than any typical run, currently at a day.

> I'm just trying to challenge the overall design to make sure we're solving the problem and not just throwing flags at it. 

Absolutely 👍 , IMO that is the discussion that we had when deciding to put this ticket into the milestone though. I appreciate the design questions though and am happy to defend the choice to make the new flag.

> Is there a scenario where you'd want the client to wait 5 mins and the server wait 2 hours?

Yes but I also think your comment about CTRL+C plays a role here too; personally, I run checks all the time where this is the [somewhat] the case. If I am iterating on some plugin (in my case, e2e test) and I _think_ it will just take 5m or less and I run it as such. If it runs and I find out it actually took 7m, I _really_ didn't want the server to err at 5m. But the CLI `--wait` gave me an opportunity to see that 5m has passed and it still isn't done. Again, then I investigate the situation and usually restart my experiment whenever I'm done.

TL;DR; I think that in most situations your sentiment is right, you'd just want the server to pass/fail on its own and then the CLI wait for that terminal status. But I can think of situations where you want to ensure the CLI stops and gives you time to investigate before cleaning anything up. I think some of these things are also related to when/how/if plugin resources are cleaned up. In the end though, this is already a customizable value, just one that multiple people edit regularly and want it to be more simple.

If we decide to prioritize per-plugin timeouts, cleanup logic, etc then I think we should bring up these ideas again though because I do acknowledge there are surely some improvements that could be made.

Issue #630, opened 2019-03-13
Queries should use dynamic client


johnSchnake:
**Describe the solution you'd like**
Changing the queries to use the dynamic client help prevent worrying about API drift.

**Anything else you would like to add:**
I think #492 would be applicable here if I'm understanding it correctly. With the dynamic client we can walk the API resources and pull whatever we want so either we can choose _everything_ or need to audit the items we are gathering.

Related to #432

johnSchnake:
Moving conversation from #492 since it is dynamic client focused.

I was informed of the rate limiting that is built into the rest config objects; I'm not 100% how this really interacts with the kubelt/api server, but I was able to set:

```
	restConf.QPS = float32(30)
	restConf.Burst = 100
```

And get all the queries to execute _much_ faster. I'm not 100% sure how this works out since I also found in my logs that the qps and burst for the API server were 5/10 respectively. But the queries (as I got them) were still reported without error.

I also learned that if you just set the QPS and not the burst, the client just hangs. TIL.

A little more investigation to help ensure we get these values set appropriately (and we should make them editable in the config in case users rn into issues) but then I should be able to put up a PR for this.

One or two more wrinkles to work out too w.r.t. how the resource filter list is used. I'm not sure if/who is really using it but with the dyanmic client it might make sense to further specify the resource filenames (or in the filter list) so that if CRDs have the same names (but are in different API groups) things work as expected.

PR #688, opened 2019-04-23
Update client go and other deps

### LINKED ISSUE #687: Update dependencies
Client-go, in particular, is many versions out of date (8.0 instead of 11.0) and their github says 8.0 is deprecated and we should update. 11.0 is the version that is used with k8s v1.14.

Other dependencies were updated as well to be compatible.

Other deps may need to be updated as well to be compatible; I don't have a reason to _not_ update dependencies so that is fine with me.

johnSchnake:
Updated client-go to 11.0 and the other k8s code to 1.14 to be
compatible.

Ran into other build issues due to problems with other deps being
incompatible so I updated those as well via `dep ensure -update`

Fixes #687
Ref #630

**Release note**:
```
None
```


johnSchnake:
Reproduced issue locally; the problem is that `sonobuoy master` no longer has the verbose flag. A few flags were added via dependencies here: https://github.com/heptio/sonobuoy/blob/master/cmd/sonobuoy/app/root.go#L53

Looking at the resolution for this.

johnSchnake:
Issue is that glog was just from client-go; client-go moved away from glog to klog here: https://github.com/kubernetes/client-go/commit/74cd8bbeee080de457e0ef6b4ad2951c205ade4d

You just have to manually initialize the flags with `InitFlags(nil)`

johnSchnake:
Just for info as well so you can see where large changes were (since the diff is big):

`git diff upstream/master --stat | grep -v "files changed" | awk -F "|" '{print $2 " " $1}'` yielded:

```
 8992 +++++++++++---------  vendor/k8s.io/api/core/v1/generated.pb.go
 6853 +++++++++++++++  vendor/golang.org/x/sys/windows/zerrors_windows.go
 4643 +++++-----  .../golang.org/x/sys/unix/zerrors_linux_sparc64.go
 4419 ++++++++++  .../k8s.io/api/autoscaling/v2beta2/generated.pb.go
 3006 +++++++  .../gogo/protobuf/proto/table_marshal.go
 2887 +++++++  .../protoc-gen-go/descriptor/descriptor.pb.go
 2822 ++++++  .../golang.org/x/sys/unix/zerrors_linux_riscv64.go
 2806 ++++++  .../protobuf/protoc-gen-go/generator/generator.go
 2589 ++++--  .../pkg/apis/apiextensions/v1beta1/generated.pb.go
 2266 +++--  .../k8s.io/api/extensions/v1beta1/generated.pb.go
 2250 +++++  .../x/sys/unix/zsyscall_linux_riscv64.go
 2245 +++++  .../gogo/protobuf/proto/table_unmarshal.go
 2207 +++--  .../apimachinery/pkg/apis/meta/v1/generated.pb.go
 2173 +++--  .../internal/datastore/datastore_v3.pb.go
 2154 +++++  .../golang.org/x/sys/unix/ztypes_linux_riscv64.go
...
```

johnSchnake:
Its odd that the `signed-off-by` is reported as missing; you can look at the one commit and see it is not.

Regardless, I'll rebase and see if it goes away.

johnSchnake:
Merging. Also think it is a good thing if we review/update deps per release so each diff is _much_ smaller than this.

Issue #432, opened 2018-05-22
Extract queries to their own plugin


chuckha:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

I propose that we extract queries out of core sonobuoy into a plugin. This code makes versioning sonobuoy against k8s painful:

https://github.com/heptio/sonobuoy/blob/3c2b66bf5d3d5692111183301c210cc334f52a95/pkg/discovery/queries.go

Without this, I believe the only other place we use the k8s api is the driver code which is now, as of 1.10 or 1.11 (k8s) only using stable APIs. 

** Why? **:

We should do this for two reasons:

### Stable API

k8s has now stabilized (as of 1.10) the APIs used. We can write a little extra code to ensure we have the compatibility guarantees we've made while the older APIs get deprecated. 
After v1.12 of k8s is released we will only need to support the stable apis which is a lot nicer than the code linked to above with all of the changing / adding apis. Technically we wouldn't need to update client-go for the sonobuoy repo again until jobs or daemonset changes (but as its stable this is unlikely to happen for a while) or when some unforeseen event takes place (sonobuoy's scope expands, new drivers added that are in alpha, etc).

### Reduced scope of sonobuoy

This change would reduce the scope of sonobuoy from having to run queries against the k8s api and serializing results and all that. That would be part of the query plugin we write. Sonobuoy would then be an aggregator that spawns plugin workers and aggregates their results and nothing more.

timothysc:
It would still be painful if it were it's own plugin.   I think prioritizing the use of the dymanic client would be more fruitful to solve the api drift problem.  

Ideally I'd like both the dymanic client and a DAG model of execution which includes the queries, but most folks use sonobuoy with a pretty singular focus.  So I think the plugin change is of less importance vs. the dynamic client.  

Thoughts?

chuckha:
I think both would be good to do.

To me it feels like sonobuoy is the aggregator/plugin driver and beyond that is scope creep.

As for getting a DAG in place, if we want a DAG that doesn't have to make a special case for the queries, I think extracting them to a plugin would be the best way to go.

But yes, dynamic client is the higher priority issue here for sure 👍

johnSchnake:
> So I think the plugin change is of less importance ...

[...]

PR #689, opened 2019-04-24
Dynamic client

### LINKED ISSUE #492: Audit default resource gathering
There are additional api-resources that could be gathered 
```
$ kubectl api-resources
NAME                              SHORTNAMES   APIGROUP                       NAMESPACED   KIND
bindings                                                                      true         Binding
componentstatuses                 cs                                          false        ComponentStatus
configmaps                        cm                                          true         ConfigMap
endpoints                         ep                                          true         Endpoints
events                            ev                                          true         Event
limitranges                       limits                                      true         LimitRange
namespaces                        ns                                          false        Namespace
nodes                             no                                          false        Node
persistentvolumeclaims            pvc                                         true         PersistentVolumeClaim
persistentvolumes                 pv                                          false        PersistentVolume
pods                              po                                          true         Pod
podtemplates                                                                  true         PodTemplate
replicationcontrollers            rc                                          true         ReplicationController
resourcequotas                    quota                                       true         ResourceQuota
secrets                                                                       true         Secret
serviceaccounts                   sa                                          true         ServiceAccount
services                          svc                                         true         Service
mutatingwebhookconfigurations                  admissionregistration.k8s.io   false        MutatingWebhookConfiguration
validatingwebhookconfigurations                admissionregistration.k8s.io   false        ValidatingWebhookConfiguration
customresourcedefinitions         crd,crds     apiextensions.k8s.io           false        CustomResourceDefinition
apiservices                                    apiregistration.k8s.io         false        APIService
controllerrevisions                            apps                           true         ControllerRevision
daemonsets                        ds           apps                           true         DaemonSet
deployments                       deploy       apps                           true         Deployment
replicasets                       rs           apps                           true         ReplicaSet
statefulsets                      sts          apps                           true         StatefulSet
tokenreviews                                   authentication.k8s.io          false        TokenReview
localsubjectaccessreviews                      authorization.k8s.io           true         LocalSubjectAccessReview
selfsubjectaccessreviews                       authorization.k8s.io           false        SelfSubjectAccessReview
selfsubjectrulesreviews                        authorization.k8s.io           false        SelfSubjectRulesReview
subjectaccessreviews                           authorization.k8s.io           false        SubjectAccessReview
horizontalpodautoscalers          hpa          autoscaling                    true         HorizontalPodAutoscaler
cronjobs                          cj           batch                          true         CronJob
jobs                                           batch                          true         Job
certificatesigningrequests        csr          certificates.k8s.io            false        CertificateSigningRequest
bgpconfigurations                              crd.projectcalico.org          false        BGPConfiguration
bgppeers                                       crd.projectcalico.org          false        BGPPeer
clusterinformations                            crd.projectcalico.org          false        ClusterInformation
felixconfigurations                            crd.projectcalico.org          false        FelixConfiguration
globalnetworkpolicies                          crd.projectcalico.org          false        GlobalNetworkPolicy
globalnetworksets                              crd.projectcalico.org          false        GlobalNetworkSet
hostendpoints                                  crd.projectcalico.org          false        HostEndpoint
ippools                                        crd.projectcalico.org          false        IPPool
networkpolicies                                crd.projectcalico.org          true         NetworkPolicy
events                            ev           events.k8s.io                  true         Event
daemonsets                        ds           extensions                     true         DaemonSet
deployments                       deploy       extensions                     true         Deployment
ingresses                         ing          extensions                     true         Ingress
networkpolicies                   netpol       extensions                     true         NetworkPolicy
podsecuritypolicies               psp          extensions                     false        PodSecurityPolicy
replicasets                       rs           extensions                     true         ReplicaSet
networkpolicies                   netpol       networking.k8s.io              true         NetworkPolicy
poddisruptionbudgets              pdb          policy                         true         PodDisruptionBudget
podsecuritypolicies               psp          policy                         false        PodSecurityPolicy
clusterrolebindings                            rbac.authorization.k8s.io      false        ClusterRoleBinding
clusterroles                                   rbac.authorization.k8s.io      false        ClusterRole
rolebindings                                   rbac.authorization.k8s.io      true         RoleBinding
roles                                          rbac.authorization.k8s.io      true         Role
priorityclasses                   pc           scheduling.k8s.io              false        PriorityClass
storageclasses                    sc           storage.k8s.io                 false        StorageClass
volumeattachments                              storage.k8s.io                 false        VolumeAttachment
```

**Environment:**

- Sonobuoy version: 0.11.3+

### LINKED ISSUE #630: Queries should use dynamic client

johnSchnake:
This change is fairly substantial so I wanted to specify
a few details:
 - All queries are now executed with the discovery/dynamic
clients walking the API server
 - QPS and Burst can be set on the config.json so that you can
speed up/slow down your queries based on speed and throttling.
 - The filenames containing the the resource query results are
updated to list group_version_resource.json as opposed to
resource.json. This makes it explicitly clear where the resources
came from and avoids a problem where different groups with the
same resource name could collide.
 - To avoid collision in general though, the queries only grab
the first resource with a given name that it encounters. This
is the behavior kubectl uses as well.
 - Resource filtering via the config.json field Resource
is still done, but explicitly compared against the resource.Name
field given by the server. Most visibly, the casing of the values
is different than it used to be.
 - If an nil Resource value is set in the config.json, all values
will be queried with the exception of secrets.
 - If an empty array of Resources is set in the config.json, no
values will be queried, same as today.
 - The default resource list exists but has a few extra values
added to it. In the future we may remove it entirely and just
default to query all. This was kept so make the casing changes
more clear to users of sonobuoy gen.

Fixes #492
Fixes #630

I really was thinking originally about changing the filtering logic more but I really left it as-is for the most part, including the default resource list. In theory, I could remove that and it would get even more info, but I didn't want to, by default, be getting events/hpa for each namespace (that was a deliberate choice in the past).

Effectively I think the current model still has shortcomings and is not the easiest to configure. This can be fixed when we create a query plugin though since it will be effectively a small green field project then and we can redesign the contract based on what we see. As a plugin it will also be trivial to opt out of the queries now that we have the `--plugin` flag.

**Release note**:
```
The Resources field in the Sonobuoy config file is now checked explicitly against the resource.Name of the objects on the server. The most apparent part of this change is that the values are all lowercase.
```


pkg/discovery/discovery.go
```diff
@@ -149,6 +169,8 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 	logrus.Infof("Results available at %v", tb)
 
 	return errCount
+
+	return 0
```
johnSchnake: debug thing; remove

pkg/dynamic/client.go
```diff
@@ -79,7 +92,172 @@ func (a *APIHelper) CreateObject(obj *unstructured.Unstructured) (*unstructured.
...
+	results := map[string][]unstructured.Unstructured{}
+	for _, gvr := range allResources {
+		fmt.Println("gvr:", gvr)
+		resourceClient := a.Client.Resource(gvr)
+
+		resources, err := resourceClient.List(listOpt)
+		if err != nil {
+			return nil, errors.Wrapf(err, "listing resource %v", gvr)
+		}
+		/*
```
johnSchnake: remove debug comments

pkg/dynamic/client.go
```diff
@@ -79,7 +92,172 @@ func (a *APIHelper) CreateObject(obj *unstructured.Unstructured) (*unstructured.
...
+	resourceMap, err := a.DiscoveryClient.ServerPreferredResources()
+	_ = err
+	allResources := map[schema.GroupVersionResource]bool{}
+	for _, apiResourceList := range resourceMap {
+		version, err := schema.ParseGroupVersion(apiResourceList.GroupVersion)
+		if err != nil {
+			fmt.Println("schnake parsing scheme err", err)
+		}
+		for _, apiResource := range apiResourceList.APIResources {
+			fmt.Printf("apiresource info: %#v\n\n", apiResource)
```
johnSchnake: remove debug output throughout

cmd/sonobuoy/app/master.go
```diff
@@ -49,11 +48,12 @@ func NewCmdMaster() *cobra.Command {
 
 func runMaster(cmd *cobra.Command, args []string) {
 
-	cfg, err := config.LoadConfig()
+	/*cfg, err := config.LoadConfig()
```
johnSchnake: This was for running locally/debugging. Do not merge this.

pkg/client/interfaces.go
```diff
@@ -94,6 +94,9 @@ type SonobuoyKubeAPIClient interface {
 	Name(*unstructured.Unstructured) (string, error)
 	Namespace(*unstructured.Unstructured) (string, error)
 	ResourceVersion(*unstructured.Unstructured) (string, error)
+
+	ListNSResources(ns string, filterResources []string) (map[string][]unstructured.Unstructured, error)
+	//ListClusterResources(filterResources []string) ([]unstructured.Unstructured, error)
```
johnSchnake: I dont think these are needed; it is all in the discovery code and will be eventually extracted to its own plugin.

cmd/sonobuoy/app/master.go
```diff
@@ -29,6 +29,7 @@ import (
 
 var noExit bool
 var kubecfg Kubeconfig
+var configPath string
```
johnSchnake: Remove

cmd/sonobuoy/app/master.go
```diff
@@ -44,6 +45,7 @@ func NewCmdMaster() *cobra.Command {
 		"Use this if you want sonobuoy to block and not exit. Useful when you want to explicitly grab results.tar.gz",
 	)
 	AddKubeconfigFlag(&kubecfg, cmd.Flags())
+
```
johnSchnake: remove

pkg/config/config.go
```diff
@@ -165,22 +169,16 @@ type Config struct {
...
 	///////////////////////////////////////////////
-	// Data collection options
+	// Query options
 	///////////////////////////////////////////////
-	Resources []string `json:"Resources" mapstructure:"Resources"`
+	Resources []string      `json:"Resources" mapstructure:"Resources"`
+	Filters   FilterOptions `json:"Filters" mapstructure:"Filters"`
+	Limits    LimitConfig   `json:"Limits" mapstructure:"Limits"`
+	QPS       float32       `json:"QPS,omitempty" mapstructure:"QPS"`
+	Burst     int           `json:"Burst,omitempty" mapstructure:"Burst"`
```
johnSchnake: Slight nit to aggregate things here but it makes it easier to grok what the options are and how they're used.

pkg/discovery/queries.go
```diff
@@ -217,101 +167,146 @@ func QueryNSResources(kubeClient kubernetes.Interface, recorder *QueryRecorder,
...
+			if !listable {
+				continue
 			}
-			nodeNames := make([]string, len(nodeList.Items))
-			for i, node := range nodeList.Items {
-				nodeNames[i] = node.Name
+
+			// Filter for explicit values
+			if len(wantResources) > 0 {
+				if !sliceContains(wantResources, res.Name) {
```
johnSchnake: This isn't sufficient right now because the list we were using before was all custom code; it didn't necessary correspond to group/kind/resource (usually very close, but may be singular vs plural or casing was different).

johnSchnake: So the defaults can actually just be removed, though people who run a custom set of resources would then be hit by this.

For maximum backward compatibility, we would need to ensure it works if you pass in an older resource list (I think this whole idea needs some review/design though since there are so many resource types it doesnt make sense for someone to have to specify each and every one they want).

I could add the lower casing in the function but it gets called quite a bit; I might also want to do it when reading in their config. The only thing is that I'm not certain whether or not the name _could_ be upper case, e.g. for CRDs. In which case lower casing the list on load time and then we'd have to lower case the resource name when checking. That starts to get awkward to spread that logic.

Is this a breaking change we want to allow? It avoids inefficiencies of doing case checking or spreading logic into a few different places to solve one problem. Would also only affect people who are running a newer version of sonobuoy with YAML produced by an older version of Sonobuoy. Maybe since this only affects the queries (and no the overall functioning of Sonobuoy or plugins) we can just use a release note to comment on it?

@chuckha curious if you have a thought on this?

pkg/config/config.go
```diff
@@ -58,88 +62,6 @@ var (
...
-// TODO (tstclair): Clean up defaulting, this is temporary
-var NamespacedDefaults = []string{
-	"ConfigMaps",
-	"ControllerRevisions",
-	"CronJobs",
-	"DaemonSets",
-	"Deployments",
-	"Endpoints",
-	//"Events",
-	//"HorizontalPodAutoscalers",
```
johnSchnake: Why was this skipped at all?

pkg/config/config.go
```diff
@@ -58,88 +62,6 @@ var (
...
-// NamespacedDefaults is the default-list of API resources to gather
-// TODO (tstclair): Clean up defaulting, this is temporary
-var NamespacedDefaults = []string{
-	"ConfigMaps",
-	"ControllerRevisions",
-	"CronJobs",
-	"DaemonSets",
-	"Deployments",
-	"Endpoints",
-	//"Events",
```
johnSchnake: Assuming this was skipped for just concern over there being a lot of them? Or was it for some security concern?

pkg/config/config.go
```diff
@@ -58,88 +62,6 @@ var (
...
-	//"HorizontalPodAutoscalers",
-	"Ingresses",
-	"Jobs",
-	"LimitRanges",
-	"NetworkPolicies",
-	"PersistentVolumeClaims",
-	"PodDisruptionBudgets",
-	//"PodLogs",
```
johnSchnake: Need to review the pod logs logic right now. I think as I coded it, it _only_ captures the logs when in the heptio-sonobuoy namespace. Similar question about events and HPA, as coded now, we always get those in the default crawl for stuff.

So the question becomes, how do the users get pod logs in all namespaces if they wanted? They used to add it to the list of resources, but the default list is empty (indicating to dynamically get everything).

A long, manually maintained list like this is awkward for users and devs alike. I think it makes sense to be more opinionated even if it makes a slightly breaking change in behavior. If we do that, I'd just want to get slightly more data instead of erring out and getting no data.

This manual list had the draw back that you had no clear idea what you could add, but if you figured it out you would add the one item and it would work. With allowing the empty value to be 'all things' it makes it harder to specify things like "all but one" or "default and one more thing".

Thinking...

johnSchnake: At the moment I've feel like I've come to the right conclusion: I was being too hasty and trying to remove too much here for the sake of "simplicity". Removing all these defaults is in a sense simple, but changing the model completely like that does either break specific tasks/flows or leaves some questions to be decided suddenly about what to do/how to do it.

Instead, I'm going to opt for the less dramatic change:
 - continue to have a list of default options we query for
 - update the names in the list so I don't have to worry about all the casing concerns; its not a fun little section of change to double check spelling of 30-50 constants but it is what it is
 - if someone still wants to really query _everything_ they can just remove the list of resources so it is nil and implicitly get _everything_.

This will enable me to maintain everything about the current flow I think: in the "sonobuoy" namespace we can remove the same values as we do now, if someone wants to add podlogs or secrets its as easy as adding something to the list.

When we extract queries to their own plugin we can modify more of this behavior in a more natural way.

pkg/config/config.go
johnSchnake: Kept for historical sake; not worth a fight at the current time. This stuff will change when we make it its own plugin.

pkg/discovery/utils.go
```diff
@@ -49,31 +50,14 @@ func FilterNamespaces(kubeClient kubernetes.Interface, filter string) ([]string,
 // SerializeObj will write out an object
 func SerializeObj(obj interface{}, outpath string, file string) error {
 	var err error
-	if err = os.MkdirAll(outpath, 0755); err == nil {
-		if eJSONBytes, err := json.Marshal(obj); err == nil {
-			err = ioutil.WriteFile(outpath+"/"+file, eJSONBytes, 0644)
-		}
+	if err = os.MkdirAll(outpath, 0755); err != nil {
```
johnSchnake: These changes to the serializers were the result of tearing out a bunch and putting back in what I needed. TL; DR; we just didnt need all these functions anymore and the ones we did could be written a bit more idiomatically to make it easier to read.

pkg/discovery/queries.go
```diff
@@ -17,33 +17,23 @@ limitations under the License.
...
 )
 
-// ObjQuery is a query function that returns a kubernetes object
-type ObjQuery func() (runtime.Object, error)
-
-// UntypedQuery is a query function that return an untyped array of objs
-type UntypedQuery func() (interface{}, error)
-
-// UntypedListQuery is a query function that return an untyped array of objs
-type UntypedListQuery func() ([]interface{}, error)
```
johnSchnake: The dynamic client has some different types so I didn't need these anymore.

johnSchnake:
Going to merge this with just lazy consensus by tomorrow morning; I know it the weekend but I wanted to at least vocalize this was happening in case someone had a particular and immediate concern.

The easiest way to test this is to just have a cluster up and run `sonobuoy master`. You'll need a `config.json` file but you can actually just set `{"Resources":[...]}`

You can then look at the output and check the resulting tarball for the resources you expected.

```
go build
echo '{"Resources":["pods"]}' > config.json
./sonobuoy master
```

I actually _added_ some tests about filtering the resources which didn't exist before but in the end the coverage technically went down slightly. IMO when we split this out into its own plugin (#432 ) we can work on the query specific testing/coverage as much (or little) as we care to.

Another big thing to note is that Sonobuoy logs resources that it is explicitly not querying for. This clarity is important I think when trying to diagnose problems (which is the whole point of the queries).

cc @stevesloka @timothysc @chuckha

PR #697, opened 2019-04-29
Add example for command runner plugin

johnSchnake:
With the new plugin generation/runner capabilities we needed
a new example to show users what they can do/how to do it.

This example just takes all the arguments it is given and
runs them with /bin/sh, saves the output, tars it up, and
tells the Sonobuoy worker where to find the results.

This example includes:
 - the code for the plugin (a single bash script)
 - the Dockerfile to build the script
 - a README.md to explain what it does and how to use it

**Release note**:
```
NONE
```

### DIFF
--- a/examples/plugins/cmd-runner/Dockerfile
+++ b/examples/plugins/cmd-runner/Dockerfile
@@ -1,0 +1,19 @@
+FROM debian:stretch-slim
+
+# Install kubectl
+# Note: Latest version may be found on:
+# https://aur.archlinux.org/packages/kubectl-bin/
+ADD https://storage.googleapis.com/kubernetes-release/release/v1.14.1/bin/linux/amd64/kubectl /usr/local/bin/kubectl
+
+ENV HOME=/config
+
+# Basic check it works.
+RUN apt-get update && \
+    apt-get -y install net-tools && \
+    apt-get -y install curl && \
+    chmod +x /usr/local/bin/kubectl && \
+    kubectl version --client
+
+COPY ./run.sh ./run.sh 
+
+ENTRYPOINT ["./run.sh"]
--- a/examples/plugins/cmd-runner/README.md
+++ b/examples/plugins/cmd-runner/README.md
@@ -1,0 +1,35 @@
+# Command Runner
+
+This example creates an Sonobuoy plugin which runs each of its arguments (via /bin/sh) and saves the output in a results file (the output for the Nth command is output to a file called "outN").
+
+Kubectl is added to the image so that it can grab information from/about the cluster.
+
+All of the results files are added into a tar file and then the 'done' file is written, telling the Sonobuoy worker where to find the output.
+
+## Example
+
+```
+$ docker build . -t user/easy-sonobuoy-cmds:v0.1
+$ docker push user/easy-sonobuoy-cmds:v0.1
+
+$ sonobuoy gen plugin \
+--name=hello-world \
+--image user/easy-sonobuoy-cmds:v0.1 \
+--arg="echo hello world" \
+--arg="kubectl cluster-info" > hello-world.yaml
+
+$ sonobuoy run --plugin hello-world.yaml
+```
+
+You can obtain all the Sonobuoy results locally by running:
+```
+outfile=$(sonobuoy retrieve) && \
+mkdir results && tar -xf $outfile -C results
+```
+
+Print our specific results with:
+```
+cat results/plugins/hello-world/results/out*
+```
+
+
--- a/examples/plugins/cmd-runner/run.sh
+++ b/examples/plugins/cmd-runner/run.sh
@@ -1,0 +1,42 @@
+#!/bin/sh
+
+set -x
+
+# This is the entrypoint for the image and meant to wrap the
+# logic of gathering/reporting results to the Sonobuoy worker.
+
+results_dir="${RESULTS_DIR:-/tmp/results}"
+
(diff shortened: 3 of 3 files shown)


examples/plugins/cmd-runner/README.md
```diff
@@ -0,0 +1,27 @@
...
+--name=hello-world \
+--image user/easy-sonobuoy-cmds:v0.1 \
+--arg="echo hello world" \
+--arg="kubectl cluster-info" > hello-world.yaml
+
+$ sonobuoy run --plugin hello-world.yaml
+
+$ outfile=$(sonobuoy retrieve) && \
+mkdir results && tar -xf $outfile -C results &&
+cat results/plugins/hello-world/results/out*
```
stevesloka: Maybe show where the output would be? Or describe where it is in the result output?

examples/plugins/cmd-runner/run.sh
```diff
@@ -0,0 +1,26 @@
...
+
+results_dir="${RESULTS_DIR:-/tmp/results}"
+
+if [[ $# -eq 0 ]]; then
+	echo "No arguments; expects each argument to be a shell command."
+	exit 1
+fi
+
+i=0
+while [ "$1" != "" ]; do
```
stevesloka: Might be good to put some comments throughout this, just to be clear what each piece is doing for new users.

examples/plugins/cmd-runner/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM alpine:3.8
```
stevesloka: Need to use a different image than alpine.

examples/plugins/cmd-runner/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Install kubectl
+# Note: Latest version may be found on:
+# https://aur.archlinux.org/packages/kubectl-bin/
+ADD https://storage.googleapis.com/kubernetes-release/release/v1.14.1/bin/linux/amd64/kubectl /usr/local/bin/kubectl
```
stevesloka: Could steal this if you want a generic image: https://github.com/heptio/kube-conformance/blob/master/Makefile#L23-L25

examples/plugins/cmd-runner/README.md
```diff
@@ -0,0 +1,27 @@
+# Command Runner
+
+This example creates an Sonobuoy plugin which runs each of its arguments (via /bin/sh) and saves the output in a results file.
+
+Kubectl is added to the image so that it can grab information from/about the cluster.
+
+All of the results files are added into a tar file and then the 'done' file is written, telling the Sonobuoy worker where to find the output.
```
stevesloka: We should document this process a bit, then link to it. I think the story still isn't told very well in our docs.

examples/plugins/cmd-runner/run.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -x
+
+# This is the entrypoint for the image and meant to wrap the
+# logic of gathering/reporting results to the Sonobuoy worker.
+
+results_dir="${RESULTS_DIR:-/tmp/results}"
+
+if [[ $# -eq 0 ]]; then
+	echo "No arguments; expects each argument to be a shell command."
```
stevesloka: If the user does this and doesn't pass args, what happens in Sonobuoy? Just curious the workflow to identify errors and what folks would run into.

examples/plugins/cmd-runner/Dockerfile
johnSchnake: Thanks; I got mixed up on which image I needed to use.

examples/plugins/cmd-runner/Dockerfile
johnSchnake: Yea I considered stealing that as well but I opted for the explicit version so people know what is in the image.

examples/plugins/cmd-runner/run.sh
johnSchnake: Thanks, good catch. This is actually the exact problem that the upstream k8s e2e tests had. It errored out and never told the worker it was done so Sonobuoy hangs. Will fix.

johnSchnake: Uses a `trap ... EXIT` now. When no arguments this error is dumped to `results_dir/out` just for consistent location/naming.

examples/plugins/cmd-runner/Dockerfile
johnSchnake: Uses debian:stretch-slim now

examples/plugins/cmd-runner/README.md
johnSchnake: Linking to https://github.com/heptio/sonobuoy/blob/master/docs/plugins.md

Was tempted to link directly to https://github.com/heptio/sonobuoy/blob/master/docs/plugins.md#contract but as we edit that file I dont want to break links just because we changed a header.

examples/plugins/cmd-runner/README.md
johnSchnake: Split up and clarified this a section a bit so they separate out getting the results vs printing these results specifically.

Also clarified the file naming for output at the top of this file.

johnSchnake:
Rebased (not that it matters for the example) and had to actually update this and add curl and ifconfig for my actual demo purposes.

PR #698, opened 2019-04-29
Fix servergroups query

johnSchnake:
In the dynamic query PR that merged recently there was a
copy/paste error which caused servergroups to get lost.

This fixes the issue and returns this functionality to its
original state.

To confirm this you can just run the master locally:
```
./sonobuoy master
```

**Release note**:
```
NONE
```

Marking as none since it was never broken in a release.


johnSchnake:
Merging; a trivially straightforward bug and fix. https://github.com/heptio/sonobuoy/pull/689/files#diff-799ef0bfefbd18913fd0fa8a3679e9e0L276 is the link to the bug introduction.

PR #705, opened 2019-05-03
Add documentation about how to pull private Sonobuoy image

### LINKED ISSUE #692: Create docs on how to pull sonobuoy from private registry
https://github.com/heptio/sonobuoy/pull/660 made this possible but its not really a simple flow because the secret needs to exist locally but sonobuoy handle that.

Outline of the process here (in case someone finds this useful before the docs are made):
 - Create credentials in a random namespace (default is fine) by following the directions here: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
 - Use `kubectl get secret -o yaml` to save that secret in yaml form and strip out the metadata like creation date and resource version.
 - Use `sonobuoy gen --config custom.json` to generate yaml for the run
 - Add the YAML for the secret to the file (ideally below the namespace/cluster items but above the aggregator pod)
 - Use `kubectl apply` to create everything

**Environment:**

- Sonobuoy version: 

johnSchnake:
The process to use a private Sonobuoy image is non-trivial since
you have to create a secret in the same namespace.

In the future, if this is common, we can simplify the process but
currently, this documentation should clarify it for the few people
who need to use this flow.

Fixes #692

**Release note**:
```
NONE
```

### DIFF
--- a/docs/pullsecrets.md
+++ b/docs/pullsecrets.md
@@ -1,0 +1,41 @@
+# Using a Private Sonobuoy Image with ImagePullSecrets
+
+This document describes how to use the ImagePullSecrets option in order to run Sonobuoy using a private Sonobuoy image.
+
+## Setting ImagePullSecrets
+
+The name of the secret to use when pulling the image can be set easily in the configuration file passed to `sonobuoy run` or `sonobuoy gen`:
+
+```
+echo '{"ImagePullSecrets":"mysecret"}' > secretconfig.json
+sonobuoy gen --config secretconfig.json
+```
+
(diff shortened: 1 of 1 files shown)


johnSchnake:
@stevesloka It is a bit of an awkward flow but, like the PR comment says, it is not commonly used but I wanted to document it in case it comes up again.

Take a look and let me know if you have any specific feedback. I'm not even linking this to the main README.md so it is just a resource we can point people to if they hit the issue and dont search for it themselves.

Going to assume consensus in a week.

PR #706, opened 2019-05-05
Check for container running before following logs

### LINKED ISSUE #703: Sonobuoy logs should start following new pods when they start

From the discussion in #301

ashish-amarnath:
Don't attempt to stream logs from containers until they are `Running`
- Fixes 

**Release note**:
```
Improves the `sonobuoy logs -f` command by waiting until containers are running to stream logs instead of failing outright and discontinuing logs from that container.
```

### DIFF
--- a/pkg/client/logs.go
+++ b/pkg/client/logs.go
@@ -20,10 +20,11 @@ import (
 	"fmt"
 	"io"
 	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
@@ -228,8 +229,46 @@ type logStreamer struct {
 	client             kubernetes.Interface
 }
 
+func (l *logStreamer) podName() string {
+	return fmt.Sprintf("%s/%s/%s", l.ns, l.pod, l.container)
+}
+
+func isContainerRunning(statuses *[]v1.ContainerStatus, containerName string) bool {
+	for _, cs := range *statuses {
+		if cs.Name == containerName && cs.State.Running != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func (l *logStreamer) waitForContainerRunning() error {
+	backoffSeconds := 1 * time.Second
+	for {
+		pod, err := l.client.CoreV1().Pods(l.ns).Get(l.pod, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get pod [%s/%s]", l.ns, l.pod)
+		}
+
+		if isContainerRunning(&pod.Status.ContainerStatuses, l.container) {
+			return nil
+		}
+
+		fmt.Printf("container %v, is not running, will retry streaming logs in %v seconds\n", l.podName(), backoffSeconds)
+		time.Sleep(backoffSeconds)
+		backoffSeconds *= 2
+	}
+}
+
 // stream will open a connection to the pod's logs and push messages onto a fan-in channel.
 func (l *logStreamer) stream() {
+	if l.logOpts.Follow {
+		err := l.waitForContainerRunning()
+		if err != nil {
+			l.errc <- errors.Wrapf(err, "failed to waiting for container [%v] in pod [%s/%s] to start running", l.container, l.ns, l.pod)
+			return
+		}
+	}
 	req := l.client.CoreV1().Pods(l.ns).GetLogs(l.pod, l.logOpts)
 	readCloser, err := req.Stream()
 	if err != nil {
--- a/pkg/client/logs_test.go
+++ b/pkg/client/logs_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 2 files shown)


johnSchnake:
Thanks for taking this on!

It seems like either I'm misreading your PR or you've misinterpreted the problem though: it isn't that we tail logs of a given pod and have a problem because it isn't in the proper status yet. The problem is that we grab a static list of pods at the start of the command and then tail just those pods for the rest of the command. So if I'm going to start the aggregator, then 3 plugins but start the logs right after the aggregator, that is the only set of logs I will ever get (I will never start getting the plugin logs because those pods weren't yet created when we got a list of pods).  We grab the static list of pods [here](https://github.com/heptio/sonobuoy/pull/706/files#diff-7a369ae51c001d42bbaee0e2b0efd386R113) and then range over it making log streamers for each pod.

We should do something to keep checking for new pods on a somewhat regular basis.

ashish-amarnath:
@johnSchnake  it is definitely the latter. Your comment clarifies the problem. I'll work on the correct fix :)

johnSchnake:
@ashish-amarnath You may be right; I'm not really sure if an error would be returned or if the API calls would handle that more gracefully. One way to test it would be to modify the code to try and hit that condition and just look at the errors you get.

It is a possibility that some of the times when I thought the pod didn't exist that it just wasn't ready and had the same behavior I suppose but I'd prefer to see the issue exist (even though a forced experiment) before trying to fix it.

I can try and make this happen and see if I hit the error.

johnSchnake:
Confirmed not only that this is an issue, but that it causes a hang in the current code.

I'll put up a PR to fix that hang. The issue is that we are writing an error to an unbuffered channel and no one is reading on the other end so we just block forever when trying to report the error (that the pod is still pending creation).

Once my little fix goes in you should be good to gracefully handle that. 👍 Thanks for bringing that up!

pkg/client/logs.go
```diff
@@ -180,8 +181,47 @@ type logStreamer struct {
...
+		for _, cs := range pod.Status.ContainerStatuses {
+			if cs.Name == l.container && cs.State.Running != nil {
+				return nil
+			}
+			if isContainerRunning(&pod.Status.ContainerStatuses, l.container) {
+				return nil
+			}
+		}
+
+		//TODO: add a time out
```
johnSchnake: I dont think we need a timeout since this only occurs when they want to follow the logs which doesn't implicily have any sort of ; if they are tailing the logs itself doesnt have a timeout either.

pkg/client/logs.go
```diff
@@ -180,8 +181,47 @@ type logStreamer struct {
...
+func (l *logStreamer) waitForContainerRunning() error {
+	backoffSeconds := 1 * time.Second
+	for {
+		pod, err := l.client.CoreV1().Pods(l.ns).Get(l.pod, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get pod [%s/%s]", l.ns, l.pod)
+		}
+
+		for _, cs := range pod.Status.ContainerStatuses {
+			if cs.Name == l.container && cs.State.Running != nil {
```
johnSchnake: It looks like you moved all the logic of this loop into the `isContainerRunning(...) method`; just need to call it instead of looping right?

ashish-amarnath: yeah.. that's right! good catch

johnSchnake:
Fix for the race has been merged; you can rebase and continue with this. You may want to see if there is a way to leverage those errors you'll now get (if that simplifies any of your logic here)

johnSchnake:
@ashish-amarnath checking on the status; do you still intend to revisit this?

pkg/client/logs.go
```diff
@@ -228,8 +229,43 @@ type logStreamer struct {
 	client             kubernetes.Interface
 }
 
+func isContainerRunning(statuses *[]v1.ContainerStatus, containerName string) bool {
+	for _, cs := range *statuses {
+
+		fmt.Printf("container %v, state= %v\n", containerName, cs.State)
```
johnSchnake: No need to have this print each time we check fi something is running.

pkg/client/logs.go
```diff
@@ -228,8 +229,43 @@ type logStreamer struct {
...
+		if err != nil {
+			return fmt.Errorf("failed to get pod [%s/%s]", l.ns, l.pod)
+		}
+
+		if isContainerRunning(&pod.Status.ContainerStatuses, l.container) {
+			return nil
+		}
+
+		time.Sleep(backoffSeconds)
+		backoffSeconds *= 2
```
johnSchnake: So we're doing this backoff with a 2N wait time but if the pod fails we will just always be waiting.

I suggested removing the fmt.Print line from the `isContainerRunning` check, but it seems reasonable to me to put one here so people know what is happening:

```
fmt.Printf("container %v, state= %v will attempt to stream logs in %v seconds\n", containerName, cs.State, backoffSeconds)
```

johnSchnake:
If you can make that change I'll test manually and we can merge today before the release if you have the time.

pkg/client/logs.go
ashish-amarnath: stale debug log line. removed :)

johnSchnake:
manual test of `sonobuoy run && sonobuoy logs -f` showed:
```
container sonobuoy/sonobuoy/kube-sonobuoy, is not running, will retry streaming logs in 1s seconds
container sonobuoy/sonobuoy/kube-sonobuoy, is not running, will retry streaming logs in 2s seconds
namespace="sonobuoy" pod="sonobuoy" container="kube-sonobuoy"
time="2019-09-13T20:02:02Z" level=info msg="Scanning plugins in ./plugins.d (pwd: /)"
time="2019-09-13T20:02:02Z" level=info msg="Scanning plugins in /etc/sonobuoy/plugins.d (pwd: /)"
time="2019-09-13T20:02:02Z" level=info msg="Directory (/etc/sonobuoy/plugins.d) does not exist"
```

👍

PR #707, opened 2019-05-06
[WIP] Logs from new pods

### LINKED ISSUE #703: Sonobuoy logs should start following new pods when they start

ashish-amarnath:
Stream logs from newly created pods

- Fixes #703 

**Release note**:
```
release-note
```

Trying to get some early feedback on the approach here.

### DIFF
--- a/pkg/client/logs.go
+++ b/pkg/client/logs.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -110,6 +111,7 @@ func (r *Reader) Read(p []byte) (int, error) {
 	return len(data), nil
 }
 
+<<<<<<< HEAD
 // getPodsForLogs retrieves the pods to stream logs from. If a plugin name has been provided, retrieve the pods with
 // only the plugin label matching that plugin name. If no pods are found, or no plugin has been specified, retrieve
 // all pods within the namespace.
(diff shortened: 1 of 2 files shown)


pkg/client/logs_test.go
```diff
@@ -201,3 +203,76 @@ func TestLogReaderNoError(t *testing.T) {
 		})
 	}
 }
+
+func getFakeKubernetesClientSet() kubernetes.Interface {
+	// TODO(ashish-amarnath): figure out how to fake this
```
johnSchnake: You and me both ;)

Many functions use the `kubernetes.Interface` because it is the top level client but in general it seems un-fakeable because it is such a large API. Instead you have to try to concentrate as much of your logical code into a testable set of functions which use a much smaller interface.

johnSchnake:
ping

@ashish-amarnath just checking in; is there anything you need on this from my end? Is this still something you plan on tackling?

pkg/client/logs_test.go
johnSchnake: I finally figured out how to do this and it was always apparently a solved problem. Client-go creates a fake interface object and you can create actions that respond to different API calls. We now have an example in the codebase to model tests that need this functionality: https://github.com/heptio/sonobuoy/blob/6d02613d2fcbf91f284918cf75c94c9d4e00df1c/pkg/plugin/aggregation/run_test.go#L131-L138

johnSchnake:
I just pinged the other logs PR of yours; let me know what you plan on doing with both.

pkg/client/logs.go
```diff
@@ -110,30 +116,46 @@ func (r *Reader) Read(p []byte) (int, error) {
...
-			return nil, errors.Wrap(err, "failed to list pods")
-		}
-
-		if len(pods.Items) != 0 {
-			return pods, nil
+func getPodsToStreamLogs(client kubernetes.Interface, cfg *LogConfig, stop chan bool, podListChan chan v1.PodList, errc chan error) {
+	for {
+		select {
+		default:
+			if cfg.Plugin != "" {
```
johnSchnake: I think an extra layer of abstraction would help clarify the flow here.

The existing `getPodsForLogs` method can remain as-is and you then have a new method which will do polling using that method, log errors, etc.

I think trying to push all the logic into this method makes it much harder to read/understand.

pkg/client/logs.go
```diff
@@ -150,50 +172,60 @@ func (s *SonobuoyClient) LogReader(cfg *LogConfig) (*Reader, error) {
...
-
-	// We must make sure the error channel has capacity enough so that it never blocks
-	// in case of early exits.
-	numContainers := 0
-	for _, pod := range pods.Items {
-		numContainers += len(pod.Spec.Containers)
-	}
-
-	errc := make(chan error, numContainers)
+	errc := make(chan error, 5)
```
johnSchnake: Why 5?

johnSchnake:
Going to put up a new variant of this PR which will try and simplify things. Thanks for your work on this @ashish-amarnath !

pkg/client/logs.go
ashish-amarnath: I was just trying something out. This shouldn't have made it into the commit.

PR #709, opened 2019-05-08
Set any plugin env vars from the CLI

### LINKED ISSUE #639: Generalize plugin configuration
Once #405 is finished we can easily tell the CLI to run a given plugin whose YAML is a local file.

The problem with that is that, just like the e2e plugin, most plugins will have some configuration which can be specified via env vars on the image. #405 simply pulls the YAML in but doesnt allow manipulation of it.

As brought up in the design doc for the former issue, it would be useful to have a way to specify env vars on an arbitrary plugin. E.g.
```
sonobuoy run --plugin custom --plugin-env custom.FOO=BAR
```

Design doc just being created here: https://docs.google.com/document/d/1X2cWDhBLNMUe8CH4LgfXA0N05vWxHdV3_zL-tIOukHg/edit?usp=sharing

johnSchnake:
Added a new flag and backing type in order to gather and
set env vars on any plugin from the command line which
makes it easier to iterate on plugins/tests.

This PR does the following:
 - adds new flag --plugin-env and backing type
 - adds new field on GenConfig in order to pass this to
pkg/client/gen/GenerateManifest(...) in order to apply
the changes after the dynamic plugins have been generated.
 - adds logic to sort the env var to ensure a stable order
for testing/use

On the command line the values are of the form:
--plugin-env pluginName.env=val

Fixes #639

**Release note**:
```
Added new flag --plugin-env allowing users to specify env var values on any plugin from the command line.
```

### DIFF
--- a/pkg/client/gen.go
+++ b/pkg/client/gen.go
@@ -119,6 +119,36 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
 		return nil, errors.Wrap(err, "plugin YAML generation")
 	}
 
+	if cfg.PluginEnvOverrides != nil {
+		for pluginName, envVars := range cfg.PluginEnvOverrides {
+			found := false
+			for _, p := range plugins {
+				if p.SonobuoyConfig.PluginName == pluginName {
+					found = true
+					newEnv := []corev1.EnvVar{}
+					removeVals := map[string]struct{}{}
+					for k, v := range envVars {
+						if v != "" {
+							newEnv = append(newEnv, corev1.EnvVar{Name: k, Value: v})
+						} else {
+							removeVals[k] = struct{}{}
+						}
+					}
+					p.Spec.Env = mergeEnv(newEnv, p.Spec.Env, removeVals)
+				}
+			}
+
+			// Require overrides to target existing plugins and provide a helpful message if there is a mismatch.
+			if !found {
+				pluginNames := []string{}
+				for _, p := range plugins {
+					pluginNames = append(pluginNames, p.SonobuoyConfig.PluginName)
+				}
+				return nil, fmt.Errorf("failed to override env vars for plugin %v, no plugin with that name found; have plugins: %v", pluginName, pluginNames)
+			}
+		}
+	}
+
 	pluginYAML := []string{}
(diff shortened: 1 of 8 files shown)


johnSchnake:
So the DD specified that if a plugin specified DNE then an error would be thrown; that is not yet implemented. I'll add that and a test.

I could update the DD to just go the other way on that; but I do think that can lead to problems where the user _thinks_ they're setting one value and they actually aren't. Let's help out the user in that case.

johnSchnake:
I can also add/modify some documentation about this. Otherwise it is g2g.

pkg/client/gen.go
```diff
@@ -119,6 +119,36 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
 		return nil, errors.Wrap(err, "plugin YAML generation")
 	}
 
+	if cfg.PluginEnvOverrides != nil {
+		for pluginName, envVars := range cfg.PluginEnvOverrides {
+			found := false
+			for _, p := range plugins {
+				if p.SonobuoyConfig.PluginName == pluginName {
```
stevesloka: Maybe throw an error if a plugin doesn't match?

johnSchnake: See just a few lines farther down: https://github.com/heptio/sonobuoy/pull/709/files#diff-bbcb11cee7bcaf5cb4cf6e2cd7dc61bbR147 Since the plugins are a slice instead of a map I can't just do a quick lookup; so I just track whether it was found in the list of plugins or not.

I try to give a really helpful message to make things like typos clear (so they can see, for instance, they typed `system-logs` instead of `systemd-logs`)

johnSchnake:
> Also, would be good to have some docs explaining how this works.

I mentioned that I think I need to put the docs into another PR altogether. This single feature (just a flag) doesn't necessitate a whole new, independent page IMO and so when trying to modify other pages the modifications started to cascade into things that weren't really related.

Are you against me just having a 'update docs' PR coming up after this or is that OK with you?

johnSchnake:
Only failing the new netlify checks which aren't indicative of anything other than that they are a WIP. Based on chat today; no other blockers on this. Merging.

PR #712, opened 2019-05-10
2019-05-10 Adding in Jekyll site

lspringer:
2019-05-10 Adding in Jekyll site

Adding:

netlify.toml
site subdirectory containing Jekyll site code


johnSchnake:
The documentation markdown pages was taken from the `gh-pages` branch but updates made in the past few releases have not been pushed to there.

Instead, all the docs are on master:
```
README.md
docs/*
```

Then since we are moving the docs down from `docs` to `site/docs` we'll need to make sure any links there are still working/updated as needed.

PR #717, opened 2019-05-13
Ensure conformance image version is prefixed with 'v'

### LINKED ISSUE #716: kube-conformance-image-version flag strips `v` leading to errors
Run `sonobuoy gen --kube-conformance-image-version=v1.14.2 | grep conform` and see that the image is from the heptio-images registry.

For v1.14.2 we should be using the upstream google containers.

The behavior of the `--kube-conformance-image-version` flag is incorrect. Originally I was going to show you how overriding the image version via that flag would lead to the upstream conformance image (instead of the Heptio one) but, to my surprise, it failed.

The issue there (I'll make another ticket for this) is that there a few paths to getting the image version:
 - the user specified 'auto' (the default) and we asked the server, resuling in something like `v1.14.2`
 - user explicity tells us the version; we even require them to state it with the `v` prefix. However, the logic to `Set` the value strips that `v` prefix because it goes through a semver library to ensure the value is reasonable.

**Environment:**

- Sonobuoy version: v0.14.2

johnSchnake:
When using flags to set the kube-conformance-image-version
the `Set` method ends up stripping the `v` prefix which it
requires the user to provide.

This causes a problem because the rest of the codebase assumes
the version will start with that prefix due to the fact that that
is how images are tagged (for Sonobuoy, Heptio kube-conformance,
and for upstream conformance images).

This change just adds the prefix back after validating the version
and updates the tests to actually check for the version that
gets persisted.

Fixes #716

A simple test:
```
$ ./sonobuoy gen --kube-conformance-image-version=v1.14.0|grep conform
$ ./sonobuoy gen --kube-conformance-image-version=v1.14.1|grep conform
```
Both of those should result in images with the tag prefixed with `v`. In addition, the first should be a heptio image, the latter a google image.

As it stands now, the `v` prefix is missing and both try to use the Heptio registry due to the missing prefix affecting the comparison.

**Release note**:
```
Fixes handling of the --kube-conformance-image-version flag so that the resulting conformance image tag is prefixed with 'v' as expected.
```

### DIFF
--- a/pkg/image/imageversion.go
+++ b/pkg/image/imageversion.go
@@ -59,7 +59,10 @@ func (c *ConformanceImageVersion) Set(str string) error {
 		if err != nil {
 			return err
 		}
-		*c = ConformanceImageVersion(version.String())
+
+		// It is important to set the string with the `v` prefix in order
+		// to be consistent with server version reporting and image tagging norms.
+		*c = ConformanceImageVersion(fmt.Sprintf("v%v", version.String()))
 	}
 
 	return nil
--- a/pkg/image/imageversion_test.go
+++ b/pkg/image/imageversion_test.go
@@ -33,16 +33,19 @@ func TestSetConformanceImageVersion(t *testing.T) {
 	tests := []struct {
 		name    string
 		version string
+		expect  string
 		error   bool
 	}{
 		{
 			name:    "version detect",
 			version: "auto",
+			expect:  "auto",
 			error:   false,
 		},
 		{
 			name:    "use latest",
 			version: "latest",
+			expect:  "latest",
 			error:   false,
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/image/imageversion.go
```diff
@@ -59,7 +59,10 @@ func (c *ConformanceImageVersion) Set(str string) error {
 		if err != nil {
 			return err
 		}
-		*c = ConformanceImageVersion(version.String())
+
+		// It is important to set the string with the `v` prefix in order
+		// to be consistent with server version reporting and image tagging norms.
+		*c = ConformanceImageVersion(fmt.Sprintf("v%v", version.String()))
```
stevesloka: Should `ConformanceImageVersion` return the properly formatted string?

johnSchnake: `ConformanceImageVersion` isn't a function, it is just casting the string to the type.

If you meant that "should the `String() string` method on the `ConformanceImageVersion` type do this formatting, I would argue no.  We have a constrained type so if we want validation like this we have to either use custom `Get()` methods (more overhead than we need for a flag wrapper) or just do the proper validation on `Set`.

To put all this logic in the `String` method would be a bit awkward because we wouldn't be able to handle the error case so we'd just end up having to put the logic in both places. And once we fix it here we don't really have a problem which would make it worth copying the logic to `String` as well.

PR #719, opened 2019-05-14
Fix race/deadlock issues with logs when a pod is not ready

### LINKED ISSUE #708: Fix race in logs gathering
`kubectl apply -f test.yaml && sonobuoy logs`

will hang every time because the pod isn't yet in the created phase leading to an error when creating the log streamer. The problem is that the error goes onto an unbuffered channel and needs to be read before the method returns. However, other logic exists that says to keep reading until the channel of messages is closed before grabbing the error from the error channel.

This results in deadlock since the closing of the channels doesnt happen until the `stream()` method is done, which is trying to put the error on the channel.

`sonobuoy logs` should just report the error and exit.

In the kubectl example above, the test yaml can be any pod really; dealers choice. Using kubectl run is different though since it is waiting for certain events to be complete rather than just firing off the creation requests.

Sonobuoy version 0.14.2

johnSchnake:
There existed an issue where if there was a problem reading the logs
of a pod it would try to put a message onto the error channel of
the logReader which was unbuffered. As a result it would block but
the problem was that the reading routine was trying to drain the
message channel first before reading errors.

The solution to this was to ensure that we sufficiently buffered the
error channel to handle these bad-reader cases. Trying to stream
from both the channels introduced issues where the messages wouldn't
be drained yet but the error would short circuit the logic causing
the logs to not get fully read.

Fixes #708

Note on the solution: I am generally highly skeptical when the solution to a race is "increase the buffer size of the channel" but after thinking on it a bit I don't think there is really a better way to do this. I did some googling and found this minimum buffer size for errors was not an anomaly. The only alternative seemed more awkward which would involve creating other channels and layers of logic which made it more confusing. This seemed the best approach: sure to remove any race, minimally changes the existing logic, still prioritizes the messages before reading the errors, etc.

w.r.t. testing the change:
A bit difficult to test but you can do so manually; I used 2 simple YAML files and checked a few cases:
 - can read normal logs in a standard, ready state
 - can properly report the error about the container not being ready
 - can report the error about one pod not being ready while at the same time reporting logs of other ready pods in the namespace.

The command I used to check the initial error reporting was:
```
kubectl apply -f test1.yaml && ./sonobuoy logs -n default
```

That should hit a race with the pod and report an error (currently it would hang forever).

Then try to just grab the logs a few more times with `./sonobuoy logs -n default` and see that the pod starts normally and you eventually get the logs as expected.

Then I copied that yaml and changed the pod name in order to get 1 pod that was ready and then race the other so I could see I got the both the logs _and_ the error:
```
kubectl apply -f test2.yaml && ./sonobuoy logs -n default
```

For the yaml file I used:
```
apiVersion: v1
kind: Pod
metadata:
  name: testpod2
spec:
  containers:
  - name: sise
    image: mhausenblas/simpleservice:0.5.0
    ports:
    - containerPort: 9876
```

**Release note**:
```
Fixed a bug which caused `sonobuoy logs` to hang if a pod was not in a ready state.
```

### DIFF
--- a/pkg/client/logs.go
+++ b/pkg/client/logs.go
@@ -85,8 +85,14 @@ func (r *Reader) Read(p []byte) (int, error) {
 	data, ok := <-r.bytestream
 	// If the bytestream is done then save the error for future calls to Read.
 	if !ok {
-		r.err = <-r.errc
-		return 0, r.err
+		r.err, ok = <-r.errc
+		// Assume EOF if error channel is closed.
+		if r.err == nil && !ok {
+			r.err = io.EOF
+		}
+		if r.err != nil {
+			return 0, r.err
+		}
 	}
 
 	// TODO(chuckha) this code and the code above in the overflow buffer is identical. Might be an indication of a cleaner way to do this.
@@ -114,7 +120,14 @@ func (s *SonobuoyClient) LogReader(cfg *LogConfig) (*Reader, error) {
 		return nil, errors.Wrap(err, "failed to list pods")
 	}
 
-	errc := make(chan error)
+	// We must make sure the error channel has capacity enough so that it never blocks
+	// in case of early exits.
+	numContainers := 0
+	for _, pod := range pods.Items {
+		numContainers += len(pod.Spec.Containers)
+	}
+
+	errc := make(chan error, numContainers)
 	agg := make(chan *message)
 	var wg sync.WaitGroup
 
@@ -146,7 +159,6 @@ func (s *SonobuoyClient) LogReader(cfg *LogConfig) (*Reader, error) {
 	go func(wg *sync.WaitGroup, agg chan *message, errc chan error) {
 		wg.Wait()
 		close(agg)
-		errc <- io.EOF
 		close(errc)
 	}(&wg, agg, errc)
 


pkg/client/logs.go
```diff
@@ -146,7 +157,6 @@ func (s *SonobuoyClient) LogReader(cfg *LogConfig) (*Reader, error) {
 	go func(wg *sync.WaitGroup, agg chan *message, errc chan error) {
 		wg.Wait()
 		close(agg)
-		errc <- io.EOF
```
johnSchnake: No need to do this if we add the 'assume eof' logic above. Otherwise I think we have to worry about another odd edge case: what if every container threw an error, filling the buffered channel, then we tried to add an eof here?

I _think_ that would work OK since the _this_ call may block but eventually the `Read()` call would pull a value off, unblocking this, and closing the error channel. Regardless of if the error channel closed, the program would end rather than hang. Or so I think w/o checking.

It seems the more simple thing is to just close the channels and let the logic happen at the read site where assuming EOF would make sense once both the message and error channels have been closed.

johnSchnake:
@ashish-amarnath Here is the fix I finally ended at.

`select`ing off of the 2 channels led to problems because the messages weren't prioritized over the errors like they were before causing a race between normal logs and the standard EOF condition (or another error like the pods not being ready).

I talked to you about not wanting to increase the buffer size on a whim, but after a lot of deliberation, I think that is actually the right design here considering the rest of the code which already exists.

PR #723, opened 2019-05-21
Add ability to specify custom annotations

### LINKED ISSUE #567: Allow sonobuoy plugins to have custom annotations
I would like to be able to specify custom annotations for sonobuoy plugin pods

Particularly I am creating a sonobuoy plugin based on powerfulseal, and in that vein I need to add kiam annotations to the powerfulseal pod. Instead of doing a 2-stage launch (launching powerfulseal pod /w annotation from the plugin pod) it would be far easier to specify the annotation into the plugin config yaml.

**Environment:**

- Sonobuoy version: 0.11.6
- Kubernetes version: (use `kubectl version`): 1.10
- Kubernetes installer & version: bootkube
- Cloud provider or hardware configuration: AWS
- OS (e.g. from `/etc/os-release`): CoreOS

johnSchnake:
In some cases it may be necessary to add custom annotations
to the sonobuoy pods (aggregator and plugins). Most notably,
this helps facilitate kiam permissions.

This change adds the fields to the necessary types so that it
can be passed around appropriatly and fill in the different
templates where it should be used.

Fixes #567

To review the changes, you mainly just need to look at the modifications to the tests/templates I made.

They are not exactly the most idiomatic tests; they hard-code a single test case and then check each part of the result manually. I'd like to improve those sorts of tests but this PR is not the place for it; instead, I just opted to focus on the feature and leave the refactoring for another day (to be clear, I just extended existing tests).

It would be nice to have a way to _really_ generate the plugin yaml from the command line in a realistic way. There is currently a command that does that for dev purposes like this, but it hardcodes some things you may want to test and it doesn't follow the same path as `sonobuoy master` which really loads the plugins/configs. I may work on trying to get those in line so that testing things like this is not so difficult.

**Release note**:
```
Added a "CustomAnnotations" field to the configuration options so that you can annotate all the Sonobuoy pods/plugins with custom values. The values are added to any set Sonobuoy would already be using, and it is left to the user to ensure they do not collide.
```

### DIFF
--- a/cmd/sonobuoy/app/gen_plugin.go
+++ b/cmd/sonobuoy/app/gen_plugin.go
@@ -40,9 +40,10 @@ const (
 
 // GenPluginConfig are the input options for running
 type GenPluginConfig struct {
-	Paths            []string
-	PluginName       string
-	ImagePullSecrets string
+	Paths             []string
+	PluginName        string
+	ImagePullSecrets  string
+	CustomAnnotations map[string]string
 }
 
 var genPluginOpts GenPluginConfig
@@ -86,6 +87,7 @@ func generatePluginManifest(cfg GenPluginConfig) ([]byte, error) {
 		config.DefaultImage,
 		"Always",
 		cfg.ImagePullSecrets,
+		cfg.CustomAnnotations,
 		cfg.Paths,
 		[]plugin.Selection{{Name: cfg.PluginName}},
 	)
--- a/pkg/client/gen.go
+++ b/pkg/client/gen.go
@@ -46,14 +46,15 @@ const (
 type templateValues struct {
 	Plugins []string
 
-	SonobuoyConfig   string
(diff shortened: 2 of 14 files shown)


johnSchnake:
To confirm this I did the following in addition to the testing changes:
```
make clean
REGISTRY=schnake make && docker push schnake/sonobuoy:customAnnotations
go install
sonobuoy gen config|jq '.CustomAnnotations={"foo":"bar"}'|jq '.WorkerImage="schnake/sonobuoy:customAnnotations"'>config.json
sonobuoy run --config=config.json
kubectl get pods -n heptio-sonobuoy -o yaml | grep foo --context=5
```
That is, I built it and pushed the image to my own registry and started Sonobuoy with a config that specified the foo/bar annotation. I then confirmed the annotations were on the pods.

Each pod/ds was updated as expected.

PR #725, opened 2019-05-22
Fix/simplify the config loading logic enabling empty resources

### LINKED ISSUE #721: Providing a config with empty, non-nil resources is not respected
```
sonobuoy gen config | jq '.Resources=[]' > noquery.conf
sonobuoy gen --config=./noquery.conf| rg Resources
```

The list of resources is set as the default list (numerous items in it).

The resulting resources list should be empty as it was explicitly stated in the config file I provided.

The problem is that the `mergo.Merge` considers a slice with 0 length to satisfy `isEmptyValue` and doesnt differentiate between nil vs empty which is common in Go.

This is also related to a flow where you may want to run 0 plugins by providing an empty slice for `Plugins`. The problem there is _also_ that that field was recently made `omitempty` so after serializing the data to the server, the server can't recover the difference between nil and empty slices. That should also be fixed.

johnSchnake:
The method to load the config was utilizing viper which is a
common library used alongside cobra CLI tools. However, we were
using very little of its functionality and some of its logic
was bypassing the intent of our data structures: most clearly,
the empty but non-nil resources were still being treated as nil
meaning that you would query all resources instead of none of them.

Summary:
 - Sending a config with empty (but not nil) resources will now
actually lead to no query being done (other than Sonobuoy
gathering its own pod logs at the current time)
 - Removes viper from the main config handling in order to simplify
the logic and clarify the code flow. Uses basic file/json handling
to unmarshal the data now.
 - Continues to load the config from the pwd or default location
with an environment override.
 - Adds a test to ensure `LoadConfig` now properly treats empty
values for the Resources.

Fixes #721

Run this locally to see it work in full:
```
$ sonobuoy gen config|jq '.Resources=[]'|jq '.WorkerImage="schnake/sonobuoy:masterConfigLoading"'>noquery.conf
$ sonobuoy run --mode=quick --config=noquery.conf --wait
$ outfile=$(sonobuoy retrieve) && mkdir results && tar -xf $outfile -C results
```
Then open `results` and confirm no data was queried. I'll note that this _does_ still result in directories being created for query results even though no data was dumped there. I think that can be a followup, small, PR.

**Release note**:
```
Fixes issue where empty resource slice was sent to the aggregator server but all resources were still queried.
```

### DIFF
--- a/pkg/config/loader_test.go
+++ b/pkg/config/loader_test.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"reflect"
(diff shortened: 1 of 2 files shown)


pkg/config/loader_test.go
```diff
@@ -71,6 +71,9 @@ func TestDefaultResources(t *testing.T) {
 	if len(cfg.Resources) != 0 {
 		t.Error("Default resources should not be applied if specified in config")
 	}
+	if cfg.Resources == nil {
+		t.Error("Empty resources should not be converted to nil")
+	}
```
johnSchnake: This is the new test that ensures `sonobuoy master` will really load empty, non-nil resources.

pkg/config/loader_test.go
```diff
@@ -86,7 +89,7 @@ func TestDefaultResources(t *testing.T) {
 	}
 
 	// Check that specifying one resource results in one resource
-	blob = `{"Resources": "Pods"}`
+	blob = `{"Resources": ["Pods"]}`
```
johnSchnake: Changing an old test but I feel pretty strongly on this. It seems to me this behavior would have been added just because this is how `viper` handled it. I see no reason to be be flexible on the data type. The `Resources` is a list, so the value should always be a list or fail parsing.

Issue #728, opened 2019-05-24
Aggregator doesnt properly handle errors when getting results


johnSchnake:
**What steps did you take and what happened:**
Tests were run in a situation with poor network behavior and the logs have lots of "connection reset by peer" lines.

The worker logic uses a client that uses retries by default

The aggregator will accept the first request and then blacklist any more attempts because it has "seen" those results. So what is happening is part of the data is being sent, the connection gets reset, the client tries again but gets turned away.

**What did you expect to happen:**
The server should accept new results in cases where errors were encountered.

**Anything else you would like to add:**
The logic for marking the results as "seen" even if we get errors makes sense (as the code comments indicate) because we don't want the server to disregard error'd results as it may make the server hang if the client doesn't retry again.

However, it seems like the logic needs to be tweaked so that we either:
 - always allow the resending of data and take the newest values
 - track which results we got with/without error and allow retries on the former

**Environment:**

- Sonobuoy version: v0.14.2

johnSchnake:
Thanks to @rbankston for reporting this bug and providing the necessary logs to diagnose it. (https://github.com/kubernetes/kubernetes/issues/74839 may have caused the original connection reset issue but it is not apparently a final, perfect fix and may run into other issues)

johnSchnake:
Added this to the current milestone as p0. We have the bandwidth to support and I consider this a pretty significant bug since (1) it is confusing to users (2) it is not easy to debug if you are not very familiar with the system (3) it may be blocking to users: if this happens regularly for some reason it can keep users from ever being able to get results from their plugins.

Working on this currently.

PR #729, opened 2019-05-24
Unify http/non-http result handling

johnSchnake:
We have two flows for this logic: one for the
HTTP server and one for the channel monitoring
results.

This change places all the logic into a central
method so that it can be modified/tested more
confidently/succinctly.

Related to #728 

No logical changes in this PR; only doing this in prep of my next change to simplify the diffs. This way we can more easily test the logic used throughout.

**Release note**:
```
NONE
```

### DIFF
--- a/pkg/plugin/aggregation/aggregator.go
+++ b/pkg/plugin/aggregation/aggregator.go
@@ -53,9 +53,31 @@ type Aggregator struct {
 	// resultEvents is a channel that is written to when results are seen
 	// by the server, so we can block until we're done.
 	resultEvents chan *plugin.Result
+
 	// resultsMutex prevents race conditions if two identical results
 	// come in at the same time.
 	resultsMutex sync.Mutex
+}
+
+// httpError is an internal error type which allows us to unify result processing
+// across http and non-http flows.
+type httpError struct {
+	err  error
+	code int
+}
+
+// HttpCode returns the http code associated with the error or an InternalServerError
+// if none is set.
+func (e *httpError) HttpCode() int {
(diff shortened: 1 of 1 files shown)


johnSchnake:
@stevesloka Sanity check on this?

The diff looks bigger than it logically is.

I did the following:
 - created the error type to pass http-ish errors
 - moved the logic out of both `IngestResults` and `HandleResults` and put it into the `processResults` method
 - the 2 original methods call process results and just log/respond to the user as necessary in that context

No tests changed and no logical change in behavior intended. Follow up PR will modify things and add new tests.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -110,47 +132,66 @@ func (a *Aggregator) isResultDuplicate(result *plugin.Result) bool {
...
 	if !a.isResultExpected(result) {
-		http.Error(
-			w,
-			fmt.Sprintf("Result %v unexpected", resultID),
-			http.StatusForbidden,
-		)
-		return
+		return &httpError{
+			err:  fmt.Errorf("result %v unexpected", resultID),
+			code: http.StatusForbidden,
```
stevesloka: A `403` seems out of place here for this type of error.

johnSchnake: Agree, looking at docs that seems like it is supposed to be auth related, not content like this. Not 100% sure of the right response though but that is for another PR; this just moved the code and didnt chose it here. Maybe https://httpstatuses.com/422 would be most appropriate.

PR #730, opened 2019-05-25
Retry results

### LINKED ISSUE #728: Aggregator doesnt properly handle errors when getting results

johnSchnake:
Will be rebased/tweaked after #729 merges which simplifies this update/testing since it puts all the logic in one place.

In the case where a connection gets reset or other
processing errors occur, the client is already
coded to automatically retry submission of the
results.

However, the server was marking the results as 'seen'
and would not reprocess them.

This change allows reprocessing results only if there
were errors processing before.

Fixes #728

FWIW, with the centralized code I think some other tests _could_ be fixed up and made to read a bit better but it's not really worth it to go and rework older tests.

**Release note**:
```
Fixed a bug which led to partial results for plugins when there were connection resets and other networking issues.
```

### DIFF
--- a/pkg/plugin/aggregation/aggregator.go
+++ b/pkg/plugin/aggregation/aggregator.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"path"
 	"sync"
+	"time"
 
 	"github.com/heptio/sonobuoy/pkg/plugin"
 	"github.com/heptio/sonobuoy/pkg/tarball"
@@ -35,7 +36,8 @@ import (
 )
 
 const (
-	gzipMimeType = "application/gzip"
+	gzipMimeType       = "application/gzip"
+	defaultRetryWindow = 15 * time.Second
 )
 
 // Aggregator is responsible for taking results from an HTTP server (configured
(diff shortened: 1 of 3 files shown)


pkg/plugin/aggregation/aggregator.go
```diff
@@ -106,6 +124,13 @@ func (a *Aggregator) Wait(stop chan bool) {
 			return
 		}
 	}
+
+	// Give all clients a chance to retry failed requests.
+	for _, failedTime := range a.FailedResults {
+		if time.Now().Sub(failedTime) < a.retryWindow {
+			time.Sleep(failedTime.Add(a.retryWindow).Sub(time.Now()))
```
stevesloka: Maybe refactor this out, the failedTime.Add.Sub.etc.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -106,6 +124,13 @@ func (a *Aggregator) Wait(stop chan bool) {
 			return
 		}
 	}
+
+	// Give all clients a chance to retry failed requests.
+	for _, failedTime := range a.FailedResults {
+		if time.Now().Sub(failedTime) < a.retryWindow {
```
stevesloka: maybe refactor this out as well, something like `shouldWaitForRetryWindow` or something?

PR #734, opened 2019-05-31
Add checks for if a container has terminated

### LINKED ISSUE #667: Sonobuoy should monitor plugin containers better
If a plugin container exits without writing the `done` file, we should detect it and report results to the best of our ability and/or report the error to the user clearly.

Currently, if a plugin exits without writing the done file, the Sonobuoy worker hangs forever waiting for it until the timeout.

There already exists some plugin monitoring logic, so it may just need to be patched.

johnSchnake:
In some cases, a plugin's main container may terminate without
reporting results. If this is the case, the sonobuoy worker
hangs forever (until the aggregator times out) because it thinks
the plugin is still running.

This change modifies the logic of the aggregator/plugin monitoring
logic in order to not only check for restarts/failure to schedule,
but also for if containers terminate without submitting results.

To ensure that we do not race with the normal plugin terminate and
result reporting, we wait 1 minute after the pod terminates to submit
the error result. In the normal flow, this would give plenty of time
for the plugin to send its results the server in which case this error
result would be tossed out.

Fixes #667

I added a test to cover the new functionality I added, but I did not add tests for the existing functionality. Those tests can be added later.

**Release note**:
```
Fixed a bug where Sonobuoy wouldn't realize a plugin container failed and would wait until the timeout to report it. Sonobuoy will now recognize that as an error and report it automatically.
```

### DIFF
--- a/pkg/plugin/aggregation/run.go
+++ b/pkg/plugin/aggregation/run.go
@@ -27,8 +27,10 @@ import (
 	"github.com/heptio/sonobuoy/pkg/backplane/ca"
 	"github.com/heptio/sonobuoy/pkg/plugin"
 	"github.com/heptio/sonobuoy/pkg/plugin/driver/utils"
+	sonotime "github.com/heptio/sonobuoy/pkg/time"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
@@ -85,7 +87,6 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 	// 1. Await results from each plugin
 	aggr := NewAggregator(outdir+"/plugins", expectedResults)
 	doneAggr := make(chan bool, 1)
-	monitorCh := make(chan *plugin.Result, len(expectedResults))
 	stopWaitCh := make(chan bool, 1)
 
 	go func() {
@@ -163,17 +164,8 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 
 	for _, p := range plugins {
 		logrus.WithField("plugin", p.GetName()).Info("Running plugin")
-		if err = p.Run(client, cfg.AdvertiseAddress, certs[p.GetName()]); err != nil {
-			err = errors.Wrapf(err, "error running plugin %v", p.GetName())
-			logrus.Error(err)
-			monitorCh <- utils.MakeErrorResult(p.GetResultType(), map[string]interface{}{"error": err.Error()}, "")
-			continue
-		}
-		// Have the plugin monitor for errors
-		go p.Monitor(client, nodes.Items, monitorCh)
+		go aggr.RunAndMonitorPlugin(ctx, p, client, nodes.Items, cfg.AdvertiseAddress, certs[p.GetName()])
 	}
-	// 5. Have the aggregator plumb results from each plugins' monitor function
-	go aggr.IngestResults(monitorCh)
 
 	// Give the plugins a chance to cleanup before a hard timeout occurs
 	shutdownPlugins := time.After(time.Duration(cfg.TimeoutSeconds-plugin.GracefulShutdownPeriod) * time.Second)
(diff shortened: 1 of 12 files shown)


pkg/plugin/driver/utils/utils.go
```diff
@@ -63,6 +64,20 @@ func IsPodFailing(pod *v1.Pod) (bool, string) {
 				return true, errstr
 			}
 		}
+
+		// Container terminated without reporting results.
+		if cstatus.State.Terminated != nil {
+			// Ensure we give some time to process the results.
+			if time.Now().Sub(cstatus.State.Terminated.FinishedAt.Time) > 1*time.Minute {
```
johnSchnake: Should move this to a const at the top for clarity.

pkg/plugin/driver/utils/utils.go
```diff
@@ -63,6 +68,20 @@ func IsPodFailing(pod *v1.Pod) (bool, string) {
 				return true, errstr
 			}
 		}
+
+		// Container terminated without reporting results.
+		if cstatus.State.Terminated != nil {
```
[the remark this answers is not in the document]
johnSchnake: No, so this is running in a go routine originated by `go p.Monitor(aChannel)` after starting the plugin `p`.

So the idea is that the plugin should get results and stop monitoring.

However, you're question did make me realize a flow in halting logic: it currently only halts if it is "cleaned up" but (1) I want to make cleanup logic optional (there is a issue for something like this already I think) and (2) the server is only cleaning up the plugins after they _all_ finish.

So if the plugins finish at very different times, pluginA (who finished and reported results) will keep monitoring and then hit this logic eventually.

So what I need to do is either modify the `stop monitoring` flow or add some extra information to this method so that we can check if results are reported yet.

Good catch! 💯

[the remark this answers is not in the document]
johnSchnake: So I continued to look at this because something seemed like it was eluding me and now I remember why I considered this safe at the time of writing.

The reason is that _this_ method will report the pod failing, but the action the monitor method takes is not to blow up, but to respond as if there were an error. However, the existing logic will throw out that result if there are already results.

However, that is still not ideal because this will _keep happening_ as long as monitoring is running, so for the reasons above we need to shut it off. So it wont break things, but it would not be a desirable flow and would fill the logs with confusing 'duplicate result' messages.

johnSchnake:
WIP at the moment; tests are passing but I haven't even done a live run of the branch yet. Some more testing to do for this. Wanted to get it visible though and maybe talk it over in case there are big issues with this approach.

johnSchnake:
Confirmed that codecov just isnt updated with my latest force push. Maybe it doesnt do well when you rebase?

Anyways, the updated report can be found here: https://codecov.io/github/heptio/sonobuoy/commit/8469a255b6e2ff91bb9c8a895d0867b3b682e387

It shows ~82% coverage and an increase for the entire codebase of .65%

pkg/plugin/aggregation/run_test.go
```diff
@@ -0,0 +1,198 @@
...
+			Namespace: "testNS",
+		},
+	}
+	testPluginExpectedResults := []plugin.ExpectedResult{
+		{ResultType: "myPlugin"},
+	}
+	healthyPod := corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"sonobuoy-run": ""}},
+	}
+	failingPod := corev1.Pod{
```
zubron: Idea for future: Now that these have been introduced in a number of tests, it would be nice to have some utilities to help with creating these healthy and failing pod objects.

pkg/plugin/aggregation/run_test.go
```diff
@@ -0,0 +1,198 @@
...
+				doneCh <- struct{}{}
+			}()
+
+			if tc.forceResults {
+				a.resultsMutex.Lock()
+				a.Results["myPlugin"] = &plugin.Result{}
+				a.resultsMutex.Unlock()
+			}
+
+			if tc.cancelContext {
```
zubron: Is it necessary to cancel this context again?

pkg/plugin/aggregation/run_test.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil, errors.Wrap(err, "couldn't create certificate")
+	}
+
+	return &tls.Certificate{
+		Certificate: [][]byte{certDER},
+		PrivateKey:  privKey,
+	}, nil
+}
+
+// get agg & check results afterwards; also check logs for duplicate stuff
```
zubron: These comments about which test cases to cover can be removed.

pkg/plugin/aggregation/run.go
```diff
@@ -209,3 +203,57 @@ func Cleanup(client kubernetes.Interface, plugins []plugin.Interface) {
...
+
+	// Control loop; check regularly if we have results or not for this plugin. If results are in,
+	// then stop the go routines monitoring the plugin. If the parent context is cancelled, stop monitoring.
+	for {
+		select {
+		case <-ctx.Done():
+			cancel()
+			return
+		case <-after(10 * time.Second):
+		default:
```
zubron: I think having the `default` case in here will result in the "sleep" not taking place. On each iteration, the default case will be satisfied and it will fall through to performing the operations after the `select`. This applies to all the other changes in sleep logic below.

pkg/plugin/aggregation/run.go
```diff
@@ -40,6 +41,9 @@ const (
 	jitterFactor         = 1.2
 )
 
+// Variable for swapping during tests.
+var after = time.After
```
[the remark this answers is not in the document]
johnSchnake: In this particular case yes, I went with a more generic approach in case I have another case where I want to do something like channel shenanigans so that I can block/guarantee where I am in the flow or track how many times I hit the sleep with a counter.

pkg/plugin/aggregation/run_test.go
johnSchnake: Yea, its pretty verbose and could be minimized a bit between the daemonset, job, and now aggregation tests.

pkg/plugin/aggregation/run.go
johnSchnake: Yes thank you for catching this!

johnSchnake: After removing that I did find that my test was failing because of how each package was using their own, unexported, `after` method. So that meant when doing this `RunAndMonitor` the `job.Monitor` logic was using its own method which couldn't be edited because it was unexported. As a result _it_ was sleeping for 10s before polling on the status.

The solution I went with was to centralize the `after` into a little `pkg/time` package and added another `pkg/time/timetest` (ala httptest from standard repo) to provide some helpful functions to shorten/ignore/reset the method.

Tests back to passing.

pkg/plugin/aggregation/run_test.go
johnSchnake: Thanks for seeing that I had done it twice.

PR #737, opened 2019-06-05
Adjust README description of support matrix

### LINKED ISSUE #736: README hardcodes supported versions but doesn't explain well
Maintainability nit/issue. I noticed the readme doesnt explain the versioning/support scheme well or how to opt-out and instead hardcodes the versions we support (which is outdated).

We should update it to be more correct and maintainable.

johnSchnake:
The versions were hardcoded, meaning it would have to be updated
on every release. Instead I explained the versioning/support
scheme and how to opt-out of it if desired. This will not have
to change unless we change those larger schemes.

Fixes #736

I want to make sure that the website properly updates when we update the readme. It is a bit unique since it is the only page _not_ in the docs/site that needs to be a part of the website.

**Release note**:
```
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,9 @@ for the following use cases:
 * Workload debugging
 * Custom data collection via extensible plugins
 
-Sonobuoy supports Kubernetes versions 1.11, 1.12 and 1.13.
+Sonobuoy supports 3 Kubernetes minor versions: the current release and 2 minor versions before. Sonobuoy is currently versioned to track the Kubernetes minor version to clarify the support matrix. For example, Sonobuoy v0.14.x would support Kubernetes 1.14.x, 1.13.x, and 1.12.x.
+
+> Note: You can skip this version enforcement by running Sonobuoy with the `--skip-preflight` flag.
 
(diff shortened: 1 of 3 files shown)


zubron:
Sorry, I just re-read your PR description. It's not clear to me what the desired behaviour is for changes to this file. Is there something in place to pull updates from this file so that they go on the website?

johnSchnake:
We're on the same page when it comes to this topic; I wasn't sure if the netlify stuff would automatically pull the main readme or not. I wanted to see what the preview showed and, as you said, this doesn't seem to work that way.

I'll find/update the other readme too and make sure they are consistent.

johnSchnake:
Preview site has the change now; it is a bit confusing because, as of now, we dont have versioned docs but the site/docs does have a folder for `latest` and another for `master`. I made a ticket about us versioning our docs so that users of older versions don't get confused about new features they dont have and such. #738 but that is for another day. For now I just copy/pasted the readme contents into each location. 

The only thing that isn't identical are the links to the sonobuoy logo which are relative paths so they are slightly different from the main README

PR #740, opened 2019-06-07
Add daemonSet monitoring tests

johnSchnake:
 - Extracts logic from daemonset.Monitor to another function
that doesnt involve polling for easier testing
 - Use fake client to mock different cluster situations
 - Add tests for each return path
 - Touch up a job.Monitor test for clarity/consistency

Related to #651 
Related to #667 (expanding testing of those methods so that we can test the higher level fix in the next PR more easily)

**Release note**:
```
NONE
```

### DIFF
--- a/pkg/plugin/driver/daemonset/daemonset_test.go
+++ b/pkg/plugin/driver/daemonset/daemonset_test.go
@@ -19,17 +19,22 @@ package daemonset
 import (
 	"crypto/sha1"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"testing"
 
 	"github.com/heptio/sonobuoy/pkg/backplane/ca"
 	"github.com/heptio/sonobuoy/pkg/plugin"
+	"github.com/heptio/sonobuoy/pkg/plugin/driver"
 	"github.com/heptio/sonobuoy/pkg/plugin/manifest"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kuberuntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/kubernetes/scheme"
+	k8stesting "k8s.io/client-go/testing"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/plugin/driver/daemonset/daemonset_test.go
```diff
@@ -166,3 +171,153 @@ func TestFillTemplate(t *testing.T) {
...
+				"No pod was scheduled on node node1 within 2562047h47m16.854775807s. Check tolerations for plugin myPlugin",
+				"No pod was scheduled on node node3 within 2562047h47m16.854775807s. Check tolerations for plugin myPlugin",
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			fclient := fake.NewSimpleClientset()
+			fclient.PrependReactor("list", "daemonsets", func(action k8stesting.Action) (handled bool, ret kuberuntime.Object, err error) {
+				t.Log("in ds handler")
```
johnSchnake: debug line

pkg/plugin/driver/daemonset/daemonset_test.go
```diff
@@ -166,3 +171,153 @@ func TestFillTemplate(t *testing.T) {
...
+			podsOnServer:            &corev1.PodList{},
+			errFromServerForPodList: errors.New("pod lookup err"),
+		}, {
+			desc:         "Missing pods results in errors for each",
+			nodes:        default3Nodes,
+			dsPlugin:     testPlugin,
+			dsOnServer:   &appsv1.DaemonSetList{Items: []appsv1.DaemonSet{validDS}},
+			podsOnServer: &corev1.PodList{},
+			expectErrResultMsgs: []string{
+				"No pod was scheduled on node node1 within 2562047h47m16.854775807s. Check tolerations for plugin myPlugin",
```
zubron: I'm guessing we have this specific time value because we don't have a creation timestamp for the DaemonSet and this is the string representation of the maximum duration. Perhaps we could add a comment for this, or check only a portion of the error message. e.g. check that it contains `No pod was scheduled on node node1 within`

pkg/plugin/driver/daemonset/daemonset_test.go
```diff
@@ -166,3 +171,153 @@ func TestFillTemplate(t *testing.T) {
...
+			Spec:       corev1.PodSpec{NodeName: node},
+		}
+	}
+	failingPod := func(node string) corev1.Pod {
+		return corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"sonobuoy-run": ""}},
+			Spec:       corev1.PodSpec{NodeName: node},
+			Status: corev1.PodStatus{
+				Conditions: []corev1.PodCondition{
+					{Reason: "Unschedulable", Message: "conditionMsg"},
```
zubron: It might make the error message checks below a bit clear/explicit if we could include the node name in the condition message because then we know which error message corresponds to which node. Just a suggestion though.

johnSchnake: Can you make an issue for that and we can make that change separately? It is rare to see that error condition at all but I agree with you it could be clarified.

pkg/plugin/driver/daemonset/daemonset_test.go
johnSchnake: Checking substrings IMO is slightly awkward and this also makes it clear in the tests what the whole message we surface to the user includes (which is what informed your idea in your other comment to add info to the error message).

If this is just a nit I'm inclined to just leave this as is since it is just a minor testing quirk which I don't think would confuse a dev too bad.

zubron: Fair enough. I agree that it's not very confusing, but I did have to think about why this expected message included the duration value that it did. I think I just tend to be overly cautious of tests related to time calculations :) In this case, I wanted to be sure that the test wouldn't incorrectly fail at some point in the future because of a different duration being included in the message. I can include adding a comment here as part of the follow up issue that I'll make for the other tests.

johnSchnake:
Fixed the important changes, some nits remain w.r.t. clarity of error messages which we can get back to later. Merging.

PR #755, opened 2019-06-18
Ensure dep has been run in CI

### LINKED ISSUE #244: CI check that `dep ensure` has been run
Right now you can submit a pull request that removes a dependency but without removing the vendored packages. This will pass CI, but the next person to run `dep ensure` will get a very large delta by surprise. We should check during CI that `dep ensure` doesn't produce any changes (possibly using `-dry-run`)

johnSchnake:
If someone fails to add a dep then the build will fail but
if they remove one and dont run it then it can lead to
deps hanging around unnecessarily and a future dev/PR getting
a surprising diff.

Fixes #244

dep _says_ that the command is supposed to exit non-zero if it would write something but that is just not true if you try it out/look at the code. Just checking that something gets written to stdout which seems (by looking into the code) to the only way to tell.

I am trying not to hinge on an exact wording like `Would have written` just in case that gets tweaked at some point. We are also pinning the version here though so it should be fine.

**Release note**:
```
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,8 +21,9 @@ Dockerfile-arm64
 *.tar.gz
 *.tar
 
-# kind is downloaded for testing in CI
+# kind/dep are downloaded for testing in CI
 kind
+dep
 
 # coverage profile
 coverage.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ before_install:
   - curl -L https://github.com/kubernetes-sigs/kind/releases/download/v0.3.0/kind-linux-amd64 --output kind && chmod +x kind
   - ./kind create cluster --config kind-config.yaml
   - export KUBECONFIG="$(./kind get kubeconfig-path --name="kind")"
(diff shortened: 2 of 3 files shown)


johnSchnake:
`./dep ensure --dry-run produced output on stdout; did you need add/remove a dependency?`

But didnt fail the build; forgot to exit 1

However, there isn't a reason the lockfile should be written; it is the same data. Didnt think this was the case the first time I tried it but /shrug. May need to check for some other value besides the lockfile. Will double check the code but maybe just an inverted grep is enough?

johnSchnake:
Pushed a commit to show that section now fails as expected. Has since removed it (but you can look at the travis history if you want to see the failure).

I ended up setting the failure against an actual `dep ensure && git status` check because (1) it is clear what/why we are checking (2) `dep ensure --dry-run` oddly always wrote that it was going to write the gopkg.lock so we would have had to do a more awkward, unintuitive check.

PR #756, opened 2019-06-18
Add versioned docs to the Sonobuoy.io site

### LINKED ISSUE #738: Versioned docs
Our site: [sonobuoy.io](sonobuoy.io) currently just has a single version of documentation. As a result it may be confusing as we develop and modify the master docs, but someone is using a version or two older of Sonobuoy.

We should version our docs and have them all available the way [Velero](https://velero.io/docs/master/) does it.

johnSchnake:
Up until now we were serving the site with just a static set of
docs and it was confusing because it had duplicate data (in the
form of 'latest' version pages which were the same as 'master')
but the site also listed the verison as v0.14.2.

This change does the following:
 - adds versioning to the page
 - adds support for versions back to v0.13.0
 - fixes up links (for the main docs) to ensure they work in
this structure. Did not update the enhancement docs which are
internal and not really used/linked to.
 - fixes a bug which caused 'master' to appear in the version list
twice
 - added a warning when viewing the 'master' branch docs that they
are unstable
 - fixed a bug which caused the versioned pages to not properly
be structured due to an accidental `div` tag in the template

Fixes #738

The number of files is extensive because of copying all the data from old releases. For the most part I would say you just need to pull up the example site from netlify (or run it locally) and see if you run into any problems.

**Release note**:
```
Improved documentation on sonobuoy.io to included static snapshots of releases.
```


johnSchnake:
Partly there:

TIL:
 - there is a `versioned` field that has to be set to true
 - below that is a list of versions we manually specify; it does not just read the directories

TODO:
 - add other versions we want to support (back to v0.13.0 I think is fine)
 - figure out the problem with rendering. When going back to the v0.13.0 version (or even the master version) then the dropdown disappears and the width seems to change. Not the typical docs look.

johnSchnake:
Layout info is passed via a section like this; right above the list of versions:
https://github.com/heptio/velero/blob/bbb11a8d23b2f730004a7f36569c75cf4566e94d/site/_config.yml#L70-L87

johnSchnake:
@jonasrosland I seem to have hit a snag and I'm not sure if it is on my end, the css, site generation, etc

If you look at the preview for the v0.13.0 docs https://deploy-preview-756--sonobuoy-live.netlify.com/docs/v0.13.0/ (from this PR) you can see that the content gets put into a `div` that is a sibling to to the LHS menu and the RHS notice about being a snapshot.

Compare that to the Velero site https://velero.io/docs/v0.3.0/ where the content gets put into the same div that the RHS warning is in.

Any idea if this is a problem in the Sonobuoy site or if it is just a difference between the preview and what will actually go live?

jonasrosland:
I'll dig in to this, thanks for letting me know.

jonasrosland:
There's an extra `</div>` at the end of https://github.com/heptio/sonobuoy/blob/6c522c85f4b3ba936101eb2e0ae3a3d0935d7f0a/site/_includes/version-warning.html#L3, if you remove that I think it should work.

johnSchnake:
Awesome, thanks for looking into this.

johnSchnake:
TIL: We also need to update the table of content mapping for each version (otherwise we'll get the same menu/links on the LHS). That is not autogenerated from the files in the folder or something like that.

jonasrosland:
> TIL: We also need to update the table of content mapping for each version (otherwise we'll get the same menu/links on the LHS). That is not autogenerated from the files in the folder or something like that.

I've worked on proper redirects for Velero, and just yesterday figured out how to manage them easily. So now we can replicate that for Contour. I'll add another PR with those changes. You'll be able to use `projectcontour.io/docs` as a pointer to the latest docs. You can go ahead with this one in the meantime and I can make all the changes needed to make it happen after merge.

johnSchnake:
I think this is ready for review; I am not aware of any outstanding issues so anything remaining is an actual oversight that another pair of eyes may be able to see for me.

site/docs/v0.14.3/README.md
```diff
@@ -135,26 +135,26 @@ welcome pull requests. Feel free to dig through the [issues][issue] and jump in.
...
 [docs]: https://github.com/heptio/sonobuoy/tree/master/docs
-[e2e]: /docs/conformance-testing.md
-[gen]: docs/gen.md
+[e2e]: conformance-testing.md
+[gen]: gen.md
 [gimme]: https://github.com/travis-ci/gimme
 [issue]: https://github.com/heptio/sonobuoy/issues
 [k8s]: https://github.com/kubernetes/kubernetes
 [linux]: https://kubernetes.io/docs/tasks/tools/install-kubectl/#tabset-1
 [oview]: https://youtu.be/k-P4hXdruRs?t=9m27s
```
jonasrosland: We should get a newer recording/demo up to better explain the project and demos of new features.
Not blocking the merge of course.

site/_data/v0-14-3toc.yml
```diff
@@ -0,0 +1,28 @@
+toc:
+  - title: Setting up
+    subfolderitems:
+      - page: Getting started
+        url: /index.html
+      - page: Config
+        url: /sonobuoy-config
+      - page: Snapshot layout
+        url: /snapshot
+      - page: ImagePullSecrects
```
jonasrosland: Should this be `ImagePullSecrets` ?

site/_data/v0-14-3toc.yml
```diff
@@ -0,0 +1,28 @@
...
+        url: /conformance-testing
+      - page: Custom Registries & Airgap Testing
+        url: /airgap
+      - page: Plugins
+        url: /plugins
+  - title: Next steps
+    subfolderitems:
+      - page: Customization
+        url: /gen
+      - page: Make your own plugins
```
jonasrosland: I think we need to beef up the docs here for the upcoming versions, explaining _how_ to create a new plugin. Then we can also link to the example directory in that text.

johnSchnake: Yea, I did a blog post last time about this and we need to give stuff like that better visibility for users who start at github (rather than the blog itself)

PR #757, opened 2019-06-18
Add "native" Makefile target

### LINKED ISSUE #715: Homebrew sonobuoy version lacks 'v' prefix
 - Ensure you dont have sonobuoy locally (to be sure to use the one from brew later)
 - brew install sonobuoy
 - `sonobuoy version` (this will lack the `v` prefix)
 - `sonobuoy gen|grep 14` (this will have the image names for sonobuoy being `...:0.14.2` which also lacks the `v` prefix and those images do not exist)

The sonobuoy from brew to be an actual release, report the same version info, and choose proper images.

Originally reported here: https://github.com/heptio/sonobuoy/issues/556#issuecomment-491567885

I also have no idea how sonobuoy gets into homebrew; I'm not involved with this process and I dont know if it is some community effort, a custom build, pulls from master...etc. Need some insight on that. I confirmed that master is just fine and that the release is also fine.

zubron:
This adds a new target which can be used in the Sonobuoy Formula for
Homebrew. Homebrew formulae typically build from source. We are
currently using Docker when building and this is not available when
building for Homebrew. To ensure that we use the same build parameters
for our own builds and builds on Homebrew, we are providing a new make
target (`native`) which which runs `go build` on the host machine. This
build command is shared across the `native` target and our other build
targets.

Fixes #715

- Fixes #715 

I've tested this build on both Linux and macOS, and the existing targets are producing the output in the same locations. I would like input on the naming of things here, and to know whether there's a more idiomatic way to do what I'm suggesting.

**Release note**:
```
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,7 @@ LINT = golint $(GOLINT_FLAGS) $(TEST_PKGS)
 
 WORKDIR ?= /sonobuoy
 DOCKER_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(BUILDMNT) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c
+GO_BUILD ?= CGO_ENABLED=0 $(GO_SYSTEM_FLAGS) go build -o $(BINARY) $(VERBOSE_FLAG) -ldflags="-s -w -X $(GOTARGET)/pkg/buildinfo.Version=$(GIT_VERSION) -X $(GOTARGET)/pkg/buildinfo.GitSHA=$(GIT_REF_LONG)" $(GOTARGET)
 
 .PHONY: all container push clean test local-test local generate plugins int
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -183,4 +184,7 @@ clean:
 deploy_kind:
 	# FIXME: This assumes kind is in cwd; intended to work with our other CI scripts and
 	# is a bit wonky for general use.
-	./kind load docker-image $(REGISTRY)/$(TARGET):$(IMAGE_VERSION) || true
\ No newline at end of file
+	./kind load docker-image $(REGISTRY)/$(TARGET):$(IMAGE_VERSION) || true
+
+brew:
```
johnSchnake: w.r.t. naming I think that I would prefer this to just be something like `native`.

I think it's perfect that you extracted/reused the logic in here and the `build_sonobuoy` target and this is the "right" way to build it locally if you have your repo set up in the appropriate manner. I have recently just run `go build` for my own little tests but using this target would be better and ensure my builds have version info on them and such. 

All of that is to say, even though `brew` was the initial motivation, it has more generic use cases and I think the name should reflect that.

Maybe:
 - `native`
 - `local`
 - another idea?

johnSchnake:
Once this goes in, do you think you can actually put up the PR for homebrew? Or at least tag the folks about this target and make sure they put up a PR.

Makefile
zubron: When first making the target, I thought there might be something a bit more specific to be done for homebrew. You're right though, it would be better to have a more generic name. I'll change it to `native`.

zubron:
> Once this goes in, do you think you can actually put up the PR for homebrew? Or at least tag the folks about this target and make sure they put up a PR.

Yep, I'm happy to do either of those things. It looks like it might require a new release though as it's the release source tarball that is fetched in the formula. I can do another 0.14.* release, or we can just wait until we're releasing v0.15.0.

johnSchnake:
It has been broken since it was created so I dont think it is such a rush that we need a special release. v0.15.0 will be just a matter of days so I'd just intend to fix it then.

PR #758, opened 2019-06-19
Apply changes from running gofmt

zubron:
This PR just applies changes resulting from running `gofmt`.
These changes were included in another branch I'm working on
but I thought it would be better to separate them into their own PR.

**Release note**:
```
None
```

### DIFF
--- a/cmd/sonobuoy/app/gen.go
+++ b/cmd/sonobuoy/app/gen.go
@@ -19,11 +19,11 @@ package app
 import (
 	"fmt"
 	"os"
-	imagepkg "github.com/heptio/sonobuoy/pkg/image"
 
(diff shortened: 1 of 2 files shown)


johnSchnake:
Looks good to me; going to rerun tests though bc they failed. However, it was a great chance to actually hit that new debug logic. Still isn't perfect but we can see some more detail than we did before:
https://travis-ci.org/heptio/sonobuoy/builds/547752201?utm_source=github_status&utm_medium=notification

I can see the plugin is marked as complete on the status, the logs did get into the tarball, but not the junit.

I'll fixup that printing logic to pass so we can see the contents of the pod logs/test run.

PR #762, opened 2019-06-19
Fixup debug line to target regular files and print file name

### LINKED ISSUE #761: Debug lines fail and I can't reproduce
See failure here: https://travis-ci.org/heptio/sonobuoy/builds/547854573

I had added some debug lines for CI when things go badly (since we don't have artifacts) but the command:
```
find results/plugins -exec cat {} \;
```
failed due to 
```
cat: results/plugins: Is a directory
cat: results/plugins/e2e: Is a directory
cat: results/plugins/e2e/results: Is a directory
```

I have tried reproducing this but it works on:
 - my mac
 - a busybox image
 - ubuntu latest
 - ubuntu:14.04.5 (the same os that the travis worker reports)

Print all the info in files the plugins directory

Any idea what I can do to reproduce this?

johnSchnake:
Fixes #761 

Unable to reproduce the problem on any number of images; though the workaround is just to clearly find only the regular files.

**Release note**:
```
NONE
```

### DIFF
--- a/travis-ci.sh
+++ b/travis-ci.sh
@@ -30,8 +30,29 @@ if [ $e2eCode -ne 0 ]; then
     ./sonobuoy status
     ./sonobuoy logs
     mkdir results; tar xzf $outFile -C results
-    find results
-    find results/plugins -exec cat {} \;
+
+    echo "Full contents of tarball:"
+    find results 
+
+    echo "Printing data on the following files:"
+    find results/plugins -type f
+    find results/plugins -type f \
+      -exec echo Printing file info and contents of {} \; \
+      -exec ls -lah {} \; \
+      -exec cat {} \; \
+      -exec echo \; \
+      -exec echo \;
+    echo "[Exit code of find was: $?]"
+
+    echo "Printing data on the following files:"
+    find results/podlogs -type f
+    find results/podlogs -type f \
+      -exec echo Printing file info and contents of {} \; \
+      -exec ls -lah {} \; \
+      -exec cat {} \; \
+      -exec echo \; \
+      -exec echo \;
+    echo "[Exit code of find was: $?]"
     exit $e2eCode
 fi
 


travis-ci.sh
```diff
@@ -31,7 +31,11 @@ if [ $e2eCode -ne 0 ]; then
     ./sonobuoy logs
     mkdir results; tar xzf $outFile -C results
     find results
-    find results/plugins -exec cat {} \;
+    find results/plugins -type f \
+      -exec echo {} \; \
```
zubron: This could be replaced with the `-printf` flag to `find`.
```suggestion
      -printf '%p\n' \
```

travis-ci.sh
```diff
@@ -31,7 +31,11 @@ if [ $e2eCode -ne 0 ]; then
...
+      -exec cat {} \; \
+      -exec echo \; \
```
zubron: And you could echo `\n` directly here using `echo -e "\n\n"`

johnSchnake: Seemed like 6 on one hand, half a dozen on the other.

I wasn't sure if the `-e` flag was as cross-platform supported. If you know one way or the other I'd be curious.

johnSchnake:
Any idea why it doesn't seem to work? https://travis-ci.org/heptio/sonobuoy/builds/547900555#L4013

It is as if none of the exec stuff runs at all.

zubron:
> Any idea why it doesn't seem to work? https://travis-ci.org/heptio/sonobuoy/builds/547900555#L4013
> 
> It is as if none of the exec stuff runs at all.

The file `results/plugins/e2e/results/e2e.log` is listed twice. Is it possible that it's being found as part of the find command but it's empty so there's nothing printed below it?

zubron:
Perhaps you could include some additional markers in those debug lines to help you know what the output from `find` is?

travis-ci.sh
zubron: Sorry, I don't know :( In that case, probably just best to leave it as is.

johnSchnake:
This was green on tests but I reran to try and get a failure. See here: https://travis-ci.org/heptio/sonobuoy/builds/548216979#L4065

This is a known issue upstream and already has an issue filed: https://github.com/kubernetes/kubernetes/issues/78699

LGTY @zubron ? I'd like to merge these now and am happy they've proven so useful in debugging this. Once we fix that upstream bug, hopefully we wont hit this path often though.

travis-ci.sh
```diff
@@ -30,8 +30,29 @@ if [ $e2eCode -ne 0 ]; then
...
+      -exec echo Printing file info and contents of {} \; \
+      -exec ls -lah {} \; \
+      -exec cat {} \; \
+      -exec echo \; \
+      -exec echo \;
+    echo "[Exit code of find was: $?]"
+
+    echo "Printing data on the following files:"
+    find results/podlogs -type f
+    find results/podlogs -type f \
```
johnSchnake: FYI I also added the `podlogs` since those can be vital for debugging.

PR #763, opened 2019-06-19
Add basic validation of configs

### LINKED ISSUE #394: Default configurations
Feature request
**What happened**:
I passed in nil to one of these https://github.com/heptio/sonobuoy/blob/master/pkg/client/interfaces.go#L127-L139 and (for Gen at least) I got a nil pointer exception
I expected (not really, but as a user i might) to get a default configuration instead of crashing hard

**Environment**:
- Kernel (e.g. `uname -a`):

I propose that if a nil is passed in we provide a default configuration.

zubron:
This adds basic checking and validation of configs passed to some of the
interface function implementations in the Sonobuoy client. It is primarily
being added to prevent panics caused by passing nil configs to these
functions. It could be expanded in future to provide more comprehensive
validation of configuration objects.

Fixes #394

- Fixes #394

This does not include comprehensive validation for some of the more
complex configuration objects (e.g. `GenConfig`). It primarily focuses on
the nil checks. A lot of the added tests in the client don't test the
"successful" case, where the execution of the function would proceed
past the validation. This is due to the complexity of the test set up that
would be required which seemed out of scope for this change.

**Release note**:
```
Fixed a bug that caused a panic if a nil configuration was provided to Sonobuoy client functions.
```

### DIFF
--- a/pkg/client/defaults.go
+++ b/pkg/client/defaults.go
@@ -17,6 +17,8 @@ limitations under the License.
 package client
 
 import (
+	"os"
+
 	"github.com/heptio/sonobuoy/pkg/config"
 )
 
@@ -42,7 +44,7 @@ func NewRunConfig() *RunConfig {
 // NewDeleteConfig is a DeleteConfig using default images, RBAC enabled, and DeleteAll enabled.
 func NewDeleteConfig() *DeleteConfig {
 	return &DeleteConfig{
-		Namespace:  config.DefaultImage,
+		Namespace:  config.DefaultNamespace,
 		EnableRBAC: true,
 		DeleteAll:  false,
 	}
@@ -53,5 +55,6 @@ func NewLogConfig() *LogConfig {
 	return &LogConfig{
 		Follow:    false,
 		Namespace: config.DefaultNamespace,
+		Out:       os.Stdout,
 	}
 }
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -29,6 +29,11 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+// ConfigValidator allows the command configurations to be validated.
+type validator interface {
+	Validate() error
+}
+
 // LogConfig are the input options for viewing a Sonobuoy run's logs.
 type LogConfig struct {
 	// Follow determines if the logs should be followed or not (tail -f).
@@ -37,6 +42,15 @@ type LogConfig struct {
 	Namespace string
 	// Out is the writer to write to.
 	Out io.Writer
+}
+
+// Validate checks the config to determine if it is valid.
+func (lc *LogConfig) Validate() error {
+	if lc.Namespace == "" {
+		return errors.New("namespace cannot be empty")
+	}
+
+	return nil
 }
 
 // GenConfig are the input options for generating a Sonobuoy manifest.
@@ -66,6 +80,15 @@ type GenConfig struct {
 	PluginEnvOverrides map[string]map[string]string
 }
 
+// Validate checks the config to determine if it is valid.
+func (gc *GenConfig) Validate() error {
+	if gc.E2EConfig == nil {
+		return errors.New("nil E2EConfig provided")
+	}
+
+	return nil
+}
+
 // E2EConfig is the configuration of the E2E tests.
 type E2EConfig struct {
 	Focus    string
@@ -84,6 +107,12 @@ type RunConfig struct {
 	Wait time.Duration
 }
 
+// Validate checks the config to determine if it is valid.
+func (rc *RunConfig) Validate() error {
+	err := rc.GenConfig.Validate()
+	return errors.Wrap(err, "GenConfig validation failed")
+}
+
 // DeleteConfig are the input options for cleaning up a Sonobuoy run.
 type DeleteConfig struct {
 	Namespace  string
@@ -92,10 +121,28 @@ type DeleteConfig struct {
 	Wait       time.Duration
 }
 
+// Validate checks the config to determine if it is valid.
+func (dc *DeleteConfig) Validate() error {
+	if dc.Namespace == "" {
+		return errors.New("namespace cannot be empty")
+	}
+
+	return nil
+}
+
 // RetrieveConfig are the input options for retrieving a Sonobuoy run's results.
 type RetrieveConfig struct {
 	// Namespace is the namespace the sonobuoy aggregator is running in.
 	Namespace string
+}
+
+// Validate checks the config to determine if it is valid.
+func (rc *RetrieveConfig) Validate() error {
+	if rc.Namespace == "" {
+		return errors.New("namespace cannot be empty")
+	}
+
+	return nil
 }
 
 // StatusConfig is the input options for retrieving a Sonobuoy run's results.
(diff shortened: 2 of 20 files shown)


pkg/client/interfaces_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"testing"
+)
+
+type ConfigTestCase struct {
+	config        ConfigValidator
+	valid         bool
+	expectedError string
+}
+
+func checkConfigIsValid(tc ConfigTestCase) error {
```
johnSchnake: I'd make this entire file one table driven test. You already defined the test case type and the "check" function, but instead I'd define those in the test function itself to prevent namespace pollution.

Then like the other table driven test I mentioned, add a `desc` field and run them all as subtests.

So it would be something like:

```
testCases := []testCase{
  {desc: "valid deleteConfig" ...},
  {desc: "invalid deleteConfig because of xyz" ...},
  {desc: "valid fooConfig" ...},
  {desc: "invalid fooConfig" ...},
t.Run(t.desc,...{

}
```

All the test cases already share the type, interface, etc. By breaking it into completely separate tests you're just making it unclear that you're really testing the same functionality in every case.

pkg/client/logs_test.go
```diff
@@ -201,3 +202,42 @@ func TestLogReaderNoError(t *testing.T) {
...
+		_, err = c.LogReader(tc.config)
+		if !tc.expectedError && err != nil {
+			t.Errorf("Expected no error, got: %v", err)
+		}
+
+		if tc.expectedError {
+			if err == nil {
+				t.Errorf("Expected provided config to be invalid but got no error")
+			} else if !strings.Contains(err.Error(), tc.expectedErrorMsg) {
+				t.Errorf("Expected error to contain '%v', got '%v'", tc.expectedErrorMsg, err.Error())
```
johnSchnake: Just FYI (if you didnt know), you can also use `%q` to automatically have go quote the string.

pkg/client/retrieve.go
```diff
@@ -40,13 +40,22 @@ var tarCommand = []string{
 	fmt.Sprintf("tar cf - %s/*.tar.gz", config.MasterResultsPath),
 }
 
-func (c *SonobuoyClient) RetrieveResults(cfg *RetrieveConfig) (io.Reader, <-chan error) {
+func (c *SonobuoyClient) RetrieveResults(cfg *RetrieveConfig) (io.Reader, <-chan error, error) {
```
johnSchnake: nit: I know this function didn't have a comment before, but since it is an exported method we should add one. Especially since we are modifying the signature. We should make it clear what the channel of errors is for and what the "raw" error is for.

pkg/client/retrieve.go
```diff
@@ -40,13 +40,22 @@ var tarCommand = []string{
...
+	if err := cfg.Validate(); err != nil {
+		return nil, nil, errors.Wrap(err, "config validation failed")
+	}
+
 	ec := make(chan error, 1)
 	client, err := c.Client()
 	if err != nil {
 		ec <- err
-		return nil, ec
+		return nil, nil, err
```
johnSchnake: In this case there isn't a need to put the err on the channel (line above this) since we aren't even returning it.

pkg/client/defaults.go
```diff
@@ -51,7 +52,7 @@ func NewDeleteConfig() *DeleteConfig {
 // NewLogConfig is a LogConfig with follow disabled and default images.
 func NewLogConfig() *LogConfig {
 	return &LogConfig{
-		Follow:    false,
-		Namespace: config.DefaultNamespace,
```
johnSchnake: Since we are leaving in the namespace defaulting in the other methods, we should do the same here, right? Was there some other reason to specifically pull out namespace from this and the delete config?

pkg/client/defaults.go
```diff
@@ -42,7 +44,6 @@ func NewRunConfig() *RunConfig {
 // NewDeleteConfig is a DeleteConfig using default images, RBAC enabled, and DeleteAll enabled.
 func NewDeleteConfig() *DeleteConfig {
 	return &DeleteConfig{
-		Namespace:  config.DefaultImage,
```
johnSchnake: We should fix this, not remove the namespace defaulting. Clearly if no one has complained they aren't using this but if we are going to leave the `NewXyzConfig()` methods then it seems like we should (1) be consistent about setting namespace (2) seems like using `NewXyzConfig()` should pass the new validation routine we're making.

w.r.t. (1) the config.New() method, NewGenConfig, and NewRunConfig would all have their namespaces already set so I think we should do the same here.

I think we talked about this but refresh my memory if we came to a different conclusion; I know we didnt wan't to push CLI stuff here, but it does seem like we can provide the sane default so that it passes validation.

pkg/client/interfaces.go
```diff
@@ -29,6 +29,11 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+// ConfigValidator allows the command configurations to be validated.
+type ConfigValidator interface {
```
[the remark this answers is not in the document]
johnSchnake: In addition, I would actually just call this `validator`. Users of the library:
 - probably dont need the validator type at all, that is just for our own testing
 - if they do need it, they can define their own validator interface in 2-3 lines.

Keeping this unexported helps limit dependencies we feel obligated to support. We can always export it later if we come up with some important reason to do so.

pkg/client/interfaces.go
```diff
@@ -39,6 +44,15 @@ type LogConfig struct {
 	Out io.Writer
 }
 
+// Validate checks the config to determine if it is valid.
+func (lc *LogConfig) Validate() error {
+	if lc.Namespace == "" {
+		return errors.New("Namespace cannot be empty")
```
johnSchnake: nit: errors should be lower cased rather than using sentence capitalization. This applies throughout.
See: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/client/defaults.go
zubron: We did talk about it. I think we were leaning towards eventually removing these if they weren't being used. In the meantime though, they should pass the validation, and given that the `config.New()` results in the `DefaultNamespace` being used, it makes sense for it to be used here for these configs. I'll fix it!

pkg/client/interfaces.go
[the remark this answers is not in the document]
zubron: Ah, I didn't think about the impact of exporting it. I will change that.

pkg/client/interfaces.go
```diff
@@ -66,6 +80,16 @@ type GenConfig struct {
 	PluginEnvOverrides map[string]map[string]string
 }
 
+// Validate checks the config to determine if it is valid.
+func (gc *GenConfig) Validate() error {
+	if gc.E2EConfig == nil {
+		return errors.New("nil E2EConfig provided")
+	}
+
+	// TODO: There are more validations that could take place here
```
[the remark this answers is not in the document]
zubron: You're right. I don't think I'll get to it straight away so I will remove it.

pkg/client/interfaces.go
```diff
@@ -84,6 +108,15 @@ type RunConfig struct {
 	Wait time.Duration
 }
 
+// Validate checks the config to determine if it is valid.
+func (rc *RunConfig) Validate() error {
+	if err := rc.GenConfig.Validate(); err != nil {
+		return errors.Wrap(err, "GenConfig validation failed")
+	}
+
+	return nil
```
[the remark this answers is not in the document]
zubron: Nice! I didn't realise that. Will update these.

pkg/client/interfaces_test.go
zubron: This structure was a side effect of these being separate tests to begin with. I extracted the checking logic, but didn't go the extra step to merge into one test function. I'll update it.

pkg/client/retrieve.go
zubron: Good spot. I missed that!

pkg/client/retrieve.go
zubron: Agreed. I'll add a comment here.

pkg/client/interfaces_test.go
johnSchnake: 👍 I probably should have seen that; I thought they were brand new tests. Thanks for combining them.

Issue #759, opened 2019-06-19
Race in termination without results error mode


johnSchnake:
**What steps did you take and what happened:**
Run an e2e test in a cloud environment and ended up getting a failure due to the container being terminated despite the worker having sent the results. Seems to be racy.

From slack:
```
The issue is that the aggregator node never got the results for the e2e plugin.

We can see in the podlogs for the worker container that it _tried_ to send the results, but we never see the corresponding message on the server side.

Is this reproducible?

There was new logic recently added that would error in these sorts of situations where the pod has terminated but we never got the results. I worry that this may indicate that the server didn’t give enough time to process the results. However, the fact that I don’t even see the message about getting a result makes me slightly unsure.

Since we have another release coming up (right after the k8s release) we can tweak that logic to try to mitigate that.
```

**What did you expect to happen:**
The worker pod logs showed it tried to send results, the server should have had the same log line but it did not, perhaps because it logs _after_ the processing is done? If that is the case we need to a few things:
 - maybe add a log line about starting the processing of results
 - ensure we dont fail for not having results if we are mid-processing
 - increase the timeout before we fail for container-terminated-but-no-results

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Sonobuoy version: v0.14.3

johnSchnake:
Will have to just monitor this going forward; the PR linked to think mitigated the issue and hopefully resolved it sufficiently. We may also want to monitor https://github.com/kubernetes/enhancements/issues/753 for other solutions than the donefile monitoring.

PR #764, opened 2019-06-20
Tweak the timing of considering pod as failing to be more permissive

johnSchnake:
If we are too aggressive with the timeframe, the main plugin container
may terminate and, while the sidecar is reporting the results and
the aggregator is reading them, the aggregator may instead too quickly
decide to consider the plugin a failure (because it hasn't seen the
final results yet).

This change just tries to extend some timing windows to give the
aggregator more time to see these results.

Lastly, it also adds logging into the worker when using retries. This
helps us better track what is happening on the worker's side.

Mitigates #759

**Release note**:
```
Timeframe extended (from 1m to 5m) for the aggregator to see results from a plugin before considering it to have terminated without reporting results. Timeframe also extended (from 15s to 2m) for which the aggregator will wait for clients to retry sending results if there is a problem in the connection.
```

### DIFF
--- a/pkg/plugin/aggregation/aggregator.go
+++ b/pkg/plugin/aggregation/aggregator.go
@@ -38,7 +38,7 @@ import (
 
 const (
 	gzipMimeType       = "application/gzip"
-	defaultRetryWindow = 15 * time.Second
+	defaultRetryWindow = 120 * time.Second
 )
 
 // Aggregator is responsible for taking results from an HTTP server (configured
@@ -261,7 +261,7 @@ func (a *Aggregator) IngestResults(ctx context.Context, resultsCh <-chan *plugin
 		case result, more = <-resultsCh:
 		}
 
-		if !more {
+		if result == nil && !more {
 			return
 		}
 
--- a/pkg/plugin/driver/utils/utils.go
+++ b/pkg/plugin/driver/utils/utils.go
@@ -30,7 +30,10 @@ import (
 )
 
 const (
-	terminatedContainerWindow = 1 * time.Minute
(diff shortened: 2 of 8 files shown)


pkg/plugin/aggregation/aggregator.go
```diff
@@ -261,7 +261,7 @@ func (a *Aggregator) IngestResults(ctx context.Context, resultsCh <-chan *plugin
 		case result, more = <-resultsCh:
 		}
 
-		if !more {
+		if result == nil && !more {
```
johnSchnake: Should always be the same as before, but wasn't sure if there was ever, somehow, that you'd get the `result, !more` instead of `nil, !more`. Just an extra clear check that we will process each result.

pkg/plugin/driver/utils/utils.go
```diff
@@ -30,7 +30,10 @@ import (
 )
 
 const (
-	terminatedContainerWindow = 1 * time.Minute
+	// terminatedContainerWindow is the amount of time after a plugins main container terminates
+	// that we consider it a failure mode. This handles the situation where the plugin container
+	// exits without returning results.
+	terminatedContainerWindow = 5 * time.Minute
```
johnSchnake: Increased 500%; agreed with @zubron 's comment offline that we should try and base this on some evidence/calculation rather than just my wild guesses. But FWIW I've successfully hand results reported with an AWS cluster even with the 1m window so I assume 5m is OK.

johnSchnake:
Rebased; will wait for green tests again just to be safe.

PR #770, opened 2019-06-24
Add v0.15.0 doc snapshot and automate process

johnSchnake:
 - Adds script to automate adding a new version of the docs
 - Removes the now deprecated `docs` directory
 - Adds docs for v0.15.0 and sets as the newest version of the docs
 - Updates the release doc to reference the script that automates
the addition of a new docs section

No formal issue for this; was related to v0.15.0 release.

**Release note**:
```
NONE
```


johnSchnake:
TODO:
 - [ ] New page doesnt seem to use the right toc; simple fix
 - We should update part of the readme that links to the docs to link to the doc site instead of the github page. That is the point of the new site to some extent. This can be in a separate PR though.

scripts/update_docs.sh
```diff
@@ -0,0 +1,76 @@
...
+
+set -x
+
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+DIR=$(cd $SCRIPTS_DIR; cd ..; pwd)
+VERSION=$1
+TOC_NAME=$(echo ${VERSION}toc|sed s/\\./-/g)
+
+read -r -d '' CONFIG_VERSION_BLOCK << EOM
+latest: v0.14.3
```
zubron: This is the hardcoded version that I mentioned. It should be changed to work for the release after this one.

johnSchnake: Fixed. It was as easy as using `v.*`. I'm guessing those `.` in the existing line were already acting as wildcards unintentionally but they just happened to match the literal `.` when perl was doing its matching.

scripts/update_docs.sh
```diff
@@ -0,0 +1,76 @@
...
+
+OLD_FOOTER_BLOCK="cta_url: \/docs\/v.*"
+NEW_FOOTER_BLOCK="cta_url: \/docs\/${VERSION}"
+
+if [ -z "${VERSION}" ]
+then
+        echo "Should be called with version as first argument. No argument given, not creating docs."
+else
+    docker run --rm \
+        -v ${DIR}:/root \
```
zubron: Running this locally creates the volume mount `/home/bridget/go/src/github.com/heptio:/root`. This results in future steps failing as the paths are incorrect. For example, `/root/site/docs/${VERSION}` should be `/root/sonobuoy/site/...`

zubron: I think it's because I'm getting an error for `BASH_SOURCE`: `./scripts/update_docs.sh: 6: ./scripts/update_docs.sh: Bad substitution`. Is this because it's being interpreted by `/bin/sh` rather than bash? Can we use `$0` instead to refer to the first argument to the script which the script name?

zubron: I can confirm that replacing `${BASH_SOURCE[0]}` with `$0` fixes it.

johnSchnake: Really? I'm getting `.../heptio/sonobuoy` whether I'm in the sonobuoy directory or the scripts directory.

Mind modifying the script a bit just for debug purposes and letting me know what the output is?

```
SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
DIR=$(cd $SCRIPTS_DIR; cd ..; pwd)

# Could you add these?
echo $SCRIPTS_DIR
echo $DIR
dirname "${BASH_SOURCE[0]}"
```

johnSchnake: Thanks for debugging it; I didnt see your most recent updates when I made my comment. Let me try your solution on my mac and see if that works just as well.

johnSchnake: Confirmed it works just fine with your suggested edit. Updated.

scripts/update_docs.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh
+
+set -x
+
+SCRIPTS_DIR="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
+DIR=$(cd $SCRIPTS_DIR; cd ..; pwd)
+VERSION=$1
+TOC_NAME=$(echo ${VERSION}toc|sed s/\\./-/g)
+
+read -r -d '' CONFIG_VERSION_BLOCK << EOM
```
zubron: These `read` commands are failing locally for me. I'm getting the error: ``

Again, it seems to be a difference in `read` in `sh` vs `bash`. If I use `bash` in the shebang at the top of the script it works as expected. Can we use bash instead?

johnSchnake: Absolutely; no loyalty, just habit. Let me know if the new version works for you.

zubron: It works :) thanks!

PR #771, opened 2019-06-24
Update README to point to docs site

johnSchnake:
 - Fixes a small formatting issue
 - Links to master docs on site, not github
 - Updated script to ensure the docs link gets updated on future releases
 - Updated the new v0.15.0 docs with the same change

**Release note**:
```
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ sonobuoy logs
 ## More information
 
 [The documentation][docs] provides further information about:
+
  * [conformance tests][conformance]
  * [plugins][plugins]
  * Testing of [air gapped clusters][airgap].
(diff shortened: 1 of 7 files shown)


johnSchnake:
I found a few broken links and then also realized all the other places we were linking from the website back to github. Fixing those up as well.

One frustrating issue is that the way the docs work, the README.md is duplicated in both the root location for the repo (so github shows it) and also in `site/docs/master/README.md` so the versioned docs show it.

The basic relative links wont work because the readme is a sibling to the other docs in `master/docs` but not in main repo.

I'm just going to make all of those doc links go to the sonobuoy.io site and then we need to automate/check that the readme gets kept in sync in CI.

johnSchnake:
Created #772 so not lose sight of some of the concerns this raised. I don't want to keep prioritizing just this sort of work at the moment though so I'm going to move onto other work besides doc stuff after this.

PR #779, opened 2019-06-26
Allow use of pod names other than "sonobuoy"

lander2k2:
Currently sonobuoy always expects it's pod name to be `sonobuoy`.
So if you run sonobuoy with a CronJob, the controller will create a pod name
something like `sonobuoy-1561584180-zrwjf` and sonobuoy will fail to update
it's own status.

This change allows sonobuoy to fetch it's pod name from an environment
variable which enables using the downward API so that sonobuoy can use
whatever pod name it may be given.

If the environment variable is _not_ set, it will fall back to the
default hard-coded pod name it currently uses.

None I'm aware of

n/a

**Release note**:
```
Allow use of alternative pod names to enable using controllers (e.g. CronJob) to run sonobuoy scans.
```

### DIFF
--- a/pkg/client/testdata/default-plugins-via-nil-selection.golden
+++ b/pkg/client/testdata/default-plugins-via-nil-selection.golden
@@ -99,6 +99,10 @@ spec:
       valueFrom:
         fieldRef:
           fieldPath: status.podIP
+    - name: SONOBUOY_POD_NAME
+      valueFrom:
+        fieldRef:
+          fieldPath: metadata.name
     image: gcr.io/heptio-images/sonobuoy:static-version-for-testing
     imagePullPolicy: IfNotPresent
     name: kube-sonobuoy
--- a/pkg/client/testdata/default.golden
+++ b/pkg/client/testdata/default.golden
@@ -99,6 +99,10 @@ spec:
       valueFrom:
         fieldRef:
           fieldPath: status.podIP
+    - name: SONOBUOY_POD_NAME
+      valueFrom:
+        fieldRef:
+          fieldPath: metadata.name
     image: gcr.io/heptio-images/sonobuoy:static-version-for-testing
     imagePullPolicy: IfNotPresent
     name: kube-sonobuoy
--- a/pkg/client/testdata/e2e-default.golden
+++ b/pkg/client/testdata/e2e-default.golden
@@ -70,6 +70,10 @@ spec:
       valueFrom:
         fieldRef:
           fieldPath: status.podIP
+    - name: SONOBUOY_POD_NAME
+      valueFrom:
+        fieldRef:
+          fieldPath: metadata.name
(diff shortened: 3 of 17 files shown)


johnSchnake:
The unit tests are failing on this because the yaml generated doesn't match the expected values (due to the new env var on the aggregator).

You can fix that by running (with the appropriate packages if it is more than this one):
```
go test github.com/heptio/sonobuoy/pkg/client -update
```
Then add those changes to the PR.

johnSchnake:
How do you see this interacting with the rest of the tooling, e.g. `sonobuoy retrieve` and `sonobuoy status` for instance currently look for the pod by name.

It sounds like you have a use case, but I'm not exactly sure how you implement it so I'd love to just hear some more exposition.

From what it sounds like, you're manually modifying the sonobuoy yaml and not using the CLI itself at all to monitor or get results. Is that right? What is your flow for those things?

lander2k2:
@johnSchnake Apologies, I should've described the use-case to begin with.

The intent here is not to use the sonobuoy CLI but rather install a CronJob as a standard cluster add-on that will run sonobuoy scans periodically.

The results will be processed by a sidecar on the master that will check status and examine results.  It may also push the results to an object store.

Hope that clarifies a little.  LMK if you have any more questions or would like more detail.

zubron:
Hi @lander2k2! Thanks for the extra info. @johnSchnake has created an issue to better capture your use case here: https://github.com/heptio/sonobuoy/issues/782

We agree that it should definitely allow for different pod names, and that it shouldn't rely on a hardcoded name. One idea that came up was to find the aggregator pod by a label that are applied to it, rather than the name. You can see in the template for the aggregator pod, we apply the label [`run: sonobuoy-master`](https://github.com/heptio/sonobuoy/blob/master/pkg/templates/manifest.go#L113).

We have a few other places in the code base where we are finding the pod by that name: [getting the status](https://github.com/heptio/sonobuoy/blob/master/pkg/plugin/aggregation/status.go#L85) and [setting the status](https://github.com/heptio/sonobuoy/blob/master/pkg/plugin/aggregation/status.go#L85). We could introduce a helper function to find the aggregator pod within a particular namespace by labels rather than name using the [`List` method](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/pod.go#L46). Using this approach throughout would ensure that randomly assigned pod names would continue to work with the CLI functions.

lander2k2:
@zubron That's an excellent point.  I will try that out (using pod label rather than downward api).  I'll add any relevant comments to the new issue as well.

lander2k2:
@zubron @johnSchnake 
I submitted an alternative PR: https://github.com/heptio/sonobuoy/pull/787
If you're happy that the other is a better implementation, I think we can close this PR.

johnSchnake:
👍 I think the replacement PR is nearly ready to merge so I'll go ahead and close this one.

PR #785, opened 2019-06-27
Preserve single file result names and keep job/ds depth the same

### LINKED ISSUE #741: Inconsistent path for results in tarball
If you run a plugin which reports a single file it will go to:

for jobs: `plugins/plugin_type` where that is the file, not the directory.
for daemonsets: `plugins/plugin_type/node`, again, thats the file, not directory.

If you return an archive from the plugin, the paths above are the _directories_ which the results get untarr'd into.

I expected those paths to always be directories, the sonobuoy worker should also be uploading the file with the proper headers so that we preserve the filename when downloading it on the aggregator. Currently we drop the remote name and (as you can see above) the name of the results file becomes the name of the node, which is confusing and loses the file extension information.

johnSchnake:
If there is a job returning results it currently goes into the
tarball at plugins/name/results/<files> but this is different
from daemonset results which have an extra directory stating
the node name (plugins/name/node/results/<files>).

To prepare for consistent handling of results, it would be preferred
that these have the same layouts so we added a new level for jobs:
"global" so a job plugin results go to plugins/name/global/results.

In addition, if a single file was uploaded as results, it was handled
in a unique way. Instead of being put into the directories listed above,
the same path would refer, instead, to a file. The contents of which
were the result.

It would be much preferred to preserve the folder structure and the
filename being uploaded. This way automated processing is easier and
users can rationalize about what files they are uploading (when they
would reasonably expect them, by name, to appear in the tarball).

Fixes #741

TODO:
 - [x] Add more tests (waiting to see coverage report but I know I'll need some more)
 - [x] Will add details here how to test it and confirm different situations.
 - [x] Fixup existing unit tests
 - [x] Add new version of tarball structure

**Release note**:
```
Made the location of plugin results within the results tarball more consistent. Job type plugins will have their results placed into `plugins/<name>/global/results/<files>` to be at a consistent depth with daemonset plugins (which list their nodes in the path). In addition, when a plugin reports a single file (not an archive) of results, then that file is put into the same directory and with the same name, it had locally on the container.
```

### DIFF
--- a/cmd/sonobuoy/app/worker.go
+++ b/cmd/sonobuoy/app/worker.go
@@ -40,7 +40,7 @@ func NewCmdWorker() *cobra.Command {
 	var workerCmd = &cobra.Command{
 		Use:    "worker",
 		Short:  "Gather and send data to the sonobuoy master instance (for internal use)",
-		Run:    runGather,
+		Run:    runGatherHelp,
 		Hidden: true,
 		Args:   cobra.ExactArgs(0),
 	}
@@ -65,7 +65,7 @@ var singleNodeCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(0),
 }
 
-func runGather(cmd *cobra.Command, args []string) {
+func runGatherHelp(cmd *cobra.Command, args []string) {
 	cmd.Help()
 }
 
@@ -112,23 +112,7 @@ func loadAndValidateConfig() (*plugin.WorkerConfig, error) {
 }
 
 func runGatherSingleNode(cmd *cobra.Command, args []string) {
-	cfg, err := loadAndValidateConfig()
-	if err != nil {
-		errlog.LogError(err)
-		os.Exit(1)
-	}
-
-	client, err := getHTTPClient(cfg)
-	if err != nil {
-		errlog.LogError(err)
-		os.Exit(1)
-	}
-
-	// A single-node results URL looks like:
-	// http://sonobuoy-master:8080/api/v1/results/by-node/node1/systemd_logs
-	url := cfg.MasterURL + "/" + cfg.NodeName + "/" + cfg.ResultType
-
-	err = worker.GatherResults(cfg.ResultsDir+"/done", url, client, sigHandler(plugin.GracefulShutdownPeriod*time.Second))
+	err := runGather(false)
 	if err != nil {
 		errlog.LogError(err)
 		os.Exit(1)
@@ -136,27 +120,34 @@ func runGatherSingleNode(cmd *cobra.Command, args []string) {
 }
 
 func runGatherGlobal(cmd *cobra.Command, args []string) {
-	cfg, err := loadAndValidateConfig()
+	err := runGather(true)
 	if err != nil {
 		errlog.LogError(err)
 		os.Exit(1)
+	}
+}
+
+func runGather(global bool) error {
+	cfg, err := loadAndValidateConfig()
+	if err != nil {
+		return errors.Wrap(err, "loading config")
 	}
 
 	client, err := getHTTPClient(cfg)
 	if err != nil {
-		errlog.LogError(err)
-		os.Exit(1)
+		return errors.Wrap(err, "getting HTTP client")
 	}
 
-	// A global results URL looks like:
-	// http://sonobuoy-master:8080/api/v1/results/global/systemd_logs
-	url := cfg.MasterURL + "/" + cfg.ResultType
+	// A single-node results URL looks like:
+	// http://sonobuoy-master:8080/api/v1/results/by-node/node1/systemd_logs
+	url := cfg.MasterURL + "/" + cfg.NodeName + "/" + cfg.ResultType
+	if global {
+		// http://sonobuoy-master:8080/api/v1/results/global/systemd_logs
+		url = cfg.MasterURL + "/" + cfg.ResultType
+	}
 
 	err = worker.GatherResults(cfg.ResultsDir+"/done", url, client, sigHandler(plugin.GracefulShutdownPeriod*time.Second))
-	if err != nil {
-		errlog.LogError(err)
-		os.Exit(1)
-	}
+	return errors.Wrap(err, "gathering results")
 }
 
 func getHTTPClient(cfg *plugin.WorkerConfig) (*http.Client, error) {
--- a/pkg/client/e2e.go
+++ b/pkg/client/e2e.go
@@ -35,6 +35,7 @@ func (*SonobuoyClient) GetTests(reader io.Reader, show string) ([]reporters.JUni
 	read := results.NewReaderWithVersion(reader, "irrelevant")
 	junitResults := reporters.JUnitTestSuite{}
 	e2eJunitPath := path.Join(results.PluginsDir, e2e.ResultsSubdirectory, e2e.JUnitResultsFile)
+	legacye2eJunitPath := path.Join(results.PluginsDir, e2e.LegacyResultsSubdirectory, e2e.JUnitResultsFile)
 
 	found := false
 	err := read.WalkFiles(
@@ -44,9 +45,9 @@ func (*SonobuoyClient) GetTests(reader io.Reader, show string) ([]reporters.JUni
 			}
 			// TODO(chuckha) consider reusing this function for any generic e2e-esque plugin results.
 			// TODO(chuckha) consider using path.Join()
-			if path == e2eJunitPath {
+			if path == e2eJunitPath || path == legacye2eJunitPath {
 				found = true
-				return results.ExtractFileIntoStruct(e2eJunitPath, path, info, &junitResults)
+				return results.ExtractFileIntoStruct(path, path, info, &junitResults)
 			}
 			return nil
 		})
--- a/pkg/client/results/e2e/consts.go
+++ b/pkg/client/results/e2e/consts.go
@@ -19,6 +19,11 @@ package e2e
 
 const (
 	// ResultsDirectory is the directory where the results will be found
-	ResultsSubdirectory = "e2e/results/"
-	JUnitResultsFile    = "junit_01.xml"
+	ResultsSubdirectory = "e2e/results/global/"
+
+	// LegacyResultsSubdirectory was the directory where e2e results were found before
+	// the global folder was added.
+	LegacyResultsSubdirectory = "e2e/results/"
+
+	JUnitResultsFile = "junit_01.xml"
 )
--- a/pkg/client/results/reader.go
+++ b/pkg/client/results/reader.go
@@ -22,11 +22,13 @@ import (
 	"compress/gzip"
 	"encoding/json"
 	"encoding/xml"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
 
+	goversion "github.com/hashicorp/go-version"
 	"github.com/heptio/sonobuoy/pkg/config"
 	"github.com/pkg/errors"
 )
@@ -44,14 +46,19 @@ const (
 	defaultNodesFile          = "Nodes.json"
 	defaultServerVersionFile  = "serverversion.json"
 	defaultServerGroupsFile   = "servergroups.json"
-)
 
-const (
 	// UnknownVersion lets the consumer know if this client can detect the archive version or not.
 	UnknownVersion = "v?.?"
 	VersionEight   = "v0.8"
 	VersionNine    = "v0.9"
 	VersionTen     = "v0.10"
+	VersionFifteen = "v0.15"
+)
+
+var (
+	// v15 is the first version we started used a typed version. Allows more clean comparisons
+	// between versions.
+	v15 = goversion.Must(goversion.NewVersion("v0.15.0"))
 )
 
 // Reader holds a reader and a version. It uses the version to know where to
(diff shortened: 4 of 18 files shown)


cmd/sonobuoy/app/worker.go
```diff
@@ -112,51 +112,42 @@ func loadAndValidateConfig() (*plugin.WorkerConfig, error) {
 }
 
 func runGatherSingleNode(cmd *cobra.Command, args []string) {
-	cfg, err := loadAndValidateConfig()
-	if err != nil {
-		errlog.LogError(err)
-		os.Exit(1)
-	}
-
-	client, err := getHTTPClient(cfg)
```
johnSchnake: Fixed these repetitive sections while I was here. The only thing different between these methods were the URLs so I extracted that logic.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -285,24 +285,24 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
 	}
 
 	// Create the output directory for the result.  Will be of the
-	// form .../plugins/:results_type/:node.json (for DaemonSet plugins) or
+	// form .../plugins/:results_type/:node/filename.json (for DaemonSet plugins) or
 	// .../plugins/:results_type.json (for Job plugins)
```
johnSchnake: Need to update this comment.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -285,24 +285,24 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
...
-	resultsFile := path.Join(a.OutputDir, result.Path())
-	resultsDir := path.Dir(resultsFile)
+	resultsDir := filepath.Join(a.OutputDir, result.Path())
+	resultFile := filepath.Join(resultsDir, result.Filename)
 
 	if err := os.MkdirAll(resultsDir, 0755); err != nil {
-		errors.Wrapf(err, "couldn't create directory %v", resultsDir)
+		errors.Wrapf(err, "couldn't create directory %q", resultsDir)
```
johnSchnake: Errors when iterating made it clear that quotes here would be easier due to confusion when the error string ends up there.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -311,7 +311,7 @@ func (a *Aggregator) handleResult(result *plugin.Result) error {
 }
 
 func (a *Aggregator) handleArchiveResult(result *plugin.Result) error {
-	resultsDir := path.Join(a.OutputDir, result.Path())
+	resultsDir := filepath.Join(a.OutputDir, result.Path())
```
johnSchnake: Somewhat preemptively changed these to `filepath`; it will help when we start ensuring compatibility on windows nodes. No change otherwise.

pkg/plugin/aggregation/handler.go
```diff
@@ -35,6 +36,10 @@ const (
 	resultsByNode = "/api/v1/results/by-node/{node}/{plugin}"
 	// resultsGlobal is the path for global (non node-specific) results to be PUT
 	resultsGlobal = "/api/v1/results/global/{plugin}"
+
+	// defaultFilename is the name given to the file if no filename is given in the
+	// content-disposition header
+	defaultFilename = "result"
```
johnSchnake: It used to be that when single files were uploaded, the node name was used as the result file name. Now, since we are keeping the node name as the folder, we needed to come up with a default name.

Not married to this so if you have a preference let me know.

pkg/plugin/aggregation/handler.go
```diff
@@ -67,17 +72,27 @@ func NewHandler(resultsCallback func(*plugin.Result, http.ResponseWriter)) http.
 	return handler
 }
 
-func (h *Handler) resultsHandler(w http.ResponseWriter, r *http.Request) {
-	logRequest(r)
-	vars := mux.Vars(r)
-
+func resultFromRequest(r *http.Request, muxVars map[string]string) *plugin.Result {
```
johnSchnake: Intend to have a test for this logic; just to ensure we pull/default values correctly.

pkg/plugin/driver/job/job.go
```diff
@@ -72,7 +72,7 @@ func NewPlugin(dfn plugin.Definition, namespace, sonobuoyImage, imagePullPolicy,
 // a Job only launches one pod, only one result type is expected.
 func (p *Plugin) ExpectedResults(nodes []v1.Node) []plugin.ExpectedResult {
 	return []plugin.ExpectedResult{
-		{ResultType: p.GetResultType()},
+		{ResultType: p.GetResultType(), NodeName: plugin.GlobalResult},
```
johnSchnake: Theoretically, if you have a node named "global" then it becomes ambiguous (when looking at the tarball by itself) if a job or daemonset was run.

However, that seems to be a fair tradeoff since by doing this we get rid of another ambiguity (if a job uploaded results with directories named with the nodes).

Both are unlikely and this is preferred since it helps make the two structures consistent.

pkg/worker/worker.go
```diff
@@ -72,8 +72,8 @@ func handleWaitFile(resultFile, url string, client *http.Client) error {
 	}()
 
 	// transmit back the results file.
-	return DoRequest(url, client, func() (io.Reader, string, error) {
+	return DoRequest(url, client, func() (io.Reader, string, string, error) {
 		outfile, err = os.Open(resultFile)
-		return outfile, mimeType, errors.WithStack(err)
+		return outfile, filepath.Base(resultFile), mimeType, errors.WithStack(err)
```
johnSchnake: This is the logic that kicks off informing the rest of the code what the filename is.

Then the handler will add it to the headers and the aggregator will read the header to name the file on its side.

johnSchnake:
Will need to update the logic for getting e2e results as well;

CI shows: `results/plugins/e2e/results/global/junit_01.xml` but fails because `ERRO[0000] could not get tests from archive: failed to find results file "plugins/e2e/results/junit_01.xml" in archive `

johnSchnake:
e2e tests and command passing; now it is just all the unit tests that need fixed up. Will be a bit of a pain because there are test tarballs that get tested and I'll have to deal with copying/versioning them so it will be a bit non-trivial.

Would love a review at this point though before I get into the weeds with that stuff.

zubron:
I'm reviewing it now! :)

pkg/client/results/e2e/consts.go
```diff
@@ -19,6 +19,6 @@ package e2e
 
 const (
 	// ResultsDirectory is the directory where the results will be found
-	ResultsSubdirectory = "e2e/results/"
+	ResultsSubdirectory = "e2e/results/global/"
```
johnSchnake: This fixes the e2e command for now, but unlike the rest of the results stuff, it wasn't versioned.

I could modify this to look in the appropriate location based on the version. One of the frustrating things about this is that the config readers are not rewindable so you only get one read-per open. That would be great to be able to change that.

johnSchnake: So even I learned something going through this code. I originally thought it was versioned independently of Sonobuoy because I saw things like [this](https://github.com/heptio/sonobuoy/blob/master/pkg/client/results/reader.go#L51-L54). However, the version it uses (to compare those to) is actually the version stored in the config file; i.e. the Sonobuoy version.

Those are just versioning it based on the major/minor number and not patch releases which is why I (in another snippet of code) moved to using go-version which will let me do more complex checks with versions that look like semver (the existing checks in the Reader just want to know if the version has that prefix, e.g. has the same major/minor version)

pkg/client/e2e.go
```diff
@@ -44,9 +45,9 @@ func (*SonobuoyClient) GetTests(reader io.Reader, show string) ([]reporters.JUni
 			}
 			// TODO(chuckha) consider reusing this function for any generic e2e-esque plugin results.
 			// TODO(chuckha) consider using path.Join()
-			if path == e2eJunitPath {
+			if path == e2eJunitPath || path == legacye2eJunitPath {
```
johnSchnake: Since the old path is at a parent directory from the new one (and so a user can't actually write data to that file that could be misleading) it is fine to just search for either of the paths.

pkg/client/e2e.go
```diff
@@ -44,9 +45,9 @@ func (*SonobuoyClient) GetTests(reader io.Reader, show string) ([]reporters.JUni
...
 				found = true
-				return results.ExtractFileIntoStruct(e2eJunitPath, path, info, &junitResults)
+				return results.ExtractFileIntoStruct(path, path, info, &junitResults)
```
johnSchnake: All of these `Extract...` functions are kind of hard to read IMO; I think in some cases we're misusing them and we don't need to be first filtering for the filepath. The only reason to do so here is to (1) mark that we actually found the file and (2) here we want to extract _either_ file into the struct instead of a hardcoded one.

pkg/plugin/aggregation/aggregator_test.go
```diff
@@ -94,7 +94,7 @@ func TestAggregation_noExtension(t *testing.T) {
 
 func TestAggregation_tarfile(t *testing.T) {
 	expected := []plugin.ExpectedResult{
-		{ResultType: "e2e"},
+		{ResultType: "e2e", NodeName: "global"},
```
johnSchnake: I've done the "nodename=global" work but it does have the tiny problem of: what if someone actually has a node with that name? Does it make a problem?

So it seems like it might but we actually err out if two plugins have the same `ResultType` so there isn't a risk of path conflics.

The only concern is when parsing the results will we interpret `e2e/results/global` as a job with results or a daemonset with results from a single node named `global`.

Since those are really just problems with reading folder names it doesn't seem like we can get away from _some_ ambiguity so it seems fine to me. I just wanted to call it out so another brain can at least consider it.

pkg/plugin/aggregation/aggregator_test.go
```diff
@@ -125,6 +125,9 @@ func TestAggregation_tarfile(t *testing.T) {
 			}
 		} else {
 			t.Errorf("AggregationServer didn't record a result for e2e tests. Got: %+v", agg.Results)
+			for k, v := range agg.Results {
+				t.Logf("Result %q: %+v\n", k, v)
```
johnSchnake: The set of results is a map so you get just the `key: <memory address>` when printing the map above. This makes it clear what each result is if you get an error (which I did while making these changes)

pkg/plugin/aggregation/aggregator_test.go
```diff
@@ -286,56 +289,58 @@ func TestAggregation_RetryWindow(t *testing.T) {
...
-		}
-		defer testDataReader.Close()
-
-		var r io.Reader
-		if tc.simulateErr {
-			r = iotest.TimeoutReader(testDataReader)
-		} else {
-			r = strings.NewReader("foo")
-		}
+		t.Run(tc.desc, func(t *testing.T) {
```
johnSchnake: This was a table-driven test but was probably implemented before subtests in golang.

I've just indented everything; I originally had an error here and couldn't tell why because it was running all test cases in one go.

pkg/plugin/aggregation/aggregator_test.go
```diff
@@ -286,56 +289,58 @@ func TestAggregation_RetryWindow(t *testing.T) {
...
+			// check time before/after wait and ensure it is greater than the retryWindow.
+			time.Sleep(tc.postProcessSleep)
+			start := time.Now()
+			agg.Wait(make(chan bool))
+			waitTime := time.Now().Sub(start)
+
+			// Add buffer to avoid raciness due to processing time.
+			diffTime := waitTime - tc.expectExtraWait
+			if diffTime > testBufferDuration || diffTime < -1*testBufferDuration {
+				t.Errorf("Expected Wait() to wait the duration %v (+/- %v), instead waited %v", tc.expectExtraWait, testBufferDuration, waitTime)
```
johnSchnake: Cleaned up the error message; it was confusing in cases 
 - where we weren't causing errors
 - where the wait time > expectedWaitTime
 - what precision we expected

pkg/plugin/aggregation/aggregator_test.go
```diff
@@ -286,56 +289,58 @@ func TestAggregation_RetryWindow(t *testing.T) {
...
 
 		// Send an error
-		resultsCh <- pluginutils.MakeErrorResult("e2e", map[string]interface{}{"error": "foo"}, "")
+		resultsCh <- pluginutils.MakeErrorResult("e2e", map[string]interface{}{"error": "foo"}, "global")
 		agg.Wait(make(chan bool))
 
-		if result, ok := agg.Results["e2e"]; ok {
-			bytes, err := ioutil.ReadFile(path.Join(agg.OutputDir, result.Path()))
+		if result, ok := agg.Results["e2e/global"]; ok {
+			bytes, err := ioutil.ReadFile(path.Join(agg.OutputDir, result.Path(), "error.json"))
```
johnSchnake: Again, the error, being a single file, used to be named the same as the expected path. This would be counterintuitive to users who are not expecting users or even single files: when running e2e for instance you assume the directory will exist and you can `cd` to it and print files. In a CI script, that would fail if you didn't have the directory you expected.

pkg/plugin/aggregation/handler.go
johnSchnake: Just added this test.

johnSchnake:
Rebased; I think this is done and ready for final review though.

To review the real output I used this command (which uses some images I made available on my dockerhub):

```
sonobuoy run --sonobuoy-image=schnake/sonobuoy:keepFilenames --plugin=./singleFile.plugin --wait --image-pull-policy=Always --plugin=./singleFileDS.plugin --plugin ./tarresults.plugin --plugin ./tarresultsds.plugin
```

It uses 4 plugins defined locally which investigate the combination of {single file vs tarfile} & {job vs daemonset}

The single file plugin def is:
```
sonobuoy-config:
  driver: Job
  plugin-name: singlefile
  result-type: singlefile
spec:
  command:
  - ./run.sh
  image: schnake/singlefileresults:v0.1
  name: plugin
  resources: {}
  volumeMounts:
  - mountPath: /tmp/results
    name: results
```

and the tar file one is:
```
sonobuoy-config:
  driver: Job
  plugin-name: tar
  result-type: tar
spec:
  args:
  - echo hello world
  - echo hello world pt 2
  command:
  - ./run.sh
  image: schnake/easy-sonobuoy-cmds:v0.1
  name: plugin
  resources: {}
  volumeMounts:
  - mountPath: /tmp/results
    name: results
```

To make the other 2, just copy the files modify `job->daemonset` and modify the name so it is unique for every plugin.

pkg/plugin/interface.go
```diff
@@ -27,6 +27,13 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+var (
```
zubron: Can this be `const` instead?

pkg/plugin/aggregation/aggregator_test.go
johnSchnake: We are dumping all the plugin info into a config map with a known name; we can also read that in and iterate over the data. You're correct that is the most accurate thing to automate; on the next PR I may do that as well but I do think that it helps to have the path also provide implicit context so that you don't have to fully parse a separate file at a known location to get that information.

pkg/client/results/reader.go
```diff
@@ -111,18 +118,30 @@ func DiscoverVersion(reader io.Reader) (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "error extracting config")
 	}
-	var version string
+
+	parsedVersion, err := validateVersion(conf.Version)
```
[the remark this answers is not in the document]
johnSchnake: I'll take a look; I originally just copied the code snippet from another location and I think the validation may have done another task there which I removed, so it may not be necessary now.

johnSchnake:
Only change this doesn't have from master is the regexp stuff for `e2e --rerun`; no problem squashing it.

Since the last change I just tweaked the code based on the comments:
 - added another test for the request headers
 - made the global string a constant and used it in a few places instead of restating the literal
 - made the path listed in an error the full path instead of just the base name of the file.

Merging.

PR #786, opened 2019-07-01
Return exit code 1 if sonobuoy can't pull all images required

johscheuer:
This PR ensures that `sonobuoy images pull` will return an exit code 1 if it can't pull all required images for the e2e tests. This is a follow up on https://github.com/heptio/sonobuoy/pull/781 and allows users to integrate a CI test in their pipelines to check if all images are present in the Docker registry (without any grep).

- Fixes #

**Release note**:
```
sonobuoy images pull will return an exit code 1 if some images are not available.
```


johnSchnake:
Thanks for the submission!

PR #787, opened 2019-07-01
Look up sonobuoy master pod name by label

lander2k2:
Currently sonobuoy always expects it's pod name to be sonobuoy.
So if you run sonobuoy with a CronJob, the controller will create a pod name
something like `sonobuoy-1561584180-zrwjf` and sonobuoy will fail to update
it's own status.  Also the sonobuoy CLI will not find the pod if you
check status or try to retrieve results.

This change removes the statically-set sonobuoy master pod name and
looks up the pod name based on the label given to that pod:
'run=sonobuoy-master`.

This assumes that this label is always set and will fail if more than
one pod exists in the namespace with that label.

- https://github.com/heptio/sonobuoy/issues/782

This PR is an alternative implementation to: https://github.com/heptio/sonobuoy/pull/779

**Release note**:
```
Allow use of alternative pod names to enable using controllers (e.g. CronJob) to run sonobuoy scans.
```

Signed-off-by: Richard Lander <landerr@vmware.com>

### DIFF
--- a/pkg/plugin/aggregation/status.go
+++ b/pkg/plugin/aggregation/status.go
@@ -82,9 +82,12 @@ func GetStatus(client kubernetes.Interface, namespace string) (*Status, error) {
 		return nil, errors.Wrap(err, "sonobuoy namespace does not exist")
 	}
 
+	// Determine sonobuoy pod name
+	SetStatusPodName(client, namespace)
+
 	pod, err := client.CoreV1().Pods(namespace).Get(StatusPodName, metav1.GetOptions{})
 	if err != nil {
-		return nil, errors.Wrap(err, "could not retrieve sonobuoy pod")
+		return nil, errors.New("could not retrieve sonobuoy pod")
 	}
 
 	if pod.Status.Phase != corev1.PodRunning {
--- a/pkg/plugin/aggregation/update.go
+++ b/pkg/plugin/aggregation/update.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 
@@ -31,7 +32,11 @@ import (
 
 const (
 	StatusAnnotationName = "sonobuoy.hept.io/status"
-	StatusPodName        = "sonobuoy"
+	StatusPodLabel       = "run=sonobuoy-master"
+)
+
+var (
+	StatusPodName = "sonobuoy"
 )
 
 // node and name uniquely identify a single plugin result
@@ -155,5 +160,29 @@ func GetPatch(annotation string) map[string]interface{} {
 				StatusAnnotationName: annotation,
 			},
 		},
+	}
+}
+
+// SetStatusPodName sets the sonobuoy master pod name based on it's label.
+func SetStatusPodName(client kubernetes.Interface, namespace string) {
+
+	listOptions := metav1.ListOptions{
+		LabelSelector: StatusPodLabel,
+	}
+
+	podList, err := client.CoreV1().Pods(namespace).List(listOptions)
+	if err != nil {
+		logrus.Errorf("Error listing pods with label '%s': %s", StatusPodLabel, err)
+		return
+	}
+
+	switch {
+	case len(podList.Items) == 0:
+		logrus.Errorf("No pods found with label '%s' in namespace %s, using default pod name 'sonobuoy'", StatusPodLabel, namespace)
+	case len(podList.Items) > 1:
+		logrus.Warningf("Found more than one pod with label '%s'. Using '%s'", StatusPodLabel, podList.Items[0].GetName())
+		StatusPodName = podList.Items[0].GetName()
+	default:
+		StatusPodName = podList.Items[0].GetName()
 	}
 }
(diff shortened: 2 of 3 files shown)


pkg/plugin/aggregation/status.go
```diff
@@ -82,9 +82,12 @@ func GetStatus(client kubernetes.Interface, namespace string) (*Status, error) {
 		return nil, errors.Wrap(err, "sonobuoy namespace does not exist")
 	}
 
+	// Determine sonobuoy pod name
+	SetStatusPodName(client, namespace)
+
 	pod, err := client.CoreV1().Pods(namespace).Get(StatusPodName, metav1.GetOptions{})
 	if err != nil {
-		return nil, errors.Wrap(err, "could not retrieve sonobuoy pod")
+		return nil, errors.New("could not retrieve sonobuoy pod")
```
johnSchnake: Why the change in the error? Seems like that would just obfuscate issues by not returning the actual error to the user.

pkg/plugin/aggregation/update.go
```diff
@@ -157,3 +161,24 @@ func GetPatch(annotation string) map[string]interface{} {
 		},
 	}
 }
+
+// Set the sonobuoy master pod name
+// The sonobuoy master pod is identified by the label 'run=sonobuoy-master`
+// This function retrieves the pod name based on that label and sets it
+// provided there is only one pod returned with that label
+func SetStatusPodName(client kubernetes.Interface, namespace string) {
```
johnSchnake: nit: godoc comment should start with `SetStatusPodName`.

pkg/plugin/aggregation/update.go
```diff
@@ -157,3 +161,24 @@ func GetPatch(annotation string) map[string]interface{} {
...
+
+	listOptions := metav1.ListOptions{
+		LabelSelector: "run=sonobuoy-master",
```
johnSchnake: Add as a constant at the top of the document.

pkg/plugin/aggregation/update.go
```diff
@@ -157,3 +161,24 @@ func GetPatch(annotation string) map[string]interface{} {
...
+	}
+	podList, err := client.CoreV1().Pods(namespace).List(listOptions)
+	if err != nil {
+		logrus.Errorf("Error listing pods with label 'run:sonobuoy-master': %s", err)
```
johnSchnake: nit, reference the to-be-created constant for the label

pkg/plugin/aggregation/update.go
```diff
@@ -157,3 +161,24 @@ func GetPatch(annotation string) map[string]interface{} {
...
+	} else {
+		if len(podList.Items) != 1 {
+			logrus.Error("More than one pod with label 'run:sonobuoy-master'")
```
johnSchnake: I'd modify this to
 - use the constant value for the label
 - inform the user as to the consequence; in this situation, It seems like we should choose a reasonable default which I think should be the first item in the list (rather than leaving the default as-is, which may not be one of the choices).
 - be sure to handle the `err==nil && len(items)==0` case. I'm not certain if k/k returns empty list w/o error or not but I'd just be defensive.

Also, I'd adjust the flow control a bit to simplify the indenting here:
```
// get list
if err != nil {
  log error and return
}
switch{
case len(items) == 0 :
  error and state we'll use the default value (and show it to user)
case len(items) > 1:
  log situation as a warning (I wouldn't use error) and choose the first one in the list (and let user know the choice)
default:
  choose first (and only item) without any logging
}
```

pkg/plugin/aggregation/status.go
lander2k2: To test this error I created a second pod with the 'run:sonobuoy-master' label in the same namespace.

This is the error if we don't make this change:
```
$ sonobuoy status
ERRO[0001] More than one pod with label 'run:sonobuoy-master'
ERRO[0001] error attempting to run sonobuoy: could not retrieve sonobuoy pod: pods "sonobuoy" not found
```

This seemed a little confusing to me b/c the user will wonder "why are you looking for a pod called 'sonobuoy' - my pod has a different name?"

This happens because the `SetStatusPodName` function doesn't set the pod name if there's an error and it reverts back to the default name "sonobuoy".  (I'm wondering now if this was wise.)

This is the error output if we *do* make this change:
```
$ sonobuoy status
ERRO[0006] More than one pod with label 'run:sonobuoy-master'
ERRO[0006] error attempting to run sonobuoy: could not retrieve sonobuoy pod
```
With the both errors provided together, I think this is clearer.  But open to suggestions.

pkg/plugin/aggregation/update.go
lander2k2: Thanks for the detailed suggestion @johnSchnake.  I've added a commit that addresses this.

pkg/plugin/aggregation/status.go
lander2k2: Update for clarity: Due to other changes based on other review items...

There is no longer an error if there are multiple pods with the `run:sonobuoy-master` label.  The first pod is selected (with a warning) and the status command runs.

If there are *no* pods with the label the output is now as follows:
```
$ sonobuoy status
ERRO[0001] No pods found with label 'run=sonobuoy-master' in namespace heptio-sonobuoy, using default pod name 'sonobuoy'
ERRO[0001] error attempting to run sonobuoy: could not retrieve sonobuoy pod
```

johnSchnake:
Looks great. I restarted the test run; it failed due to an upstream flakiness (known issue).

If you could just rebase onto master and squash down to one commit I'll merge this once it goes green on CI.

lander2k2:
Done and done :).  Thanks for reviewing!

johnSchnake:
💯

PR #788, opened 2019-07-01
New query options to get pod logs

MengmengZHANG:
Enhance query options to get pod logs, currently we support only LimitSize and LimitTime. Will support other limit options, e.g. adding timestamps to the logs.

- Fixes https://github.com/heptio/sonobuoy/issues/777

1) deprecate two fields:
- LimitSize
- LimitTime

Because:
- they have different format and name comparing to k8s.io/api/core/v1/PodLogOptions
- they have string type, which requires manually parsing and validating them.

2) Add more parameters, with proper type, to let json.unmarshall do the parsing and validating logic

**Release note**:
```
```


pkg/discovery/pods.go
```diff
@@ -36,31 +36,42 @@ const (
 	PodLogsLocation = "podlogs"
 )
 
-// Only set values if they have values greater than 0 (as in they user specified).
-func getPodLogOptions(container string, limitBytes, sinceSeconds int64) *v1.PodLogOptions {
+// mapping to k8s.io/api/core/v1/PodLogOptions.
+func getPodLogOptions(cfg *config.Config) *v1.PodLogOptions {
```
johnSchnake: I think keeping the `container` in the signature is a good idea so that it is clear to any caller that that extra piece of information is needed. Otherwise you have to set it manually afterwards (which you do)

johnSchnake: Well...hm... I see this prevented you from regenerating the rest of the options set when iterating over the containers.

Lets just check what happens on k8s side if you leave the container empty in this; would it still be usable and default to the first container?

As long as it is usable I'm ok with it.

pkg/discovery/queries.go
```diff
@@ -260,7 +261,8 @@ func filterResources(gvrs map[schema.GroupVersion][]metav1.APIResource, namespac
 }
 
 // QueryPodLogs gets the pod logs for each pod in the given namespace.
-func QueryPodLogs(kubeClient kubernetes.Interface, recorder *QueryRecorder, ns string, cfg *config.Config) error {
+func QueryPodLogs(kubeClient kubernetes.Interface, recorder *QueryRecorder, ns string, cfg *config.Config,
+	podLogOptions *v1.PodLogOptions) error {
```
johnSchnake: Lets not add this new parameter; they can get it just from the cfg so it doesnt add any new info.

pkg/discovery/discovery.go
```diff
@@ -144,13 +144,15 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 		QueryServerData(kubeClient, recorder, cfg),
 	)
 
+	podLogOptions := getPodLogOptions(cfg)
+	logrus.Infoln("PogLogOptions: ", podLogOptions)
```
johnSchnake: nit Remove

pkg/config/config.go
```diff
@@ -228,7 +269,8 @@ func (c SizeOrTimeLimitConfig) TimeLimitDuration(defaultVal time.Duration) time.
 	return val
 }
 
-func (c SizeOrTimeLimitConfig) timeLimitDuration() (val time.Duration, defaulted bool, err error) {
+// Deprecated: use PodLogLimits.SinceSeconds instead
```
johnSchnake: Just calling out the deprecation logic:

I think the way you did it here is fine; doing dual support for a time and then removing it altogether after another version or two.

Often, we move a little faster and are OK with breaking changes since we have a relatively short support window but I am not against this careful step since we are changing the config structure itself.

Can you open an issue stating that we need to remove these fields in v0.16? I think that is enough time to let it sit there.

site/docs/master/sonobuoy-config.md
```diff
@@ -54,7 +54,14 @@ Filters
    - A Kubernetes [label selector][labelselector] which will be added to every query run.
 
 Limits
- - Options for limiting the size of the pod logs (in bytes) or the how far back in time to gather logs (in seconds). These will be passed onto Kubernetes [PodLogOptions][podlogopts]
-
+ - Options for limiting limits the scope of response when getting logs from pods. These will be passed onto Kubernetes [PodLogOptions][podlogopts]
```
johnSchnake: 💯 For updating the docs :)

Can you also update https://github.com/heptio/sonobuoy/blob/master/site/docs/master/sonobuoy-config.md#query-options ?

johnSchnake:
Looks good. I will try and run it manually a few times as another sanity check while you make your updates.

Can you also squash down to a single commit describe these changes?
And don't forget to sign your commit (using `git amend --signoff`)

pkg/config/config.go
MengmengZHANG: Done. https://github.com/heptio/sonobuoy/issues/789

site/docs/master/sonobuoy-config.md
johnSchnake: Oh I misread; then you were spot on. I was thinking you were updating the master readme, but you accurately updated the master version of this doc. 👍 Sorry for any confusion I made.

johnSchnake:
Looks great. If you can just rebase onto master and squash down to one commit I can merge.

My flow is:
```
git fetch upstream && git rebase -i upstream/master
```
Then change all but the first commit to just be a "fixup" (or "f") rather than "pick".

If you want to reword the first commit too you can change that first prefix to be "r" for reword and fixup your commit message.

💯 on this, thanks for the contribution.

MengmengZHANG:
Rebase done

PR #791, opened 2019-07-02
Fix preflight test error message

waynr:

It fixes a confusing error message in the preflight checks. It's not strictly
needed, but it is nice as a user when error messages make sense.

There is currently no issue for this problem.

**Release note**:
```
Fix dns preflight test error message
```

### DIFF
--- a/pkg/client/preflight.go
+++ b/pkg/client/preflight.go
@@ -18,6 +18,7 @@ package client
 
 import (
 	"fmt"
+	"strings"
 
 	version "github.com/hashicorp/go-version"
 	"github.com/heptio/sonobuoy/pkg/buildinfo"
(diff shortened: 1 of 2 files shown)


johnSchnake:
💯 That is definitely confusing.

Bonus points if you're interested in expanding it to explain to the user what labels it was looking for/where. Although the message would now be more clear, I can imagine they'd immediately know what Sonobuoy considers a "dns pod" especially if they are experimenting with their own deployment or dns solution.

johnSchnake:
CI will fail with this:
```
--- FAIL: TestDNSCheck (0.00s)
    --- FAIL: TestDNSCheck/Requires_at_least_one_pod (0.00s)
        preflight_test.go:168: Expected error to be "no dns pod tests found" but got "no dns pods found with the labels [foo] in namespace kube-system"
FAIL
```
I think everything else should be fine. After you push the unit test fix and CI is green I'll merge.

Thanks for the contribution!

johnSchnake:
Sorry for the miscue; that second commit should be squashed into the first so that it is signed as well, otherwise the sign off check fails.

johnSchnake:
💯 Thanks for the quick iterations on those. Merging since I validated the unit test signal locally.

waynr:
@johnSchnake thanks for the quick reviews!

PR #795, opened 2019-07-03
Add redirect for docs

jonasrosland:
This fixes the need to specify `master` or `v0.15.0` or similar when linking to the latest docs from the README and other sources.

It will automatically redirect https://sonobuoy.io/docs to the latest version, as specified in _config.yml.

It also redirects any links specified in _data/shortlinks.yml, to their latest counterpart.
For example, https://sonobuoy.io/docs/gen will redirect to the latest version, currently https://sonobuoy.io/docs/v0.15.0/gen.

This also fixes the need to specify the latest version in the navigation header.

Signed-off-by: jonasrosland <jrosland@vmware.com>

### DIFF
--- a/site/_includes/site-header.html
+++ b/site/_includes/site-header.html
@@ -7,7 +7,7 @@
             <li class="home"><a href="/" title="{{ site.title }}">Home</a></li>
             <li class="blog"><a href="/blog" title="Blog Posts">Blog</a></li>
             <li class="community"><a href="/community" title="Community">Community</a></li>
-            <li class="docs"><a href="/docs/v0.15.0">Documentation</a></li>
+            <li class="docs"><a href="/docs">Documentation</a></li>
             <li class="resources"><a href="/resources" title="Resources">Resources</a></li>
         </ul>
         <div class="nav-mobile-link">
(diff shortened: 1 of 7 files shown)


site/_includes/site-header.html
```diff
@@ -7,7 +7,7 @@
             <li class="home"><a href="/" title="{{ site.title }}">Home</a></li>
             <li class="blog"><a href="/blog" title="Blog Posts">Blog</a></li>
             <li class="community"><a href="/community" title="Community">Community</a></li>
-            <li class="docs"><a href="/docs/v0.15.0">Documentation</a></li>
+            <li class="docs"><a href="/docs">Documentation</a></li>
```
johnSchnake: Also need to adjust the URL for the footer: https://github.com/heptio/sonobuoy/pull/795/files#diff-0a55ba953b4a00424ce1150c8cb5e4efR17

johnSchnake:
One small modification needed; then I think it is good to go.

Some of the front-end code is over my head but it just came from a 3rd party, commonly used repo (https://github.com/avillafiorita/jekyll-datapage_gen) and seems to work as expected.

Thanks for taking this on @jonasrosland ; I had no idea that getting redirects to work would have taken a PR of this size.

site/_includes/site-header.html
jonasrosland: Thanks for catching that, I've updated the PR.

PR #796, opened 2019-07-03
Allow cleanup to be skipped on a per plugin basis

### LINKED ISSUE #299: RFE: Debug scenarios, block on failure... don't cleanup
There exist conditions where operators want to block on a failure condition and not cleanup.  

This would: 
Fail a given test and potentially block on the failed test, and not continue to other tests. 

Blocking would allow an operator or developer to then jump in to determine the cause of the failure.

zubron:
This PR Introduces a new field to the plugin definition which allows
cleanup for a given plugin to be skipped. This can be used in the
case where a user wants to inspect the resources after a plugin
has finished running.
 
If this field is set to true, cleanup for that plugin will be skipped.
If it is omitted or set to false, cleanup will happen as usual.

- Fixes #299 

This change will also require documentation to be updated
which I haven't yet done.

**Release note**:
```
Introduced a new plugin configuration option which allows clean up of resources for that plugin to be skipped.
```

### DIFF
--- a/pkg/plugin/aggregation/run_test.go
+++ b/pkg/plugin/aggregation/run_test.go
@@ -22,6 +22,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
 	k8stesting "k8s.io/client-go/testing"
 )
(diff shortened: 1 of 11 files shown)


pkg/plugin/aggregation/run_test.go
```diff
@@ -197,3 +198,87 @@ func getTestCert() (*tls.Certificate, error) {
 		PrivateKey:  privKey,
 	}, nil
 }
+
+type MockCleanupPlugin struct {
```
johnSchnake: Why create a whole new plugin struct here instead of just referencing jobs or daemonset plugin types?

zubron: I started out using a job plugin type but realised that taking that approach would require this test to have knowledge of the internals of the job's `Cleanup` method in order to check that it was called (by checking that the [`CleanedUp` field](https://github.com/heptio/sonobuoy/blob/master/pkg/plugin/driver/job/job.go#L179) was set). For this test, I only wanted to check that the call was skipped correctly. Given that the `Cleanup` function takes a slice of `plugin.Interface`, it seemed less brittle to provide a mock implementation of that interface rather than rely on the implementation of `job.Cleanup`.

johnSchnake: > knowledge of the internals of the job's Cleanup method

You're referring to the `cleanup` field?

That is somewhat minor IMO but since it isn't part of the interface that seems fair enough. Not a blocker on my end and it does end up being a very isolated test, so that's good.

johnSchnake:
Another `/run_e2e.sh: line 65: wait: `': not a pid or valid job spec` problem in CI.

I restarted the build.

PR #800, opened 2019-07-06
Add FieldSelectors and LabelSelector for GetPogLogs

### LINKED ISSUE #778: Query selectors
We currently allow a label selector and a list of namespaces to be used to specify which podlogs to gather.

It makes sense to also allow a fieldselector to give more flexibility.

The logic for the log gathering is currently:
 - get a list of namespaces to query later https://github.com/heptio/sonobuoy/blob/d34e9cf1558766473e312526e887c0559915af0d/pkg/discovery/discovery.go#L102
 - create a listOptions object using the labelselector https://github.com/heptio/sonobuoy/blob/d34e9cf1558766473e312526e887c0559915af0d/pkg/discovery/queries.go#L275-L289
 - list those pods and get logs for every container with the given podlogoptions https://github.com/heptio/sonobuoy/blob/ee6ecc84a101343434290b029ea3949d6ac575bd/pkg/discovery/pods.go#L56

This new fieldSelector would just enable power users to do more, but I don't want the design to complicate things if a user doesn't want to use it.

MengmengZHANG:
Add more control parameters to PodLogLimits
```
type PodLogLimits struct {
    Namespaces string 
    SonobuoyNamespace *bool 
    FieldSelectors []string 
    LabelSelector string
    ...
}
```
When collecting pod logs, the scope of pods is defined by:
>  (Namespaces OR SonobuoyNamespace OR FieldSelectors) AND LabelSelector

Then for each pod, the format and size of logs is defined by other fields, e.g. SinceSeconds

- Fixes #778

SonobuoyNamespace's default value is true, which causes some problem when merging the user's input config and the default config. A workaround is implemented to avoid this issue:
- use *bool to differentiate input = nil and input= false
- only nil input will be overwritten by default config
- Before the merge, save user's input, and restore a non-nil user's input if the merge library has overwritten it.

Hopefully the merge library provides a better solution for this [open issue]( https://github.com/imdario/mergo/issues/89)

To test:
- sonobuoy gen config > test.config
- remove plugins from test.config: "Plugins": []
- adapt the provided parameters, e.g. FieldSelectors
- sonobuoy run --sonobuoy-image=xxx --config test.config --wait --image-pull-policy=Always


cmd/sonobuoy/app/gen.go
```diff
@@ -298,9 +298,22 @@ func mergeConfigs(dst, src *config.Config) {
 		emptyResources = true
 	}
 
+	// Workaround to differentiate a false value and a nil value
+	// Only override dst.Limits.PodLogs.SonobuoyNamespace when it's nil
+	// See https://github.com/imdario/mergo/issues/89
+	var sonobuoyNamespace *bool
+	if dst.Limits.PodLogs.SonobuoyNamespace != nil {
+		sonobuoyNamespace = new(bool)
+		*sonobuoyNamespace = *dst.Limits.PodLogs.SonobuoyNamespace
```
johnSchnake: TL;DR: 👍 

Long message: 

Frustrating to see something like this have to get an awkward implementation like this but I do understand the problem. For all the great stuff `mergo.Merge` can do I guess simple things like this make it clear that even though the idea is conceptually simple, the devil is in the details.

It may be fine as-is, but it could be resolved if we changed the field to something that inverted the default value. E.g. if we had `SkipSonobuoyNamespace` which defaulted to false, that would remove the need to do this workaround since we'd treat a nil object the same as the default, right? But that is more awkward to a user which is a higher priority than the simplicity of the implementation on something like this. :-/

pkg/config/config.go
```diff
@@ -168,8 +169,51 @@ type LimitConfig struct {
...
 // PodLogLimits limits the scope of response when getting logs from pods.
-// It exposes a subset of k8s.io/api/core/v1/PodLogOptions.
+// The scope of pods is defined by (Namespaces OR SonobuoyNamespace OR FieldSelectors) AND LabelSelector
+// Then for each pod, the format and size of logs is defined by other fields, e.g. SinceSeconds
 type PodLogLimits struct {
+	// A regular expression for the targeted namespaces, follows the format of Golang Regexp
+	Namespaces string `json:"Namespaces" mapstructure:"Namespaces"`
+
+	// If set to true, take into account the pod logs from Sonobuoy's namespace, i.e. Config.Namespace
+	SonobuoyNamespace *bool `json:"SonobuoyNamespace" mapstructure:"SonobuoyNamespace"`
```
johnSchnake: This is a `*bool`, is that just so that if nil we use the default value (which is true)? Looking at the definition like this does make it seem odd. (5% more of me is wondering if we should make that change to this field so the default is false and the field represents to _skip_ the sonobuoy namespace)

pkg/discovery/discovery.go
```diff
@@ -183,6 +191,32 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
...
+		if len(nsfilter) > 0 {
+			nsfilter = fmt.Sprintf("%s|%s", nsfilter, cfg.Namespace)
+		} else {
+			nsfilter = cfg.Namespace
+		}
+	}
+	if len(nsfilter) > 0 {
+		nsListLogs, _ := FilterNamespaces(kubeClient, nsfilter)
+		for _, ns := range nsListLogs {
+			cfg.Limits.PodLogs.FieldSelectors = append(cfg.Limits.PodLogs.FieldSelectors, "metadata.namespace="+ns)
```
johnSchnake: Let's not have a method which will mutate the cfg like this.
 - it makes the signature confusing since you don't know what it really intends to do to the object
 - it will make it permanently unclear if a config object you have is ready to be used or if you need to call this method to get it ready.

Instead, I'd take this method and have it return a list of field selectors to run queries on.

pkg/discovery/pods.go
```diff
@@ -63,19 +63,25 @@ func getPodLogOptions(cfg *config.Config) *v1.PodLogOptions {
 }
 
 // gatherPodLogs will loop through collecting pod logs and placing them into a directory tree
-func gatherPodLogs(kubeClient kubernetes.Interface, ns string, opts metav1.ListOptions, cfg *config.Config) error {
+func gatherPodLogs(kubeClient kubernetes.Interface, ns string, opts metav1.ListOptions, cfg *config.Config,
```
johnSchnake: I'd modify the comment to explain why you're adding the `visitedPods` field (e.g. that this may be run in a namespace with varying listoptions which might overlap).

tiny nit: I'd modify it to be `map[string]struct{}` :
 - makes it clear that we don't really want a map of `true|false` values but that we are just using the map as a list of names which can be checked quickly
 - saves a tiny bit of space (not that there is a memory concern here really but just FYI)

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
+        - **Namespaces**: string
+            - A regular expression for the targeted namespaces, following the format of Golang Regexp. 
+            - Default is empty string
+            - To get logs from all namespaces use ".*"
+        - **SonobuoyNamespace**: bool
+            - If set to true, take into account the pod logs from Sonobuoy's namespace
+            - Default value is true
+            - Valid inputs: true | false | null
+        - **FieldSelectors**: []string
+            - A list of filed selects, with OR logic.
```
johnSchnake: s/filed selects/field selectors

pkg/discovery/discovery.go
johnSchnake: In addition, this method should definitely have a set of tests to cover it.

I don't think it needs to take the `kubeClient` either. It seems like we should just be able to do a few things:
 - given a config, get the namespace filter (a combination of Namespaces and SonobuoyNamespace)
 - given the config object and a list of namespaces, generate a list of field selectors to query

Will comment at the calling site about this too to try and make it clear.

pkg/discovery/discovery.go
```diff
@@ -151,10 +150,19 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
...
-		trackErrorsFor("querying pod logs under namespace " + ns)(
-			QueryPodLogs(kubeClient, recorder, ns, cfg),
-		)
+	// query pod logs
+	if cfg.Resources == nil || sliceContains(cfg.Resources, "podlogs") {
+		preprocessPodLogLimits(cfg, kubeClient)
+		nsListLogs, _ := FilterNamespaces(kubeClient, ".*")
+		for _, ns := range nsListLogs {
+			trackErrorsFor("querying pod logs under namespace " + ns)(
+				QueryPodLogs(kubeClient, recorder, ns, cfg),
```
johnSchnake: This flow is pretty unintuitive; I see what you're doing now that I've looked into the implementation of each function but it shouldn't require that.

I think it would be more clear (and testable) to do the following:

 - use the config to get a list of namespaces to query for logs
   - use that list of namespaces in the `for` loop
 - within the `QueryPodLogs` options, generate the necessary field selectors for the query. This is somewhat inefficient since you'll do it each time you query a namespace but the code will be more readable, we'll get rid of methods that mutate the config, and you'll have a few simple-to-test functions for getting the namespaces/field selectors from a static piece of data. 
   - in the existing implementation we also end up filtering the namespaces inside the preprocessing function and right afterwards; the extra API call is probably more expensive than iterating over our struct a few more times to make those field selectors.

cmd/sonobuoy/app/gen.go
```diff
@@ -298,9 +298,22 @@ func mergeConfigs(dst, src *config.Config) {
...
+	}
+
 	// Provide defaults but don't overwrite any customized configuration.
 	mergo.Merge(dst, src)
+
 	if emptyResources {
 		dst.Resources = []string{}
 	}
+	if sonobuoyNamespace != nil {
+		dst.Limits.PodLogs.SonobuoyNamespace = sonobuoyNamespace
```
johnSchnake: Add a testcase to ensure this workaround works and gets maintained.

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
-    - SinceSeconds
-    - SinceTime
-    - Timestamps
-    - TailLines
-    - LimitBytes
+ - Options for limiting limits the scope of response.
+ - **Limits.PodLogs** limits the scope when getting logs from pods. 
+     - The scope of pods is defined by:
+       ```
+       Namespaces OR SonobuoyNamespace OR FieldSelectors) AND LabelSelector
```
johnSchnake: I think you meant to start this with an `(`

cmd/sonobuoy/app/gen.go
MengmengZHANG: I have discussed this with @zubron, our concern is that setting it in inverted way will break the logic:
`(Namespaces OR SonobuoyNamespace OR FieldSelectors`

If user set:
- Namespaces = ".*"
- SkipSonobuoyNamespace = true

The thing becomes:
`(Namespaces OR FieldSelectors) AND SkipSonobuoyNamespace`, which :
- adds one more layer of different operator, and maybe confusing for the user.
- need to have specific code to check `SkipSonobuoyNamespace`, cannot just expand it to append to the fieldSelectors list. So we lost the generic way to handle the quick buttons as additional fieldSelectors, and the logic becomes a lot of if...else...

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
- - The supported parameters are:
-    - Previous
-    - SinceSeconds
-    - SinceTime
-    - Timestamps
-    - TailLines
-    - LimitBytes
+ - Options for limiting limits the scope of response.
+ - **Limits.PodLogs** limits the scope when getting logs from pods. 
+     - The scope of pods is defined by:
```
johnSchnake: Slightly confusing; I'd be a little more verbose to make it clear (I know you spell out each field below, but this is the first thing they'll see and shows how those fields relate). Maybe even just remove this section to clarify down below in each field (going to put a few suggestions there)

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
+       ```
+       Namespaces OR SonobuoyNamespace OR FieldSelectors) AND LabelSelector
+       ```
+     - The supported parameters are:
+        - **Namespaces**: string
+            - A regular expression for the targeted namespaces, following the format of Golang Regexp.
```
johnSchnake: nit: I'd remove `following the format of Golang Regexp`. There are a few things a golang regexp cant do but I don't think they really apply much here and can just confuse the user into thinking `golang regexp` is something different than normal regexp.

site/docs/master/sonobuoy-config.md
MengmengZHANG: got it.

site/docs/master/sonobuoy-config.md
MengmengZHANG: got it

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
+ - **Limits.PodLogs** limits the scope when getting logs from pods. 
+     - The scope of pods is defined by:
+       ```
+       Namespaces OR SonobuoyNamespace OR FieldSelectors) AND LabelSelector
+       ```
+     - The supported parameters are:
+        - **Namespaces**: string
+            - A regular expression for the targeted namespaces, following the format of Golang Regexp. 
+            - Default is empty string
+            - To get logs from all namespaces use ".*"
```
johnSchnake: 💯 to the suggestion here to help users in a common use case.

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
+        - **SonobuoyNamespace**: bool
+            - If set to true, take into account the pod logs from Sonobuoy's namespace
```
johnSchnake: Not clear on a first read; suggestion:
```
If set to true, ensures pod logs are gathered in the namespace Sonobuoy is running in. Can be set along with a `Namespaces` field or on its own.
```

pkg/config/config.go
MengmengZHANG: The default value is handled at:

pkg/config/config.go: func New() *Config {
        ...
        cfg.Namespace = DefaultNamespace
        cfg.Limits.PodLogs.SonobuoyNamespace = new(bool)
	*cfg.Limits.PodLogs.SonobuoyNamespace = true
}

A nil value will be overwritten by this default value when doing config merge.
The same logic as cfg.Namespace, whose default value is ".*"

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
+            - Default value is true
+            - Valid inputs: true | false | null
```
johnSchnake: Why do we want to let the user specify the null value? Is there a reason or it all has to do with the boolean-defaults-to-true issue?

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
+        - **FieldSelectors**: []string
+            - A list of filed selects, with OR logic. 
+            - For example, to get logs from two specified namespaces
+            `FieldSelectors = ["metadata.namespace==default","metadata.namespace==heptio-sonobuoy"]` 
+            - Each field selector contains one or more chained operators, with AND logic
+            - For example, to get logs from a specified pod 
+            `FieldSelectors = ["metadata.namespace==default,metadata.name==pod1"]` 
+            - Each field selector follows the same format as
```
johnSchnake: 💯 for the docs here.

cmd/sonobuoy/app/gen.go
johnSchnake: Good point on the example 
 - Namespaces = ".*"
 - SkipSonobuoyNamespace = true

That is very confusing looking and seems like you're explicitly going to _not_ get the logs from the Sonobuoy namespace.

sgtm

pkg/discovery/discovery.go
MengmengZHANG: The thing is, it's a bit difficult to get the targeted namespaces, which could be defined in FieldSelectors. Then we need to parse the FieldSelectors ourselves, validating it, take out the selector which contains "metadata.namespace=xx", and append to the targeted namespace list. 

Comparing to all these additional effort, the current solution is will not take time more than O(N), and all invalidate namespace will be filtered out when doing list of pods query. We will only proceed to get pod logs for valid pods.

site/docs/master/sonobuoy-config.md
MengmengZHANG: okay, just trying to say that goland use something like ".*" while other language maybe use "*"

site/docs/master/sonobuoy-config.md
MengmengZHANG: I just want to say user can either not provide this field or provide this field with a null value. We can remove it since the null pointer is just a coding implementation, no need to expose to user.

pkg/discovery/discovery.go
MengmengZHANG: looking into https://github.com/kubernetes/client-go/issues/410

MengmengZHANG: After discussion, here is what we are proposing:
1) get a list of namespaces by: Namespaces regex and SonobuoyNamespace bool
2) for each namespace, get all pod logs clientset.CoreV1().Pods(ns).List(...)
3) then do another query with field selectors, get pod logs clientset.CoreV1().Pods("").List(field selectors)

Then the OR logic will be fulfilled:
`(Namespaces OR SonobuoyNamespace OR FieldSelectors) `

pkg/discovery/discovery.go
MengmengZHANG: The method could take input of config, and output a regex for the targeted namespaces

site/docs/master/sonobuoy-config.md
MengmengZHANG: understood

pkg/discovery/discovery.go
MengmengZHANG: unit testing done.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -24,6 +24,10 @@ import (
 	"github.com/heptio/sonobuoy/pkg/plugin"
 )
 
+const (
+	RawInput = "not nil"
```
johnSchnake: nit: no reason to export this

cmd/sonobuoy/app/gen_test.go
```diff
@@ -276,9 +279,104 @@ func TestResolveConfig(t *testing.T) {
...
+			if len(tc.cliInput) > 0 {
+				fs := GenFlagSet(tc.input, EnabledRBACMode)
+				if err := fs.Parse(strings.Split(tc.cliInput, " ")); err != nil {
+					t.Fatalf("Failed to parse CLI input %q: %v", tc.cliInput, err)
+				}
+			}
+
+			conf := tc.input.resolveConfig()
+
+			if *conf.Limits.PodLogs.SonobuoyNamespace != *tc.expected.Limits.PodLogs.SonobuoyNamespace {
```
johnSchnake: nit: You're defining the test as if it is going to test any/all fields of the config and also test config vs CLI interactions.

If here you're just testing the one section of the config you can get rid of the CLI input field/parsing and remove lots of the structure on the resulting object (no need to define `expected` to be the entire struct if you just want to confirm the `*bool` value.

Likewise, you could make it less verbose by not defining a whole genconfig object for eadch test if you just want to test the effect of changing the `SonobuoyNamespace` or `PodLogLimits`

pkg/discovery/discovery.go
```diff
@@ -154,13 +154,22 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
...
 	// query pod logs
 	if cfg.Resources == nil || sliceContains(cfg.Resources, "podlogs") {
-		preprocessPodLogLimits(cfg, kubeClient)
-		nsListLogs, _ := FilterNamespaces(kubeClient, ".*")
-		for _, ns := range nsListLogs {
-			trackErrorsFor("querying pod logs under namespace " + ns)(
-				QueryPodLogs(kubeClient, recorder, ns, cfg),
-			)
+
+		// Eliminate duplicate pods when query by namespaces and query by filedSelectors
```
johnSchnake: s/filedSelectors/fieldSelectors

pkg/discovery/discovery_test.go
```diff
@@ -0,0 +1,56 @@
...
+func TestGetPodLogNamespaceFilterWithEmptyInput(t *testing.T) {
+	// input
+	cfg := config.New()
+	cfg.Namespace =  config.DefaultNamespace
+	*cfg.Limits.PodLogs.SonobuoyNamespace = false
+	cfg.Limits.PodLogs.Namespaces = ""
+	// expected
+	expectedNsFilter := ""
+
+	nsFilter := getPodLogNamespaceFilter(cfg)
```
johnSchnake: Combine these three tests into a table-driven test (with each case being a subtest).

pkg/discovery/queries.go
```diff
@@ -257,7 +257,12 @@ func filterResources(gvrs map[schema.GroupVersion][]metav1.APIResource, namespac
 }
 
 // QueryPodLogs gets the pod logs for each pod in the given namespace.
-func QueryPodLogs(kubeClient kubernetes.Interface, recorder *QueryRecorder, ns string, cfg *config.Config) error {
+// If namespace is not provided, get pod logs from field selectors
+// visitedPods will eliminate duplicate pods when execute overlapping queried,
+// e.g. query by namespaces and query by filedSelectors
```
johnSchnake: s/filedSelectors/fieldSelectors

pkg/discovery/queries.go
```diff
@@ -257,7 +257,12 @@ func filterResources(gvrs map[schema.GroupVersion][]metav1.APIResource, namespac
 }
 
 // QueryPodLogs gets the pod logs for each pod in the given namespace.
-func QueryPodLogs(kubeClient kubernetes.Interface, recorder *QueryRecorder, ns string, cfg *config.Config) error {
+// If namespace is not provided, get pod logs from field selectors
+// visitedPods will eliminate duplicate pods when execute overlapping queried,
```
johnSchnake: s/queried/queries

pkg/discovery/queries.go
```diff
@@ -257,7 +257,12 @@ func filterResources(gvrs map[schema.GroupVersion][]metav1.APIResource, namespac
 }
 
 // QueryPodLogs gets the pod logs for each pod in the given namespace.
-func QueryPodLogs(kubeClient kubernetes.Interface, recorder *QueryRecorder, ns string, cfg *config.Config) error {
+// If namespace is not provided, get pod logs from field selectors
```
johnSchnake: I'd end this with a period `.`; it doesn't seem to flow into the next line as a sentence.

tiny nit: I think it reads better to say "pod logs using field selectors" instead of "from field selectors"

pkg/discovery/queries.go
```diff
@@ -269,15 +274,21 @@ func QueryPodLogs(kubeClient kubernetes.Interface, recorder *QueryRecorder, ns s
...
-		opts.FieldSelector = filedSelector
+	if len(ns) > 0 {
+		logrus.Infof("Collecting Pod Logs by namespace (%v)", ns)
 		err := gatherPodLogs(kubeClient, ns, opts, cfg, visitedPods)
 		if err != nil {
 			return err
 		}
+	} else {
+		logrus.Infoln("Collecting Pod Logs by FieldSelectors", cfg.Limits.PodLogs.FieldSelectors)
+		for _, filedSelector := range cfg.Limits.PodLogs.FieldSelectors {
```
johnSchnake: s/filedSelector/fieledSelector or maybe even s/filedSelector/fs since it is such a small context.

MengmengZHANG:
I have squashed the commits into one, and retest all the possible test cases, working fine. I have also provided the commands that I used to test in the pull request summary.

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
+        `FieldSelectors = ["metadata.namespace==default,metadata.name==pod1"]` 
+        - Each field selector follows the same format as
+        `k8s.io/apimachinery/pkg/apis/meta/v1/types/ListOptions/FieldSelector`
+        - Can be set along with a Namespaces/SonobuoyNamespace field or on its own.
+    - **LabelSelector**: string
+        - Filtering candidate pods by their labels
+        - Using the same format as
+        `k8s.io/apimachinery/pkg/apis/meta/v1/types/ListOptions/LabelSelector`
+        - For example:
+        `FieldSelecto = "app=nginx,layer in (frontend, backend)"`
```
stevesloka: nit: `FieldSelecto` misspelling

MengmengZHANG:
There is one single "=" instead of "==". The site/docs/master/sonobuoy-config.md shows examples of the format:
"FieldSelectors": ["metadata.namespace=default"],

stevesloka:
Are these docs wrong then? I copy/pasted: https://github.com/heptio/sonobuoy/pull/800/files#diff-8f0acbcbc04c588fc62d68fff9d8a5fcR69

MengmengZHANG:
sorry it's a typo

> Are these docs wrong then? I copy/pasted: https://github.com/heptio/sonobuoy/pull/800/files#diff-8f0acbcbc04c588fc62d68fff9d8a5fcR69

site/docs/master/sonobuoy-config.md
```diff
@@ -54,14 +54,45 @@ Filters
...
+        - A regular expression for the targeted namespaces. 
+        - Default is empty string
+        - To get logs from all namespaces use ".*"
+    - **SonobuoyNamespace**: bool
+        - If set to true, get pod logs from the namespace Sonobuoy is running in. Can be set along with a `Namespaces` field or on its own.
+        - Default value is true
+    - **FieldSelectors**: []string
+        - A list of filed selectors, with OR logic. 
+        - For example, to get logs from two specified namespaces
+        `FieldSelectors = ["metadata.namespace==default","metadata.namespace==heptio-sonobuoy"]`
```
MengmengZHANG: typo, use single =

stevesloka:
@MengmengZHANG  I took out the double equals and have the same result, still got logs for all namespaces.

MengmengZHANG:
> @MengmengZHANG I took out the double equals and have the same result, still got logs for all namespaces.

Seems @stevesloka doesn't use the new container image, will check offline.

stevesloka:
All seems to work @MengmengZHANG, just need to clean up those docs.

site/docs/master/sonobuoy-config.md
MengmengZHANG: in fact it's LabelSelector, resolved.

MengmengZHANG:
Thanks @stevesloka , I have updated the docs. Could you merge the pull request?

PR #801, opened 2019-07-08
Ensure only failed tests are run with --rerun-failed

### LINKED ISSUE #797: sonobuoy e2e --rerun-failed command runs the entire test suite
1. Ran a sonobuoy test that resulted in some failures.
2. Downloaded the results tarball using `sonobuoy retrieve`
3. Deleted the heptio-sonobuoy namespace
4. Used the following command to re-run the failed tests: `sonobuoy e2e 201907031649_sonobuoy_85d03238-77c9-4c5e-8df7-6bf91e8d6a58.tar.gz --rerun-failed`

This resulted in a new deployment of sonobuoy running all the conformance tests:

```sh
# k logs sonobuoy-e2e-job-7a7439fae526457d e2e
+ set +x
+ /usr/local/bin/ginkgo '--focus=\[Conformance\]' '--skip=Alpha|\[(Disruptive|Feature:[^\]]+|Flaky)\]' --noColor=true /usr/local/bin/e2e.test -- --disable-log-dump --repo-root=/kubernetes --provider=local --report-dir=/tmp/results --kubeconfig=
+ tee /tmp/results/e2e.log
I0703 18:57:36.596085      20 test_context.go:406] Using a temporary kubeconfig file from in-cluster config : /tmp/kubeconfig-979058637
I0703 18:57:36.596399      20 e2e.go:241] Starting e2e run "bb773da5-079d-4f0b-b354-c6083908bb43" on Ginkgo node 1
Running Suite: Kubernetes e2e suite
===================================
Random Seed: 1562180254 - Will randomize all specs
Will run 215 of 4411 specs
```

Sonobuoy should only run the tests that failed.

zubron:
This reintroduces the `e2e --rerun-failed` behaviour which had been
accidentally removed as part of a refactor. As part of reintroducing this
behaviour, it was noticed that the generated rergex for running specific
tests was invalid and would cause a panic when running ginkgo. In the
generated regex, we were double escaping `[]` characters. This doesn't
seem to be necessary as we do not double escape these characters for
other regexes (e.g. [`defaultSkipList`](https://github.com/heptio/sonobuoy/blob/master/pkg/client/mode.go#L42)).

- Fixes #797 

This behaviour was removed in [#315](https://github.com/heptio/sonobuoy/pull/315/files#diff-f6f1cbdfc63048c2340177ccd708d6b6L108). I have verified the fix works manually but I don't know if there is a better way to test this so that is covered under CI. I wasn't sure if it was worth adding another more testing logic to our Travis CI script so I would appreciate any input that you might have.

**Release note**:
```
Fixed a bug where the `--rerun-failed` flag for the `e2e` command would run all the tests, not only the failed tests.
```

### DIFF
--- a/pkg/client/e2e.go
+++ b/pkg/client/e2e.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -75,11 +76,9 @@ func (*SonobuoyClient) GetTests(reader io.Reader, show string) ([]reporters.JUni
 // Focus returns a value to be used in the E2E_FOCUS variable that is
 // representative of the test cases in the struct.
 func Focus(testCases []reporters.JUnitTestCase) string {
-	// YAML doesn't like escaped characters and regex needs escaped characters. Therefore a double escape is necessary.
-	r := strings.NewReplacer("[", `\\[`, "]", `\\]`)
 	testNames := make([]string, len(testCases))
 	for i, tc := range testCases {
-		testNames[i] = r.Replace(tc.Name)
+		testNames[i] = regexp.QuoteMeta(tc.Name)
 	}
 	return strings.Join(testNames, "|")
 }
--- a/cmd/sonobuoy/app/e2e.go
+++ b/cmd/sonobuoy/app/e2e.go
@@ -127,6 +127,11 @@ func e2es(cmd *cobra.Command, args []string) {
 	}
 
 	fmt.Printf("Rerunning %d tests:\n", len(testCases))
+
+	// Ensure that we run the failed test cases and don't skip any
+	runCfg.E2EConfig.Focus = client.Focus(testCases)
+	runCfg.E2EConfig.Skip = ""
+
 	if err := sonobuoy.Run(runCfg); err != nil {
 		errlog.LogError(errors.Wrap(err, "error attempting to rerun failed tests"))
 		os.Exit(1)
--- a/pkg/client/e2e_test.go
+++ b/pkg/client/e2e_test.go
@@ -119,3 +119,47 @@ func TestString(t *testing.T) {
 		})
 	}
 }
+
+func TestFocus(t *testing.T) {
+	testCases := []struct {
+		desc   string
+		cases  PrintableTestCases
+		expect string
+	}{
+		{
+			desc:   "No test should result in an empty string",
+			cases:  []reporters.JUnitTestCase{},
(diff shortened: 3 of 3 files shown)


pkg/client/e2e.go
```diff
@@ -75,8 +75,7 @@ func (*SonobuoyClient) GetTests(reader io.Reader, show string) ([]reporters.JUni
 // Focus returns a value to be used in the E2E_FOCUS variable that is
 // representative of the test cases in the struct.
 func Focus(testCases []reporters.JUnitTestCase) string {
-	// YAML doesn't like escaped characters and regex needs escaped characters. Therefore a double escape is necessary.
-	r := strings.NewReplacer("[", `\\[`, "]", `\\]`)
+	r := strings.NewReplacer("[", `\[`, "]", `\]`)
```
johnSchnake: The fact that this escaping changed may be a result of how we changed some yaml serialization. We used to manually create the plugins but I modified that to just use the go-yaml library so I think we changed how we were quoting things, therefore probably how things needed to be internally escaped.

johnSchnake:
So I thought the manual test was going to be a formallity but it actually failed:
```
+ /usr/local/bin/ginkgo '--focus=\[k8s.io\] Pods should be submitted and removed \[Conformance\]' --skip= --noColor=true /usr/local/bin/e2e.test -- --disable-log-dump --repo-root=/kubernetes --provider=local --report-dir=/tmp/results --kubeconfig=
+ set +x
+ tee /tmp/results/e2e.log
...
...
===================================
Random Seed: 1562599276 - Will randomize all specs
Will run 0 of 3585 specs
...
```

So the regexp we make there doesn't actually run any specs. Will tinker for a bit on this and get back to you.

zubron:
Oh, that's disappointing :( I wonder if the `.` in the test name prefix is impacting it and also needs to be escaped. I think the test case that I tried it with had the prefix `[sig-apps]`.

johnSchnake:
So actually I am now frustrated and relieved :)

I think it works fine. I do think that the `.` could be escaped but it isn't probably going to change any selection. I would worry about some more generic problems though and think it would be wise to go ahead and use https://golang.org/pkg/regexp/#QuoteMeta instead of our own, simple replacer.

The problem _I_ ran into was that to test this I grabbed a random tarball, forced a failure in the log, then tried to rerun it. The issue was that the test name for the test changed between that [older] tarball and the tests that run against my cluster. So it legitimately didn't match any test. No fundamental problem.

PR #803, opened 2019-07-13
Pin images used in kind config and reduce number of workers

### LINKED ISSUE #767: CI should pin version of Kind images we're using
We have recently pinned the version of Kind to a specific version, but we still just do `kind cluster create` and instead we should be specifying the version of cluster to use instead of relying on the default.

We also, in our kind config, specify 3 workers. It may slightly speed up CI to reduce it to just 2 workers. We'd still have a multi-cluster node and a reasonable test of the systems involved (Reducing it to 1 worker or 0 workers and just a master may not be as full a test).

zubron:

- Fixes #767

### DIFF
--- a/kind-config.yaml
+++ b/kind-config.yaml
@@ -2,5 +2,7 @@ kind: Cluster
 apiVersion: kind.sigs.k8s.io/v1alpha3
 nodes:
 - role: control-plane
+  image: kindest/node:v1.15.0
 - role: worker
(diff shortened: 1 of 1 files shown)


zubron:
> Approved after merging/rebasing the change for #804 ; that was the one you really intended to go in first, right?

These changes could be merged in either order, but I'm pretty sure rebasing will be required either way.

PR #807, opened 2019-07-17
Create a UUID for loaded configs if necessary

### LINKED ISSUE #806: Using a config without a UUID results in an error when writing results
Originally reported by @MengmengZHANG, it was noticed that if you provide a config to sonobuoy without a UUID, it will fail to collect the results for the run. This behaviour can be demonstrated by running `sonobuoy gen` and then modifying the `config.json` data within the `sonobuoy-config-cm` and removing the `UUID` field:

```
apiVersion: v1
data:
  config.json: |
    {"Description":"DEFAULT","Version":"v0.15.0","ResultsDir":"/tmp/sonobuoy","Resources":["apiservices","certificatesigningrequests","clusterrolebindings","clusterroles","componentstatuses","configmaps","controllerrevisions","cronjobs","customresourcedefinitions","daemonsets","deployments","endpoints","ingresses","jobs","leases","limitranges","mutatingwebhookconfigurations","namespaces","networkpolicies","nodes","persistentvolumeclaims","persistentvolumes","poddisruptionbudgets","pods","podlogs","podsecuritypolicies","podtemplates","priorityclasses","replicasets","replicationcontrollers","resourcequotas","rolebindings","roles","servergroups","serverversion","serviceaccounts","services","statefulsets","storageclasses","validatingwebhookconfigurations","volumeattachments"],"Filters":{"Namespaces":".*","LabelSelector":""},"Limits":{"PodLogs":{"Namespaces":"","SonobuoyNamespace":true,"FieldSelectors":[],"LabelSelector":"","Previous":false,"SinceSeconds":null,"SinceTime":null,"Timestamps":false,"TailLines":null,"LimitBytes":null,"LimitSize":"","LimitTime":""}},"QPS":30,"Burst":50,"Server":{"bindaddress":"0.0.0.0","bindport":8080,"advertiseaddress":"","timeoutseconds":10800},"Plugins":[{"name":"systemd-logs"}],"PluginSearchPath":["./plugins.d","/etc/sonobuoy/plugins.d","~/sonobuoy/plugins.d"],"Namespace":"heptio-sonobuoy","WorkerImage":"gcr.io/heptio-images/sonobuoy:v0.15.0","ImagePullPolicy":"IfNotPresent","ImagePullSecrets":""}
kind: ConfigMap
metadata:
  labels:
    component: sonobuoy
  name: sonobuoy-config-cm
  namespace: heptio-sonobuoy
```

Using this manifest (running `kubectl apply -f <generated_manifest.yml>`), the following error will eventually be shown in the output from `sonobuoy logs`:

```
time="2019-07-17T16:31:13Z" level=error msg="error assembling results tarball: archive/tar: write too long"
time="2019-07-17T16:31:13Z" level=info msg="Results available at /tmp/sonobuoy/201907171631_sonobuoy_.tar.gz"
time="2019-07-17T16:31:13Z" level=info msg="no-exit was specified, sonobuoy is now blocking"
```
Retrieving these results will appear to be successful, but the resulting tarball will contain a tarball with the same name which is invalid because the write failed. We can see from the logs that the tarball name is missing the UUID: `201907171631_sonobuoy_.tar.gz`. Within [`discovery.Run`](https://github.com/heptio/sonobuoy/blob/master/pkg/discovery/discovery.go#L65), the UUID is used to create a unique directory to store the results in. Since the UUID is empty, this is not generated. Later, a [tarball is created within this directory](https://github.com/heptio/sonobuoy/blob/master/pkg/discovery/discovery.go#L186-L187), and the directory is added to the tarball which results in the tarball being added to itself and causes the write error.

In previous versions of Sonobuoy (up to v0.14.2), the loaded config was merged with a [default config](https://github.com/heptio/sonobuoy/blob/v0.14.2/pkg/config/loader.go#L35). This meant that if a UUID was not provided, it would be generated. This behaviour is now a regression, sonobuoy should still generate a UUID if not provided in the config.

**Environment:**
- Sonobuoy version: v0.15.0
- Kubernetes version: (use `kubectl version`): v.1.15.0 (client and server)

zubron:
When loading a config without a UUID, we don't generate one and it
remains empty. It is necessary for the config to have a UUID as it is
used to create a uniquely named directory for storing results. Without
this, we don't correctly create a child directory to store the results.
We create a tarball within a directory that we also try to add to that
tarball, effectively adding the created tarball to itself resulting in a
write error.

- Fixes #806

**Release note**:
```
Fixed a bug where a config without a UUID would result in an error when writing results
```

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -27,6 +27,7 @@ import (
 	"github.com/heptio/sonobuoy/pkg/plugin"
 	pluginloader "github.com/heptio/sonobuoy/pkg/plugin/loader"
 	"github.com/pkg/errors"
+	uuid "github.com/satori/go.uuid"
 )
 
 const (
@@ -77,6 +78,12 @@ func LoadConfig() (*Config, error) {
 	// Make the results dir overridable with an environment variable
 	if resultsDir, ok := os.LookupEnv("RESULTS_DIR"); ok {
 		cfg.ResultsDir = resultsDir
+	}
+
+	// If the loaded config doesn't have its own UUID, create one
+	if cfg.UUID == "" {
+		cfgUuid, _ := uuid.NewV4()
+		cfg.UUID = cfgUuid.String()
 	}
 
 	// 5 - Load any plugins we have
--- a/pkg/config/loader_test.go
+++ b/pkg/config/loader_test.go
@@ -118,16 +118,35 @@ func TestSaveAndLoad(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// And we can't predict what advertise address we'll detect
+	// We can't predict what advertise address we'll detect
 	cfg2.Aggregation.AdvertiseAddress = cfg.Aggregation.AdvertiseAddress
-	// And UUID's won't match either
-	cfg.UUID = ""
-	cfg2.UUID = ""
 
 	if !reflect.DeepEqual(cfg2, cfg) {
 		t.Fatalf("Defaults should match but didn't \n\n%v\n\n%v", cfg2, cfg)
 	}
+}
 
+func TestLoadConfigSetsUUID(t *testing.T) {
+	cfg := New()
+	cfg.UUID = ""
+
+	if blob, err := json.Marshal(&cfg); err == nil {
+		if err = ioutil.WriteFile("./config.json", blob, 0644); err != nil {
+			t.Fatalf("Failed to write default config.json: %v", err)
+		}
+		defer os.Remove("./config.json")
+	} else {
+		t.Fatalf("Failed to serialize %v", err)
+	}
+
+	loadedCfg, err := LoadConfig()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if loadedCfg.UUID == "" {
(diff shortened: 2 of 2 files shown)


pkg/config/loader.go
```diff
@@ -79,6 +80,12 @@ func LoadConfig() (*Config, error) {
 		cfg.ResultsDir = resultsDir
 	}
 
+	// If the loaded config doesn't have its own UUID, create one
```
MengmengZHANG: in the old version of sonobuoy, the UUID is generated by `cfg := New()`; and in the current version the `cfg := &Config{}`; is it better to just use `cfg := New()`, so that all the default values, not only the UUID will be generated? Or it can break some other logic?

zubron: In the [PR which introduced this change](https://github.com/heptio/sonobuoy/pull/725), it seems that the intent was to prevent other default values being introduced, particularly `Resources`. It looked to me like the setting of `UUID` was something that was missed in that change. Hopefully, @johnSchnake can give us a bit more insight into the intended behaviour here when he's back from vacation.

Issue #306, opened 2018-02-28
Generalize plugin results reporting


chuckha:
The first pass of e2e is only solving for the common case of conformance e2e tests.

We would ideally like to reuse this command to fit any e2e-esque set of tests (such as heptio's extended e2e tests).

johnSchnake:
I think the answer is to generalize and rebrand the command in general.

I'm going to be writing up a design doc for this but it seems like a really good place for improvement this cycle.

Right now my thinking is along the lines of wanting to get out status codes and 1-2 different types of results parsing (e.g. junit). This would probably be done via some extra data the plugin can drop to help sonobuoy know where a certain piece of information is. TBD.

PR #810, opened 2019-07-23
Remove plugin.Definition

johnSchnake:
The type plugin.Definition completely duplicates fields that exist
inside of manifest.Manifest and adds nothing new. Though I prefer
the name plugin.Definition, the users currently use the manifest.Manifest
object locally and expect the nested sonobuoy-config object so we
needed to keep that in order to not break existing plugins. In addition,
the plugin.Definition type was missing the Driver field, so the server
was actually losing that data when using that type.

This change should have no impact other than clarifying the landscape
of types on the backend.

Helps facilitate #306

**Release note**:
```
NONE
```

### DIFF
--- a/pkg/plugin/interface.go
+++ b/pkg/plugin/interface.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"path"
 
-	"github.com/heptio/sonobuoy/pkg/plugin/manifest"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 )
@@ -60,16 +59,6 @@ type Interface interface {
 	GetName() string
 	// SkipCleanup returns whether cleanup for this plugin should be skipped or not.
 	SkipCleanup() bool
-}
-
-// Definition defines a plugin's features, method of launch, and other
-// metadata about it.
-type Definition struct {
-	Name         string
-	ResultType   string
-	SkipCleanup  bool
-	Spec         manifest.Container
-	ExtraVolumes []manifest.Volume
 }
 
 // ExpectedResult is an expected result that a plugin will submit.  This is so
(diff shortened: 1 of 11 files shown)


pkg/plugin/manifest/manifest.go
```diff
@@ -24,10 +24,20 @@ import (
...
+
+	// Name is the user-facing name for the plugin.
+	PluginName string `json:"plugin-name"`
+
+	// ResultType should uniquely identify the plugin and is used by the aggregator
+	// to track what results should be reported back to it.
+	ResultType string `json:"result-type"`
+
+	// SkipCleanup informs Sonobuoy to leave the pods created for this plugin running,
+	// after the run completes instead of deleting them as part of default, cleanup behavior.
```
johnSchnake: Added comments here since I had started adding them on the plugin.Definition in my other branch related to #306

pkg/plugin/loader/loader_test.go
```diff
@@ -235,7 +235,7 @@ func TestFilterList(t *testing.T) {
 		{Name: "test3"},
 	}
 
-	expected := []*manifest.Manifest{definitions[0]}
+	expected := []manifest.Manifest{definitions[0]}
```
johnSchnake: In a bunch of situations, we were using pointers when I don't think it was necessary (and we had gone from a non-pointer plugin.Definition to a pointer to a manifest.Manifest).

pkg/plugin/interface.go
```diff
@@ -62,16 +61,6 @@ type Interface interface {
 	SkipCleanup() bool
 }
 
-// Definition defines a plugin's features, method of launch, and other
-// metadata about it.
-type Definition struct {
```
johnSchnake: This is the main thing I just removed here; it is just a flattened version of the `manifest.Manifest` except it doesnt have the `Driver` field. It leads to us having an extra layer to transfer data through and doesn't provide any clear benefit.

PR #811, opened 2019-07-24
Remove use of global sonobuoy pod name

zubron:
Previously the sonobuoy pod name was stored in a global variable and was
updated before it was used. Now, instead of updating the global variable
the function that determines the pod name returns it instead.

**Release note**:
```
NONE
```

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -96,9 +96,6 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 		}
 	}
 
-	// Determine sonobuoy pod name
-	pluginaggregation.SetStatusPodName(kubeClient, cfg.Namespace)
-
 	// Set initial annotation stating the pod is running. Ensures the annotation
 	// exists sooner for user/polling consumption and prevents issues were we try
 	// to patch a non-existant status later.
@@ -110,8 +107,8 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 	)
 
 	// 2. Get the list of namespaces and apply the regex filter on the namespace
-	logrus.Infof("Filtering namespaces based on the following regex:%s",  cfg.Filters.Namespaces)
-	nslist, err := FilterNamespaces(kubeClient,  cfg.Filters.Namespaces)
+	logrus.Infof("Filtering namespaces based on the following regex:%s", cfg.Filters.Namespaces)
+	nslist, err := FilterNamespaces(kubeClient, cfg.Filters.Namespaces)
 	if err != nil {
 		errlog.LogError(errors.Wrap(err, "could not filter namespaces"))
 		return errCount + 1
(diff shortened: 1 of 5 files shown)


pkg/discovery/discovery.go
```diff
@@ -243,6 +240,12 @@ func setStatus(client kubernetes.Interface, namespace string, status *pluginaggr
 		return errors.Wrap(err, "failed to marshal the patch")
 	}
 
-	_, err = client.CoreV1().Pods(namespace).Patch(pluginaggregation.StatusPodName, types.MergePatchType, patchBytes)
+	// Determine sonobuoy pod name
+	podName, err := pluginaggregation.GetStatusPodName(client, namespace)
+	if err != nil {
+		return errors.Wrap(err, "failed to get name of status pod")
```
johnSchnake: `status pod` I think is a little confusing.

Suggestion `failed to get the name of the aggregator pod to set the status on`

pkg/plugin/aggregation/status.go
```diff
@@ -83,9 +83,12 @@ func GetStatus(client kubernetes.Interface, namespace string) (*Status, error) {
 	}
 
 	// Determine sonobuoy pod name
-	SetStatusPodName(client, namespace)
+	podName, err := GetStatusPodName(client, namespace)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get name of status pod")
```
johnSchnake: same naming comment

pkg/plugin/aggregation/update.go
```diff
@@ -163,26 +166,25 @@ func GetPatch(annotation string) map[string]interface{} {
...
+func GetStatusPodName(client kubernetes.Interface, namespace string) (string, error) {
 	listOptions := metav1.ListOptions{
 		LabelSelector: StatusPodLabel,
 	}
 
 	podList, err := client.CoreV1().Pods(namespace).List(listOptions)
 	if err != nil {
-		logrus.Errorf("Error listing pods with label '%s': %s", StatusPodLabel, err)
-		return
+		return "", errors.Wrap(err, "Error listing pods with label %q")
```
johnSchnake: Should wrap with lower case phrases; only start with a capital when logging since you know it wont be further wrapped/prefixed.

pkg/plugin/aggregation/update_test.go
```diff
@@ -47,3 +54,75 @@ func TestCreateUpdater(t *testing.T) {
 		t.Errorf("expected status to be failed, got %v", updater.status.Status)
 	}
 }
+
+func TestGetStatusPodName(t *testing.T) {
```
johnSchnake: 👍 having the test. I just talked to someone who was suggesting we tweak the logic this method would use to take into account pod status (filter to only consider running pods). Something to consider, but regardless of the decision there, having the test will help us ensure that functionality.

pkg/plugin/aggregation/update.go
zubron: Thanks for pointing that out. I had taken this from the previous log line and forgot to update that.

pkg/plugin/aggregation/update_test.go
zubron: I'm happy to make that change but would prefer to do it in a follow up branch just to avoid adding another change in functionality here.

johnSchnake: Oh for sure; I didnt mean to do that here. Was just adding context for why I like having this test now (that it will immediately be useful for testing/adding requested functionality)

pkg/discovery/discovery.go
johnSchnake: There are a few other lines like this that I stopped commenting it on; fyi (so you dont just look at the lines I pointed out)

PR #812, opened 2019-07-24
Write plugin definitions in their respective folders

johnSchnake:
The plugin info used to be written in the config but that was
removed and also didn't have any of the information like sessionID,
fields that could be altered by the config, etc.

This change dumps the loaded configuration data into the folder
at which users will most likely look for it.

In order to avoid an import cycle, it also removes a type that was
not used anywhere.

Separated this small bit of logic from my larger PR on making
plugin results more clearly reported.

Related to #306

**Release note**:
```
Adds a `definition.json` file for each plugin run in its respective `plugin/<name>` folder in the results tarball, making it easier to see what plugin information was used to produce the output and other bits of useful data like the session ID (useful if looking for resources or in logs) and whether or not the plugin cleaned up successfully.
```


pkg/discovery/discovery.go
```diff
@@ -200,6 +213,23 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
...
+func dumpPlugin(p plugin.Interface, outputDir string) error {
+	b, err := json.Marshal(p)
+	if err != nil {
+		return errors.Wrapf(err, "encoding plugin %v definition to yaml", p.GetName())
+	}
+
+	err = ioutil.WriteFile(
+		filepath.Join(outputDir, results.PluginsDir, p.GetName(), pluginDefinitionFilename),
+		b,
+		os.FileMode(0666),
```
zubron: Minor point, but would 644 be enough for the permissions here? That's the permissions that other files in the resulting tarball use.

johnSchnake: 👍 Thanks for pointing that out; I just chose a random value and thought I'd come back to it and forgot. That would be fine for me.

PR #813, opened 2019-07-25
Post-process results to standard format and provide generic results printing command

### LINKED ISSUE #306: Generalize plugin results reporting

johnSchnake:
Adds a post-processing step where we walk the files the plugin produced and dump info into a single file whose format we define.

For junit, we record each tests and it's pass/fail status. For raw files (since we may not know how large they are) we just record each file location.

This file (which now contains all tests/output files) can be parsed by clients in order to more easily view plugin results.

The plugin manifests have new fields `ResultFormat` and `ResultFile` to inform this post-processing step.

Fixes #306

**Release note**:
```
TBD
```


Makefile
```diff
@@ -24,7 +24,7 @@ REGISTRY ?= gcr.io/heptio-images
 IMAGE = $(REGISTRY)/$(TARGET)
 DIR := ${CURDIR}
 DOCKER ?= docker
-LINUX_ARCH := amd64 arm64
+LINUX_ARCH := amd64
```
johnSchnake: Dont commit this; it is just to iterate faster so it doesnt build the others.

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if input.detailed {
+		printResultsDetails(obj, input)
+	} else {
+		printResultsSummary(obj)
+	}
+	return nil
+}
+
+func printResultsDetails(o *results.Item, input resultsInput) {
```
johnSchnake: All these `result` output stuff is tbd. It wasn't too terribly defined in the design doc (focus was on server side stuff). We can iterate a bit on it after we finalize the other stuff.

cmd/sonobuoy/app/root.go
```diff
@@ -53,6 +53,9 @@ func NewSonobuoyCommand() *cobra.Command {
 	cmds.AddCommand(NewCmdRun())
 	cmds.AddCommand(NewCmdGenPlugin())
 	cmds.AddCommand(NewCmdImages())
+	cmds.AddCommand(NewCmdResults())
+	// Debug only
+	cmds.AddCommand(NewSchnake())
```
johnSchnake: Don't commit; just my own debug command.

pkg/client/results/doc.go
```diff
@@ -17,3 +17,22 @@ limitations under the License.
 // Package results provides a low level API to extract data from a Sonobuoy
 // result archive.
 package results
+
+// Item is the central format for plugin results. Various plugin
+// types can be transformed into this simple format and set at a standard
+// location in our results tarball for simplified processing by any consumer.
+type Item struct {
```
johnSchnake: Not sure why this ended up in `doc.go`; I'll move it to somewhere more appropriate.

pkg/client/results/processing.go
```diff
@@ -0,0 +1,169 @@
...
+	// if another can not be determined.
+	StatusUnknown = "unknown"
+
+	// PostProcessedResultsFile is the name of the file we create when doing
+	// postprocessing on the plugin results.
+	PostProcessedResultsFile = "sonobuoy_results.yaml"
+)
+
+type postProcessor func(string) (Item, error)
+type fileSelector func(string, os.FileInfo) bool
```
johnSchnake: Just interesting to note that github doesn't properly do syntax highlighting here.

pkg/client/results/processing.go
```diff
@@ -0,0 +1,169 @@
...
+
+// aggregateStatus defines the aggregation rules for status. Failures bubble
+// up and otherwise the status is assumed to pass.
+func aggregateStatus(items ...Item) string {
+	for _, v := range items {
+		if len(v.Items) > 0 {
+			v.Status = aggregateStatus(v.Items...)
```
johnSchnake: This will definitely need tests and currently doesn't either work as intended or wasn't called in the right place. Last I checked I had some empty values that shouldn't have been empty.

pkg/client/results/processing.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+	return StatusPassed
+}
+
+func PostProcessPlugin(p plugin.Interface, dir string) (Item, error) {
+	fmt.Printf("checking types %+v", p, p.GetResultFormat())
+	switch p.GetResultFormat() {
+	case "junit", "e2e":
+		fmt.Println("got junit item")
```
johnSchnake: Lots of debug items thoughout; reminder to remove them all.

pkg/client/results/processing.go
```diff
@@ -0,0 +1,169 @@
...
+	default:
+		fmt.Println("default item?")
+
+	}
+	return Item{}, nil
+}
+
+func processPluginWithProcessor(p plugin.Interface, baseDir string, processor postProcessor, selector fileSelector) (Item, error) {
+	pdir := path.Join(baseDir, PluginsDir, p.GetName())
+	_, isDS := p.(*daemonset.Plugin)
```
johnSchnake: Just special handling down below for daemonsets; default to the other handling for jobs and any "other" types that could in theory exist (since the someone using this as a library could be making their own type for the plugin interface).

pkg/client/results/processing_test.go
```diff
@@ -0,0 +1,64 @@
...
+	}{
+		{
+			desc:   "Job junit 1 file",
+			key:    "job-junit-1",
+			plugin: getPlugin("job", "junit", xmlFile),
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			item, err := ProcessRawPlugin(tc.plugin, tc.dir)
```
johnSchnake: Originally written when this method existed; I dont think it even does any more.

pkg/client/results/raw.go
```diff
@@ -0,0 +1,36 @@
...
+// processRaw will return an Item object with the File value set to the path in question. If the
+// file is unable to be stat'd then the status of the Item is StatusFailed (StatusPassed otherwise).
+func processRawFile(rawPath string) (Item, error) {
+	status := StatusPassed
+	if _, err := os.Stat(rawPath); err != nil {
+		status = StatusFailed
+	}
+
+	return Item{
+		Name:   rawPath,
```
johnSchnake: Make this `filepath.Base` of the path. Even though it would be duplicate info, may want to add the `metadata.file` with the full path to be consistent with how the junit results set that field.

pkg/client/results/reader.go
```diff
@@ -295,3 +297,57 @@ func ConfigFile(version string) string {
...
+	/*if resultObj.Aggregate.File != "" {
+		var buf bytes.Buffer
+		wantPath := filepath.Join(PluginsDir, plugin, resultObj.Aggregate.File)
+		q.Q("looking for", wantPath) // DEBUG
+		err := r.WalkFiles(func(curPath string, info os.FileInfo, err error) error {
+			if curPath == wantPath {
+				q.Q("found", wantPath) // DEBUG
+			}
+			q.Q(curPath, wantPath) // DEBUG
+			return ExtractBytes(wantPath, curPath, info, &buf)
```
johnSchnake: Delete old code

pkg/discovery/discovery.go
```diff
@@ -241,6 +258,117 @@ func getPodLogNamespaceFilter(cfg *config.Config) string {
...
+	for i, t := range junitResults.TestCases {
+		status := "unknown"
+		switch {
+		case t.Skipped == nil && t.FailureMessage == nil:
+			status = "passed"
+		case t.FailureMessage != nil:
+			resultObj.Status = "failed"
+			status = "failed"
+		case t.Skipped != nil:
+			status = "skipped"
```
johnSchnake: Should be using constants

pkg/plugin/aggregation/run_test.go
```diff
@@ -239,6 +239,10 @@ func (cp *MockCleanupPlugin) SkipCleanup() bool {
 	return cp.skipCleanup
 }
 
+func (cp *MockCleanupPlugin) GetResultFile() string {
+	return ""
+}
```
johnSchnake: Just necessary to add since I added the new method to the interface.

Trivial for people who want to update the library to add a similar method.

pkg/client/results/junit.go
```diff
@@ -52,3 +57,46 @@ func Passed(testCase reporters.JUnitTestCase) bool {
...
 }
+
+func processJunitFile(junitPath string) (Item, error) {
+	junitResults := reporters.JUnitTestSuite{}
+
+	// Passed unless a failure is encountered for aggregate status.
+	resultObj := Item{
+		Name:     filepath.Base(junitPath),
+		Status:   StatusPassed,
+		Metadata: map[string]string{"file": junitPath},
```
johnSchnake: Be sure to add metadata to the raw output stuff too

johnSchnake: Should be relative path instead of full path?

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,167 @@
...
+	for _, v := range o.Items {
+		if v.File == "" {
+			fmt.Printf("%v|%v\n", v.Name, v.Status)
+		} else {
+			// load file and print it
+			r, cleanup, err := getReader(input.archive)
+			if err != nil {
+				fmt.Println("Schnake err", err)
```
johnSchnake: logrus

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,188 @@
...
+	r := results.NewReaderWithVersion(gzr, results.VersionTen)
+	return r, func() { gzr.Close(); f.Close() }, nil
+}
+
+func result(input resultsInput) error {
+	r, cleanup, err := getReader(input.archive)
+	defer cleanup()
+	if err != nil {
+		errlog.LogError(err)
+		os.Exit(1)
```
zubron: This error should be returned so that the exit can be handled by the `Run` function in the command.

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	obj, err := r.PluginResults(input.plugin)
```
zubron: This error isn't being checked/returned.

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,188 @@
...
+	if input.detailed {
+		printResultsDetails("", obj, input)
+	} else {
+		printResultsSummary(*obj)
+	}
+	return nil
+}
+
+func printResultsDetails(prefix string, o *results.Item, input resultsInput) {
```
zubron: If you wanted to test this functionality a bit more easily, maybe you could pass a writer to this function which the formatted output could be written to rather than trying capture what is on stdout.

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,188 @@
...
+	found := false
+	s := []byte{}
+	err := r.WalkFiles(
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if path == filepath {
+				found = true
+				freader := info.Sys().(io.Reader)
```
zubron: Should we be checking that the type assertion was successful here?

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,188 @@
...
+func walkForSummary(o results.Item, p, f, s int, failList []string) (numPassed, numFailed, numSkipped int, failed []string) {
+	if len(o.Items) > 0 {
+		for _, v := range o.Items {
+			p, f, s, failList = walkForSummary(v, p, f, s, failList)
+		}
+		return p, f, s, failList
+	}
+
+	switch o.Status {
+	case "passed":
```
zubron: The `Status*` constants could be used here.

pkg/client/results/junit.go
```diff
@@ -52,3 +58,51 @@ func Passed(testCase reporters.JUnitTestCase) bool {
...
+	// Passed unless a failure is encountered for aggregate status.
+	resultObj := Item{
+		Name:     filepath.Base(currentFile),
+		Status:   StatusPassed,
+		Metadata: map[string]string{"file": relPath},
+	}
+
+	infile, err := os.Open(currentFile)
+	if err != nil {
+		resultObj.Status = StatusFailed
```
zubron: I'm not sure if the status should be `Failed` if we fail to open the file or decode it as there could be non-test related reasons e.g. incorrect file permissions or something. Perhaps `StatusUnknown` would work for these cases? I don't feel very strongly about it though, so if you had particular reasons for using that status that's fine.

pkg/client/results/processing.go
```diff
@@ -0,0 +1,181 @@
...
+	// Only pass if no failures found.
+	return StatusPassed
+}
+
+func PostProcessPlugin(p plugin.Interface, dir string) (Item, error) {
+	i := Item{}
+	var err error
+
+	switch p.GetResultFormat() {
+	case "junit", "e2e":
```
zubron: Perhaps these could be const values? Also, will there be any list of supported result types?

johnSchnake: Added constants; didnt exactly add a "list" explicitly, but the supported types are just the ones that we have constants for. Our documentation will reference those.

pkg/client/results/reader.go
```diff
@@ -295,3 +297,57 @@ func ConfigFile(version string) string {
...
+	resultObj := &Item{}
+
+	found := false
+	err := r.WalkFiles(
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil || found == true {
+				return err
+			}
+			// TODO(chuckha) consider reusing this function for any generic e2e-esque plugin results.
+			// TODO(chuckha) consider using path.Join()
```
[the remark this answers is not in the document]
johnSchnake: The copied code was just used as the starting place; comments removed.

The code is now used in a general sense to get any result file for a given plugin.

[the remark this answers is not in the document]
johnSchnake: I just left the original code/comment alone since it is tied into the `e2e` command which has a bunch of other things that make it unique.

cmd/sonobuoy/app/results.go
johnSchnake: Yea, I could do that. I opted to just use `Example...` tests which tests stdout output automatically. I lost the ability to do a table driven test, but for just testing formatting of a CLI command I think that is OK.

johnSchnake:
Codecov doesnt seem to handle force pushing onto branches too well; hard to navigate to see new coverage.

Coverage details here: https://codecov.io/gh/heptio/sonobuoy/commit/2d768fab5a88d6a8cd3ca7b128a65bd215dcec21
Currently at 61% coverage

pkg/plugin/loader/loader.go
```diff
@@ -66,11 +66,13 @@ func LoadAllPlugins(namespace, sonobuoyImage, imagePullPolicy, imagePullSecrets
 		if err != nil {
 			return []plugin.Interface{}, errors.Wrapf(err, "couldn't load plugin definition file %v", file)
 		}
+		fmt.Printf("Loaded file: %+v\n\n", string(definitionFile))
```
johnSchnake: These shouldn't have gotten added; remove

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,255 @@
...
+
+	obj = getItemInTree(obj, input.node)
+	if obj == nil {
+		return fmt.Errorf("node named %q not found", input.node)
+	}
+
+	switch input.mode {
+	case resultModeDetailed:
+		printResultsDetails([]string{}, obj, input)
+	case resultModeDump:
```
johnSchnake: Didnt add the logic here; intended to just print the file to stdout

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,239 @@
...
+			// already visible on the object.
+			treePath = append(treePath, o.Name)
+			fmt.Printf("%v ", strings.Join(treePath, "|"))
+			io.Copy(os.Stdout, filereader)
+		}
+	}
+}
+
+func printResultsSummary(o *results.Item) {
+	fmt.Printf("Plugin: %v\nStatus:%v\n", o.Name, o.Status)
```
johnSchnake: missing space after status

johnSchnake:
Changes to point out in recent pushes:
 - modified json serialization of `Item` type to omitempty on a few fields and use lower case. Tests modified accordingly
 - modified the `WalkFiles` logic to mirror that of typical `filepath.WalkFunc` logic: processing should stop when an error is encountered. This allows us to stop processing when we find the file we want and we can return an `io.Reader`. This allows us to (1) reuse that logic in a few places (2) be more efficient instead of loading the whole item into a byte slice.
 - modified some names in a few test cases to be more clear
 - added logic for `--mode dump` to dump the whole results file
 - fixed issue where the json objects in a junit, detailed result set had their own name as part of their `metadata.path` value (duplicating the value in the `name` field)

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,240 @@
...
+
+	if o.File == "" {
+		// Print each leaf node as a json object. Add the path as a metadata field for access by the end user.
+		if o.Metadata == nil {
+			o.Metadata = map[string]string{}
+		}
+		o.Metadata["path"] = strings.Join(treePath, "|")
+		b, err := json.Marshal(o)
+		if err != nil {
+			logrus.Errorf("Error marshalling item to json: %v", err)
```
zubron: By logging these errors rather than returning them and letting them bubble up, the error from `result` will be `nil` which will affect the overall exit code. Is this intended?

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,240 @@
...
+			return
+		}
+		fmt.Println(string(b))
+	} else {
+		// Load file with a new reader since we can't assume this reader has rewind
+		// capabilities.
+		r, cleanup, err := getReader(input.archive)
+		if err != nil {
+			logrus.Errorf("Failed to read archive to get file %v: %v", o.File, err)
```
zubron: `getReader` can still return a `cleanup` function when it returns an error. If we return here, we miss the cleanup call below so it should also be made here.

pkg/client/results/testdata/mockResults/plugins/ds-junit-01/ds-junit-01.golden.json
```diff
@@ -0,0 +1 @@
...:"passed","meta":null,"file":"","items":[{"name":"global","status":"passed","meta":null,"file":"","items":[{"name":"output.xml","status":"passed","meta":{"file":"results/global/output.xml"},"file":"","items":[{"name":"[k8s.io] Pods should be submitted and removed [NodeConformance] [Conformance]","status":"passed","meta":null,"file":"","items":null},{"name":"[sig-node] ConfigMap should fail to create ConfigMap with empty key [Conformance]","status":"passed","meta":null,"file":"","items":null},{"name":"[sig-storage] Downward API volume should set DefaultMode on files [LinuxOnly] [NodeConformance] [Conformance]","status":"passed","meta":null,"file":"","items":null},{"name":"[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: dir-link-bindmounted] [Testpattern: Dynamic PV (default fs)] subPath should support existing directories when readOnly specified in the volumeSource","status":"skipped","meta":null,"file":"","items":null},{"name":"[sig-storage] In-tree Volumes [Driver: rbd][Feature:Volumes] [Testpattern: Pre-provisioned PV (default fs)] subPath should support restarting containers using file as subpath [Slow]","status":"skipped","meta":null,"file":"","items":null}]}]}]}
```
zubron: It looks like `null` fields are still being included in the output. I thought they would have been skipped with the `omitempty` tag.

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,240 @@
...
+			return
+		}
+		resultFile := path.Join(results.PluginsDir, input.plugin, o.File)
+		filereader, err := r.FileReader(resultFile)
```
zubron: This error isn't being checked.

pkg/client/results/processing.go
```diff
@@ -0,0 +1,210 @@
...
+type fileSelector func(string, os.FileInfo) bool
+
+// Item is the central format for plugin results. Various plugin
+// types can be transformed into this simple format and set at a standard
+// location in our results tarball for simplified processing by any consumer.
+type Item struct {
+	Name     string            `json:"name" yaml:"name"`
+	Status   string            `json:"status" yaml:"status"`
+	Metadata map[string]string `json:"meta,omitempty" yaml:"meta,omitempty"`
+	File     string            `json:"file,omitempty" yaml:"file,omitempty"`
```
zubron: Is `File` needed if we're also including the path within `Metadata`? It's only being set when processing raw files. I'm not sure if it should only be included there or if it's being missed from the junit processor.

pkg/client/results/testdata/mockResults/plugins/ds-junit-01/results/global/output2.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
johnSchnake: Yes I can do that. I generated them all via just a bunch of copy/paste but now that the test cases are done I can make the file contents/names more clear.

pkg/client/results/testdata/mockResults/plugins/ds-junit-01/results/global/output.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+  <testsuite tests="204" failures="0" time="0.0574002">
```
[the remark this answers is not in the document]
johnSchnake: No, I could actually remove the files that dont matter (actual logs/results). The only thing that matters is the summary.

cmd/sonobuoy/app/results.go
```diff
@@ -0,0 +1,255 @@
...
+	mode       string
+	node       string
+	skipPrefix bool
+}
+
+func NewCmdResults() *cobra.Command {
+	data := resultsInput{}
+	cmd := &cobra.Command{
+		Use:   "results archive.tar.gz",
+		Short: "Inspect e2e test results. Optionally rerun failed tests",
```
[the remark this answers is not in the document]
johnSchnake: Great catch, thanks.

pkg/client/results/testdata/mockResults/plugins/ds-junit-01/ds-junit-01.golden.json
johnSchnake: I can fix those up, though it doesnt matter. In this situation, the item is same once loaded/compared.

pkg/client/results/testdata/mockResults/plugins/ds-junit-01/results/global/output.xml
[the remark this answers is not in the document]
johnSchnake: Sorry, I misunderstood the question.

No it doesn't matter but one day it actually might. For now I'll leave it as is so that it is even discernable where the data comes from.

I don't use those values (I manually iterate over the data) so that I can get passed/failed/skipped whereas this only shows tests (total) and failed.

pkg/client/results/testdata/mockResults/plugins/ds-junit-01/ds-junit-01.golden.json
zubron: Ah, sorry. I had a couple of comments which were outdated, one of which applied to the example test which _does_ show the empty fields being omitted. I thought these test files had been regenerated following those `omitempty` changes. Never mind this comment then!

cmd/sonobuoy/app/results.go
johnSchnake: meh, I dont think I spent that much time thinking about it. I'll go ahead and just bubble everything up. This only means that if one node is poorly formed then we abort instead of continuing but I dont think that is that big of a deal.

cmd/sonobuoy/app/results.go
johnSchnake: 👍 Also modified the code to always return a non-nil function (and documented it as such) so I don't have to worry about nil checks.

pkg/client/results/processing.go
johnSchnake: So the `File` field is meant to be a breadcrumb for raw files, currently, an item should set either `File` or `Items` but not both.

The metadata field was added later and the idea of putting the full filepath made sense for tests since it wasn't listed elsewhere. You're right that it is now redundant for the `raw` type.

I think you're right that we can remove it, let me work on that change. Instead of checking if `.File == ""` we will just check `if .Metadata!=nil && .Metadata["file"] != ""`. It is a bit more verbose but it makes a single source of truth for the filename.

Issue #809, opened 2019-07-23
Able to specify more pod options


johnSchnake:
**Describe the solution you'd like**
A user should be able to specify more options to be set on the pod running their plugin, eg. hostPID.

In the past we've done this by adding the fields to the plugin definition then passing the values through to the template that generates the pod/daemonset but it is frustrating/futile to keep adding each field that could be set on a pod. It requires updating the plugin defintion, the template, and a number of methods to pass the field around properly.

Instead, I propose that we allow the user to specify any/all fields via a new field on the plugin definition:

```
sonobuoy-config:
  <current config options like plugin name>
spec:
  <container spec for their plugin>
podSpec:
  <new object which would allow them to set any field on the pod. These fields would be merged with our defaults>
```

The `podSpec` field would just be a typed object from client-go and should make it relatively simple, in one go, to allow all the pod fields to be customizable.

**Anything else you would like to add:**
Questions:
 - would we want/need to add a way to set fields specifically on just daemonsets? There arent many and I think this is not as high a priority.
 - would we want to allow users to specify more containers here? We will use their main plugin container (from the spec field) and add the sonobuoy worker as a sidecar, but would we even want to allow them to set other containers here? As a first iteration, I think not.

This is a prerequisite for getting more plugins runnable via sonobuoy since some may require elevated privilege on the host.

Issue #293, opened 2018-02-27
const all the defaults.


timothysc:
Right now we have default f(n) that should be referencing consts.  

/cc @chuckha

johnSchnake:
Some of those default functions are not getting used or are just _wrong_. Will review and update to make more clean/understandable

johnSchnake:
Considering this closed; clearly we will always have the intent to make clean code that references constants rather than magic strings/values but no point in keeping this vague ticket open any longer.

PR #816, opened 2019-07-26
Update some values to use consts

zubron:
This includes a small number of changes to use
`const` where we previously using `var` or literals.

- Part of #293

There are definitely more places that could have been
updated, particularly where generating the manifests 
and plugins, however given that there may be significant
changes there with #809, I haven't gone further with that.

**Release note**:
```
NONE
```

### DIFF
--- a/pkg/plugin/manifest/serializer.go
+++ b/pkg/plugin/manifest/serializer.go
@@ -27,6 +27,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 )
 
+const (
+	kindContainer string = "container"
+	kindVolume    string = "volume"
+	kindManifest  string = "manifest"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/plugin/manifest/serializer.go
```diff
@@ -27,6 +27,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 )
 
+const (
+	container string = "container"
+	volume    string = "volume"
+	manifest  string = "manifest"
```
johnSchnake: Since these are actually being used as the `kind` can we rename them to be `kindManifest`, `kindVolume`, etc. So it is more clear in which contexts those constants are usable (as opposed to just making the english word a constant to avoid spelling mistakes or whatnot)

PR #824, opened 2019-08-01
Allow multiple result files to be post-processed

### LINKED ISSUE #821: results.Item datatype should allow specifying multiple resultFiles
When specifying a plugin definition, if you want to manually specify which files to consider you can only specify one file. We should be able to specify multiple, analogous to how the default behavior will apply to multiple files.

I think that eventually, we should be able to process different files with different logic, but the contract to do that is a little unclear. For now, we should just stick to processing all the files with the same logic.

zubron:
Instead of allowing users to only process one, or all the result files,
allow them to specify an array of file names in their plugin definition
to be post-processed.

- Fixes #821

I've added another copy of `sliceContains`. I will move this into a shared location following this PR.

**Release note**:
```
Multiple files can now be specified in the plugin definition for post-processing.
```


zubron:
Hmm, might be a bit much to have tests for each different type to
check that specifying multiple files works.

PR #829, opened 2019-08-08
Add results info to the status annotation and status cmd output

### LINKED ISSUE #819: Result information should go into PluginStatus
Once we have the capability to know if tests actually passed/failed (instead of just if a plugin ran or not) we should be updating our `status` metadata with that information. This way you can see whether or not a conformance run (or other plugin) was a success or not without even downloading the tarball.

I propose adding a new field to the `PluginStatus` object:
```
type PluginStatus struct {
	Plugin string `json:"plugin"`
	Node   string `json:"node"`
	Status string `json:"status"`

	// Propose new field below
	ResultStatus   string `json:"resultStatus"`
}
```

This value would be a string which would default to `unknown`. Once a we do post-processing on plugins, we should update the status on the aggregator pod to show the top level result (pass/failed/etc).

I am unsure if we should also walk the entire tree of data and also add the aggregate data (# passed/failed/etc).  If we want to do that we should add another field, something like:
```
ResultStatusDetails string
```
The problem/question about that is how we decide to format that so it is human-readable but also machine consumable. Maybe better would be something like:
```
ResultStatusDetails map[string]int
```
So it would look like `{"passed":5,"failed":1}`. This would be more machine consumable/scriptable but not necessarily bad for a human to read either.

Prereq: https://github.com/heptio/sonobuoy/pull/813

I consider this a relatively high priority since, even though the PR above does all this aggregation, we aren't really utilizing it as much as we can. By adding this new status information, we make it so that users don't even have to download the tarball in order to get high-level result information. That can fundamentally change some user flows to be faster/easier.

johnSchnake:

- Fixes #819 

Example output as of now:
```
"sonobuoy.hept.io/status": "{\"plugins\":[{\"plugin\":\"e2e\",\"node\":\"global\",\"status\":\"complete\",\"result-status\":\"passed\",\"result-counts\":{\"passed\":1,\"skipped\":3584}},{\"plugin\":\"systemd-logs\",\"node\":\"kind-control-plane\",\"status\":\"complete\",\"result-status\":\"passed\",\"result-counts\":{\"passed\":1}}],\"status\":\"complete\",\"tar-info\":{\"name\":\"201908081539_sonobuoy_e3a5f742-7cb2-490e-b0ee-91c6e2c8decb.tar.gz\",\"created\":\"2019-08-08T15:39:45.1724467Z\",\"sha256\":\"bbf3d5e43fe26e101d6c10b0a829ce94e6b2954366a46775172c0696793ce20b\",\"size\":601067}}"
```

TODO:
 - [x] Update the `status` command to show some of this info
 - [ ] Clean up the code and ensure someone using this as a library can easily gather this information.
 - [x] Main piece of code that needs a test is the code that checks all the leaf nodes' status.
 - [x] TBD: Should I also _always_ add a `total` field to the result counts? It would make it so people dont have to add them, but it also is a special case that wouldn't really be a count of a particular status so I'm not sure it is a great idea.

**Release note**:
```release-note
Added plugin results (based on post-processing results) to the PluginStatus object: both the overall status and the count of all the individual test results. In addition, the Status object itself (which wraps all the PluginStatus objects) has added information on the tarball name, size, and SHA256.
```

### DIFF
--- a/cmd/sonobuoy/app/status.go
+++ b/cmd/sonobuoy/app/status.go
@@ -17,10 +17,12 @@ limitations under the License.
 package app
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 	"sort"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/pkg/errors"
@@ -35,6 +37,29 @@ var statusFlags struct {
 	namespace string
 	kubecfg   Kubeconfig
 	showAll   bool
+	json      bool
+}
+
+type pluginSummaries []pluginSummary
+
+type pluginSummary struct {
+	plugin string
+	status string
+	result string
+	count  int
+}
+
+// For sort.Interface
+func (p pluginSummaries) Len() int { return len(p) }
+func (p pluginSummaries) Swap(i, j int) {
+	p[i], p[j] = p[j], p[i]
+}
+func (p pluginSummaries) Less(i, j int) bool {
+	pi, pj := p[i], p[j]
+	if pi.plugin == pj.plugin {
+		return pi.status < pj.status
+	}
+	return pi.plugin < pj.plugin
 }
 
 func NewCmdStatus() *cobra.Command {
@@ -50,7 +75,11 @@ func NewCmdStatus() *cobra.Command {
 	AddKubeconfigFlag(&statusFlags.kubecfg, flags)
 	flags.BoolVar(
 		&statusFlags.showAll, "show-all", false,
-		"Don't summarize plugin statuses, show all individually",
+		"Don't summarize plugin statuses, show results for each node",
+	)
+	flags.BoolVar(
+		&statusFlags.json, "json", false,
+		"Print the status object as json",
 	)
 
 	return cmd
@@ -73,9 +102,12 @@ func getStatus(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	if statusFlags.showAll {
+	switch {
+	case statusFlags.showAll:
 		err = printAll(os.Stdout, status)
-	} else {
+	case statusFlags.json:
+		err = printJSON(os.Stdout, status)
+	default:
 		err = printSummary(os.Stdout, status)
 	}
 
@@ -110,12 +142,17 @@ func humanReadableStatus(str string) string {
 	}
 }
 
-func printAll(w io.Writer, status *aggregation.Status) error {
-	tw := tabwriter.NewWriter(w, 1, 8, 1, '\t', tabwriter.AlignRight)
+func printJSON(w io.Writer, status *aggregation.Status) error {
+	enc := json.NewEncoder(w)
+	return enc.Encode(status)
+}
 
-	fmt.Fprintf(tw, "PLUGIN\tNODE\tSTATUS\n")
+func printAll(w io.Writer, status *aggregation.Status) error {
+	tw := defaultTabWriter(w)
+
+	fmt.Fprintf(tw, "PLUGIN\tNODE\tSTATUS\tRESULT\t\n")
 	for _, pluginStatus := range status.Plugins {
-		fmt.Fprintf(tw, "%s\t%s\t%s\n", pluginStatus.Plugin, pluginStatus.Node, pluginStatus.Status)
+		fmt.Fprintf(tw, "%s\t%s\t%s\t%s\t\n", pluginStatus.Plugin, pluginStatus.Node, pluginStatus.Status, pluginStatus.ResultStatus)
 	}
 
 	if err := tw.Flush(); err != nil {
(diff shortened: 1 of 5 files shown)


johnSchnake:
Still in a WIP phase but the gist of the logic is here. Let me know what you think.

pkg/discovery/discovery.go
```diff
@@ -255,16 +306,35 @@ func getPodLogNamespaceFilter(cfg *config.Config) string {
...
+}
+
+func updatePluginStatus(client kubernetes.Interface, namespace string, pluginType string, pluginResultStatus string, pluginResultCounts map[string]int) error {
+	podStatus, err := pluginaggregation.GetStatus(client, namespace)
+	if err != nil {
+		return errors.Wrap(err, "failed to get the existing status")
+	}
+
+	for i := range podStatus.Plugins {
+		if podStatus.Plugins[i].Plugin == pluginType {
```
zubron: Is the plugin result type unique per run? I think I remember seeing that it was but it's slightly confusing to me here because `podStatus.Plugins[i].Plugin` seems like it would be the plugin's name. I know that our built in plugins use the same values for the name and the result type though.

zubron:
I think this is looking good so far. I agree with your comment about adding a `total` field to the result counts. I think that if a sum is needed, it can be calculated. It makes more sense to me to have only the actual statuses as keys in that object.

pkg/discovery/discovery.go
johnSchnake: Yes the type is unique; agree that name/type duality is confusing and I want to fix that. See #830 ; someone reported an issue which was caused because of that too.

johnSchnake:
>  Clean up the code and ensure someone using this as a library can easily gather this information.

Going to review the code myself for cleanup, but this information is available as part of the status on the object so I think this is already pretty easy with this library (there are methods to get the status). Just not to get this info from a tarball without the pod. That can be a TBD I think.

cmd/sonobuoy/app/status.go
```diff
@@ -110,12 +142,17 @@ func humanReadableStatus(str string) string {
...
+func printJSON(w io.Writer, status *aggregation.Status) error {
+	enc := json.NewEncoder(w)
+	return enc.Encode(status)
+}
+
 func printAll(w io.Writer, status *aggregation.Status) error {
-	tw := tabwriter.NewWriter(w, 1, 8, 1, '\t', tabwriter.AlignRight)
+	tw := defaultTabWriter(w)
 
-	fmt.Fprintf(tw, "PLUGIN\tNODE\tSTATUS\n")
```
johnSchnake: The lack of a trailing `\t` caused this last column to not be aligned at all

cmd/sonobuoy/app/status.go
```diff
@@ -110,12 +142,17 @@ func humanReadableStatus(str string) string {
 	}
 }
 
+func printJSON(w io.Writer, status *aggregation.Status) error {
+	enc := json.NewEncoder(w)
+	return enc.Encode(status)
+}
+
 func printAll(w io.Writer, status *aggregation.Status) error {
-	tw := tabwriter.NewWriter(w, 1, 8, 1, '\t', tabwriter.AlignRight)
```
johnSchnake: The alignment was ignored since the padding char was `\t`, it also seemed to make it impossible to slightly tweak the output in a clear fashion. Changed this out for space padding.

cmd/sonobuoy/app/status.go
```diff
@@ -127,30 +164,38 @@ func printAll(w io.Writer, status *aggregation.Status) error {
...
+	tw := defaultTabWriter(w)
 	totals := map[string]map[string]int{}
-	for _, plugin := range status.Plugins {
-		if _, ok := totals[plugin.Plugin]; !ok {
-			totals[plugin.Plugin] = make(map[string]int)
+
+	// Effectively making a pivot chart to count the unique combinations of status/result.
+	statusResultKey := func(p aggregation.PluginStatus) string {
+		return p.Status + ":" + p.ResultStatus
```
johnSchnake: We were just iterating over values and rolling on on some value before (status). Now we roll up on the tuple (status, resultStatus).

cmd/sonobuoy/app/status.go
```diff
@@ -161,23 +206,6 @@ func printSummary(w io.Writer, status *aggregation.Status) error {
 	return nil
 }
 
-type pluginSummaries []pluginSummary
```
johnSchnake: FYI: moved this type data to the top of the page; somewhat of a nit/preference but I have always understood types to be defined at the top of the file under constants/variables.

cmd/sonobuoy/app/status_test.go
```diff
@@ -94,7 +95,7 @@ func TestPrintStatus(t *testing.T) {
 			}
 
 			if b.String() != test.expected {
-				t.Errorf("expected output to be %q, got %q", test.expected, b.String())
```
johnSchnake: Tweaked this so that each of "expected" and "got" would start on a new line; just thought it looked more clear at a glance.

PR #831, opened 2019-08-10
Sync master/root README.md and provide a script to do this

### LINKED ISSUE #772: Automate doc maintenance as much as possible
Recently a PR went in which automatically added a new 'versioned' section to the docs.

However, another thing still needs work: the master readme and links in general.

The main repo readme is duplicated in docs/master and the repo root in order to get github and the docs site to work.

The github page should link to the docs site, but the docs site should use relative links to other pages so that site previews and versioned docs link together properly.

In addition, we need to make it more simple for someone to modify the readme and ensure that it stays in sync with the docs/master readme. A simple CI check would suffice, but it will be slightly complicated since the files wont match exactly due to the link-manipulation that needs to be done between the two.

Not a huge priority, but a bug in our ear for when we have to do this again for the next release.

johnSchnake:
The README for the master version of the docs and in the root
of the repo (used by Github) should always be in sync but
require minor changes due to different relative locations of
resources and links.

Also added this script as part of our CI to ensure that the two
files don't drift.

Fixes #772

Also found that there was drift; the master docs had a link to the page about `wait` functionality, but the root readme did not. The link wasn't actually used/needed so I removed it.

**Release note**:
```
NONE
```

### DIFF
--- a/scripts/sync_readme.sh
+++ b/scripts/sync_readme.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -x
+
+SCRIPTS_DIR="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
+DIR=$(cd $SCRIPTS_DIR; cd ..; pwd)
+ROOT_README=$DIR/README.md
+MASTER_SITE_README=$DIR/site/docs/master/README.md
+
+# Users should update the README.md in the root of the repo and
+# run this script to sync the README.md in the master docs. The reason
+# the root version is taken as the source of truth is because it is more
+# simple to remove prefixes from URLs and paths than to figure out
+# where to add them.
+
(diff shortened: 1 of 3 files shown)


scripts/sync_readme.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -x
+
+SCRIPTS_DIR="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
+DIR=$(cd $SCRIPTS_DIR; cd ..; pwd)
+ROOT_README=$DIR/README.md
+MASTER_SITE_README=./site/docs/master/README.md
```
zubron: This should be `$DIR/site/...` so that it's independent of where it's being run from.

PR #832, opened 2019-08-11
Add SEO functionality to the site

jonasrosland:
These changes should enable the SEO plugin functionality on the site, leading to better search results.

Signed-off-by: Jonas Rosland <jrosland@vmware.com>

### DIFF
--- a/site/Gemfile
+++ b/site/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 gem 'github-pages'
 gem 'redcarpet'
+gem 'jekyll-seo-tag'
--- a/site/_config.yml
+++ b/site/_config.yml
@@ -2,9 +2,12 @@
 title: Sonobuoy
 email:
 author: Sonobuoy Authors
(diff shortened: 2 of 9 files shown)


johnSchnake:
Seems 99% fine to me; the one problem I found is that included in the metadata that the SEO stuff adds is a `<title>` tag. On most pages (including the main landing page) that is fine since we already set a title and (apparently) the browser just respects the first one (but maybe that is just Chrome; not certain of other browsers).

For the docs pages though it adds:
```

<!-- Begin Jekyll SEO tag v2.5.0 -->
--
  | <title>[![Build Status][status]][travis] \| Sonobuoy</title>

```

and we didnt already have a title set; the plugin usually makes it from a header I think. We need a better title on the page than that.

jonasrosland:
Good catch! I've updated the docs template to not include the `title` portion, that should work now.

johnSchnake:
Awesome, glad the fix was so straightforward. Approving and merging.

PR #834, opened 2019-08-12
Remove use of templates for plugins

zubron:
**What this PR does / why we need it:**
This change removes the use of templates for creating resources when
running plugins and instead creates the objects directly to use with the
Kubernetes API. This change should introduce no change in behaviour. It
is just refactoring in preparation for adding support to customise pod
options.

I have added tests for the new changes in `base.go`. I have only adapted the existing tests for checking that the manifests are created correctly. I can expand on these tests if needed. I have also tested this manually by comparing the resources created for both Job and DaemonSet plugins before and after this change and they are equivalent.

**Release note**:
```
NONE
```


cmd/sonobuoy/app/gen_plugin.go
```diff
@@ -1,127 +0,0 @@
-/*
```
johnSchnake: Should be able to rebase onto `upstream/master` and this commit shouldn't come up anymore.

pkg/plugin/driver/base.go
```diff
@@ -42,23 +45,6 @@ type Base struct {
 	CustomAnnotations map[string]string
 }
 
-// TemplateData is all the fields available to plugin driver templates.
-type TemplateData struct {
```
johnSchnake: 🎉 for datatypes we get to remove.

pkg/plugin/constants.go
```diff
@@ -19,4 +19,7 @@ package plugin
 const (
 	// GracefulShutdownPeriod is how long plugins have to cleanly finish before they are terminated.
 	GracefulShutdownPeriod = 60
+
+	// ResultsDir is the diectory where results will be available in Sonobuoy plugin containers.
```
johnSchnake: s/diectory/directory

pkg/plugin/driver/base_test.go
```diff
@@ -101,3 +104,132 @@ func TestSkipCleanup(t *testing.T) {
...
+				Value: "",
+			},
+		}
+		for _, e := range expectedEnvVars {
+			if !envContains(wc.Env, e) {
+				t.Errorf("expected container environment to contain %q", e)
+			}
+		}
+		for _, e := range wc.Env {
+			if e.Name == "NODE_NAME" {
```
johnSchnake: nit: would prefer a `switch` instead of an `if...else...if else....if else...`

zubron: Sure, and thanks for the suggestion. I'm not particularly happy with this test :-/

johnSchnake: It is a little odd the way you have subtests but it isn't a table driven test.

The other options it seems:
 - It is testing a single large object so you could dump it into one big test instead of bothering with the sub-tests. The error messages would just provide the context
 - make a table driven test/subtests and split off the larger chunks of logic (which make it hard to read) into little helper funcs. So you'd have:
```
tc := testCases{
  desc string
  expectFunc func(<types we're checking>) error
}{
  { desc: "Env vars set",
    expectFunc: checkEnvVars},
...
}
```

The benefit to (1) is that its straight forward to code but the downside it isn't clear when reading any particular section of the test what has been tested, will be tested, could be added, etc.

The benefit of (2) is that you can look at the table and atleast see a list of all the things you intend to test about it (even if only vague details). The downside is you have to reorg the code a bit to put those test functions in an accessible place and it may seem a bit odd since you won't be reusing those functions elsewhere. But you could, within that test itself, define those functions inline and then only make the table at the end:

```
func TestXYZ(t *testing.T){
  checkEnvVars:= func(...) error{...}
  checkFields:= func(...) error{...}

  tc := testCases {...}{...}
...
```

I think I would probably prefer (2) if I were writing it, but I wont block it either way.

zubron: I decided to change this test and do it the way you mentioned in (2). Let me know what you think.

pkg/plugin/driver/base_test.go
```diff
@@ -101,3 +105,161 @@ func TestSkipCleanup(t *testing.T) {
...
+
+	}
+
+	testCases := []struct {
+		desc  string
+		check func(v1.Container) error
+	}{
+		{
+			"Fields set correctly",
+			checkFields,
```
johnSchnake: 👍 I do like reading the test more this way. It is also clear how I could/should expand the functionality if I needed to test something else.

PR #835, opened 2019-08-13
Add image building/pushing back to Travis

### LINKED ISSUE #808: Requirement to move builds off of older Jenkins servers
Some details TBD but we can either:
 - migrate those jobs to new servers
 - consider if we want to move those jenkins jobs into something like a dockerhub build process which we now have available via VMware

If we go with (2) above then we need to also decide about moving our older images from grc.io to dockerhub; not sure what the thoughts are on that.

johnSchnake:
Got authorization to re-add credentials for image pushing to
Travis via their recommended approach outlined here:
https://docs.travis-ci.com/user/encrypting-files/

Fixes #808

**Release note**:
```
NONE
```


.travis.yml
```diff
@@ -1,40 +1,31 @@
...
-
+- docker
 install: true
-
-script: 
-  - VERBOSE=true make test int 
-  - ./travis-ci.sh 
-
+script:
+- VERBOSE=true make test int
```
johnSchnake: Most of these changes are due to the fact that the `travis` cli tool edits/rewrites the `.travis.yml` file. So it is probably just a serialization improvement even if both are technically valid.

johnSchnake:
So this is a 👍 to moving to CI at some point I think:
https://graysonkoonce.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/
https://github.com/travis-ci/travis-ci/issues/6652

We'll need to adjust this script to deal with fact that the `TRAVIS_BRANCH` isn't what we think here.

travis-deploy.sh
```diff
@@ -12,8 +12,15 @@ function goreleaser() {
     curl -sL https://git.io/goreleaser | bash
 }
 
-if [ ! -z "$TRAVIS_TAG" ]; then
+function gcr_push() {
+    openssl aes-256-cbc -K $encrypted_222a2009ef6d_key -iv $encrypted_222a2009ef6d_iv -in heptio-images-c14f11347d8b.json.enc -out heptio-images-c14f11347d8b.json -d
+    gcloud auth activate-service-account --key-file heptio-images-c14f11347d8b.json
+    # https://github.com/travis-ci/travis-ci/issues/9905
```
zubron: This issue was closed. Do we still need to do the `unset` below?

.travis.yml
```diff
@@ -1,40 +1,33 @@
...
+- docker
 install: true
-
-script: 
-  - VERBOSE=true make test int 
-  - ./travis-ci.sh 
-
+script:
+- export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
+- echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, PR=$TRAVIS_PULL_REQUEST, BRANCH=$BRANCH"
```
johnSchnake: Added these 2 for clarity; very unintuitive how `$TRAVIS_BRANCH` won't be the PRs branch. See the issues/article I linked to elsewhere on this PR.

travis-deploy.sh
johnSchnake: /shrug

I was just adding back the exact code that we previously had done this work. Let's just leave it (if it works) since this is all probably temporary anyways.

zubron: > this is all probably temporary anyways
That is a very good point. Ignore that comment then :)

PR #836, opened 2019-08-14
Use gcloud auth instead of docker-credential-gcr

johnSchnake:
We used docker-credential-gcr only once we moved to Jenkins and
it was available on the agents already. This change moves back
to using gcloud auth like it was when we originally published
images via Travis.

Related to #808 

Unable to test this on non-master branches since creds are protected. Will have to merge and iterate.

**Release note**:
```
NONE
```


johnSchnake:
xref https://github.com/heptio/sonobuoy/commit/c27ab1a150330047683a0f314cb08e9283465ae8

PR #837, opened 2019-08-14
Add ability to override pod spec options

### LINKED ISSUE #809: Able to specify more pod options

zubron:
This change introduces the ability to override PodSpec options for
plugins. A new flag `--show-default-podspec` has been added to the
`gen` and `gen plugin` commands. When used, Sonobuoy will include the
default PodSpec options for the given plugin driver in the resulting
yaml. This allows users to change these values or they can provide their
own. If one is provided in the plugin definition, Sonobuoy will use that
as is, otherwise it will use the default.

- Fixes #809 

**Release note**:
```
Sonobuoy now includes the option to edit the PodSpec used when running plugins. Users can provide their own PodSpec or adapt the default one used by Sonobuoy. The default can be obtained using the '--show-default-podspec' flag with the 'gen' and 'gen plugin' commands.
```

### DIFF
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -78,6 +78,10 @@ type GenConfig struct {
 	// out of band from the plugins because of how the dynamic plugins are not
 	// yet able to be manipulated in this way.
 	PluginEnvOverrides map[string]map[string]string
+
+	// ShowDefaultPodSpec determines whether or not the default pod spec for
+	// the plugin should be incuded in the output.
+	ShowDefaultPodSpec bool
 }
 
 // Validate checks the config to determine if it is valid.
--- a/pkg/plugin/driver/base.go
+++ b/pkg/plugin/driver/base.go
@@ -22,6 +22,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"strings"
 
 	"github.com/heptio/sonobuoy/pkg/plugin"
 	"github.com/heptio/sonobuoy/pkg/plugin/manifest"
@@ -60,6 +61,11 @@ func (b *Base) GetSecretName() string {
 // GetResultType returns the ResultType for this plugin (to adhere to plugin.Interface).
 func (b *Base) GetResultType() string {
 	return b.Definition.SonobuoyConfig.ResultType
+}
+
+// GetDriver returns the Driver for this plugin.
+func (b *Base) GetDriver() string {
+	return b.Definition.SonobuoyConfig.Driver
 }
 
 // SkipCleanup returns whether cleanup for this plugin should be skipped or not.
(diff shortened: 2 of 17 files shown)


pkg/client/interfaces.go
```diff
@@ -78,6 +78,9 @@ type GenConfig struct {
 	// out of band from the plugins because of how the dynamic plugins are not
 	// yet able to be manipulated in this way.
 	PluginEnvOverrides map[string]map[string]string
+
+	// ShowDefaultPodSpec
```
johnSchnake: todo: add a godoc style comment that adds info, not just avoids the warning :-p

pkg/plugin/driver/base.go
```diff
@@ -211,3 +216,65 @@ func (b *Base) CreateWorkerContainerDefintion(hostname string, cert *tls.Certifi
...
+				Operator: v1.TolerationOpExists,
+			},
+		},
+		Volumes: []v1.Volume{},
+	}
+}
+
+// DefaultPodSpec returns the default pod spec used for the given plugin driver type.
+func DefaultPodSpec(d string) (v1.PodSpec, error) {
+	switch d {
```
johnSchnake: I believe I tried to avoid caring too much about the casing; on the server when dealing with plugins I just `strings.ToLower` the driver to ensure I don't fail someone for `daemonset` vs `Daemonset`

pkg/plugin/driver/base.go
```diff
@@ -211,3 +216,65 @@ func (b *Base) CreateWorkerContainerDefintion(hostname string, cert *tls.Certifi
...
+	case "DaemonSet":
+		return defaultDaemonSetPodSpec(), nil
+	case "Job":
+		return defaultJobPodSpec(), nil
+	default:
+		return v1.PodSpec{}, errors.Errorf("unknown driver %q", d)
```
johnSchnake: You could also avoid dealing with the error case here at all and just do something like default to the `defaultJobPodSpec` when it is unknown. I think that is a pretty reasonable thing to do instead of blowing up at this stage. Just makes it a sane default.

That would just simplify the signature and call sites of this. Just an idea; not a requirement.

zubron: :+1: That's a sensible default. Adding the error caused a chain of changes elsewhere that I'd rather avoid.

johnSchnake:
Where are you at on this? Still marked as a draft but I don't see a lot that really needs work. Maybe a double check over key pieces of logic and ensure they are tested (but I don't want to stall it just to add tests for trivial logic)

zubron:
@johnSchnake I was just adding a few more tests to some of the parts that I knew weren't covered such as the job/daemonset driver packages. I should be done with it very shortly though and will ping you when ready.

PR #838, opened 2019-08-14
Add Details map to post-processed Items

### LINKED ISSUE #827: New results data doesnt include stdout
A common use case is not only wanting to know about the failures but to also drill down into the logs they output.

We should, in some capacity, be capturing that data as well.

It is so long that I don't think it is necessarily the best thing to always print it, but I have to ideas for now:
 1. Dont copy the actual logs, but instead add another metadata entry about the file/byteoffset so that we could read it if we wanted.
 2. Add another output mode, something like, `detailedLog` (? just an idea) that would also print that information. It is _very_ long so printing it by default, even in the `detailed` mode is a bit overkill.

johnSchnake:
Currently just used for junit and specifically meant to address
a gap in necessary capabilities for processing e2e results.

When tests fail, you want to see what tests fail but also why they
failed. Without this functionality, you still have to resort to
untarring the entire tarball just to see basic logs like "timeout"

Fixes #827

To test this out with a failure I used:
```
sonobuoy run --sonobuoy-image=schnake/sonobuoy:stdoutResultsInfo --plugin e2e --e2e-focus="\[Driver: local\]\[LocalVolumeType: dir-link-bindmounted\] \[Testpattern: Pre-provisioned PV \(default fs\)\] subPath should support existing directories when readOnly specified in the volumeSource" --wait --image-pull-policy=Always --wait-output=spinner && outfile=$(sonobuoy retrieve) && rm -rf results && mkdir results && tar -xf $outfile -C results && code results && sonobuoy delete --wait
```

Then
```
sonobuoy results $outfile --mode detailed|jq 'select(has("details"))'
```

An interesting note is that this works:
```
sonobuoy results $outfile --mode detailed|jq 'select(has("details"))' | jq .details.failure
```

But not:
```
sonobuoy results $outfile --mode detailed|jq 'select(has("details"))' | jq .details.system-out
```

I'm not sure if it is my shell or jq or just a really bad idea to have the hyphenated key. I can modify that if needed to just strip non [a-zA-Z] values but I'd prefer to just keep the junit key names.

Update: See https://github.com/stedolan/jq/issues/38
You can do it but need to use `.["system-out"]` but then also need to worry about your shell escaping so mine looked awkward: `jq 'select(has("details"))|.details'|jq .\[\"failure\"\]`

**Release note**:
```
Adds a new `Details` field to the `results.Item` type which allows the post-processing to capture test output and report on it.
```


johnSchnake:
I actually dont think I'll modify anything for the output. In report mode no changes are needed. In detailed mode they asked for details so this seems reasonable. For dump mode they'll want it anyways.

I dont think I need to add a "a little bit of details" mode where we list all the tests but _not_ the output. If demand for that comes up, we can implement it later with some more thought about the naming.

PR #844, opened 2019-08-15
Modify some old docs to avoid liquid syntax warnings

### LINKED ISSUE #775: Netlify gives lots of warnings for syntax
Just take a look at any rendering of the preview sites of netlify and you'll see lots of messages like:
```
Liquid Warning: Liquid syntax error (line 199): [:dot, "."] is not a valid expression in "{{.SonobuoyConsumer}}" in docs/v0.14.2/enhancements/streamlined-templates.md
```

I'm guessing there is a conflict between the templating logic and the code blocks in those files.

No warnings.

Probably nothing bad about this and we don't put the enhancements on the website, but it is just a problem IMO because (1) you have to wonder if it does/could cause problems (2) it clutters the logs so you can't tell if anything actually had warnings/errors you needed to see.

**Environment:**

- Sonobuoy version: 

johnSchnake:
Markdown docs that tried to show go template style code like
`{{ .Foo }}` would cause liquid syntax warnings in netlify.

We always just assumed they were warnings but it actually did fail
to parse those sections so those pages are rendered incorrectly.
That is not too serious since in general we dont link to those
but it causes it to be hard to see _real_ problems in the netlify
logs and also, if you go to the page manually, shows the wrong
data.

This resolves that issue by wrapping the sections in question with
tags to make it treat those sections as 'raw' data.

Fixes #775

https://github.com/Shopify/liquid/issues/927

**Release note**:
```
NONE
```

### DIFF
--- a/site/docs/v0.13.0/enhancements/streamlined-templates.md
+++ b/site/docs/v0.13.0/enhancements/streamlined-templates.md
@@ -9,6 +9,7 @@ By way of illustration, here is the current `heptio-e2e` plugin file. Every part
 of user configuration is in green, all boilerplate required by Sonobuoy is in
 red.
 
+{% raw %}
 ```diff
 - ---
 - apiVersion: v1
@@ -67,6 +68,7 @@ red.
 -   - emptyDir: {}
 -     name: results
 ```
+{% endraw %}
 
 ## Summary
 
@@ -114,6 +116,7 @@ spec:
 This will be merged into a larger Pod document provided by the Sonobuoy YAML
 driver
 
+{% raw %}
 ```yaml
 ---
 apiVersion: v1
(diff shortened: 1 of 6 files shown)


johnSchnake:
One different class of issue to resolve now; https://app.netlify.com/sites/sonobuoy-live/deploys/5d55bfc4ede157000870857c

```
Liquid Warning: Liquid syntax error (line 24): Expected end_of_string but found pipe in "page.hero.promo1.url | size > 0" in index.html
```

which I assume is similarly easy to resolve

johnSchnake:
All warnings resolved on netlify now: https://app.netlify.com/sites/sonobuoy-live/deploys/5d55c0e7a9a1cd00082125f1

PR #845, opened 2019-08-16
Set OwnerReference on Plugin resources

### LINKED ISSUE #818: The master should be setting the metadata.OwnerReference for all the pods that it creates

The sonobuoy master / aggregator will be creating pods / daemonsets for the plugins to runs. However if the master pods gets deleted , the plugin pods could potentially remain orphaned. 
To solve this scenario, the plugin pods should have their metadata.ownerReference set to the master pod. In such a scenario, as soon as the master gets deleted, all resource that have the ownerReference set to the master will also be cleaned up.

This is particularly useful in cases where the scan needs to be terminated preemptively . 

**Environment:**

- Sonobuoy version: sha256:0f67f88d2626e35f0d0351803eb1812975ec57f7ba45cdb842b42c39c3f74adf
- Kubernetes version: (use `kubectl version`): 14

zubron:
When creating Pod or DaemonSet resources when running plugins, we didn't
mark them as owned by the Sonobuoy aggregator pod. This meant that if
the aggregator pod was deleted, the resources created by that pod would
be orphaned and would have to be deleted manually. This change sets the
OwnerReference for all resources to be the Sonobuoy aggregator pod so
that when the aggregator is deleted, all child resources are also
deleted.

- Fixes #818 

**Release note**:
```
Sonobuoy now sets the `OwnerReference` on all Pod and DaemonSet resources that are created when running plugins. This means that if the main aggregator pod is deleted, all resources created by that pod will also be deleted.
```

### DIFF
--- a/pkg/plugin/aggregation/update.go
+++ b/pkg/plugin/aggregation/update.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -35,6 +36,13 @@ const (
 	StatusPodLabel       = "run=sonobuoy-master"
 	DefaultStatusPodName = "sonobuoy"
 )
+
+// NoPodWithLabelError represents an error encountered when a pod with a given label can't be found
+type NoPodWithLabelError string
+
+func (n NoPodWithLabelError) Error() string {
+	return string(n)
+}
 
 // node and name uniquely identify a single plugin result
 type key struct {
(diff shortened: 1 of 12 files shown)


zubron:
Marking this as a draft until I add some tests for `GetAggregatorPod`.

pkg/plugin/aggregation/update.go
```diff
@@ -166,25 +174,41 @@ func GetPatch(annotation string) map[string]interface{} {
...
-		return podList.Items[0].GetName(), nil
+		return &podList.Items[0], nil
+	}
+}
+
+// GetAggregatorPodName gets the sonobuoy aggregator pod name based on its label.
+func GetAggregatorPodName(client kubernetes.Interface, namespace string) (string, error) {
+	ap, err := GetAggregatorPod(client, namespace)
+
+	if _, ok := err.(NoPodWithLabelError); ok {
```
johnSchnake: nit: when doing multi-step error checking like this I tend to prefer a switch statement so it is a single block and the `if err != nil` section doesn't drift away from the call generating the error.

```
if err != nil {
        switch err.(type) {
        case * NoPodWithLabelError:
            fmt.Println(err.Error())
        default:
            fmt.Println("ah!?")
        }
    }
```

It is one extra layer of indentation, but effectively one unit of error handling. (again, just a nit)

PR #846, opened 2019-08-16
Bump version of Sonobuoy and add new docs for v0.15.2

johnSchnake:


johnSchnake:
The doc script leaves links as `sonobuoy.io/docs/...` on the readme in the new version it creates. Will fixup in a followup PR; manually fixed them here.

johnSchnake:
Also want to update the release doc to mention publishing the docs for the version at the same time as the version bump. It makes sense to put those in the same step together. We can always edit them later but no reason not to add them in the same commit.

PR #847, opened 2019-08-19
Rename master to aggregator

### LINKED ISSUE #238: rename master to controller
In the spirit of https://github.com/kubernetes/website/issues/6525 I'd like to rename master to controller.

This will take a bit of time, but all new things should refer to master as controller.

johnSchnake:
Slightly more intuitive and useful name. Kept the old name
as an alias which we can decide to remove later if we want.

Did not update a few of the config values or script names
that are named with 'master' in order to avoid potentially
breaking changes.

Fixes #238

Adding comments inline but also adding a few things here for clarity:
 - Did not rename `run_master.sh`
 - Did not update `run_master.sh` to use the new `aggregator`
 - Renamed the `master` command to `aggregator` but kept `master` as an alias so there should be no problem with old clients/scripts calling newer Sonobuoy images.

**Release note**:
```
NONE
```


cmd/sonobuoy/app/master.go
```diff
@@ -27,47 +28,57 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var noExit bool
-var kubecfg Kubeconfig
+type aggregatorInput struct {
+	noExit  bool
+	kubecfg Kubeconfig
+}
```
johnSchnake: Just a single type/variable instead of 2 globals. I then create a single instance of this variable below in `NewCmdAggregator` which returns a closure around that instance.

cmd/sonobuoy/app/master.go
```diff
@@ -27,47 +28,57 @@ import (
...
+	input := aggregatorInput{}
 	cmd := &cobra.Command{
-		Use:    "master",
-		Short:  "Runs the master/aggregator component (for internal use)",
-		Long:   "Sonobuoy is an introspective kubernetes component that generates reports on cluster conformance, configuration, and more",
-		Run:    runMaster,
+		Use:   "aggregator",
+		Short: "Runs the aggregator component (for internal use)",
+		Long:  "Sonobuoy is an introspective kubernetes component that generates reports on cluster conformance, configuration, and more",
+		Run: func(cmd *cobra.Command, args []string) {
```
johnSchnake: Didn't want to change this to `RunE` but will do that with a bunch of commands at the same time eventually. We should be writing our code in a way that allows us to bubble up errors to the main `Execute` method so we can centralize error handling rather than having logging/`os.Exit` all over the place.

cmd/sonobuoy/app/worker.go
```diff
@@ -65,10 +66,6 @@ var singleNodeCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(0),
 }
 
-func runGatherHelp(cmd *cobra.Command, args []string) {
-	cmd.Help()
-}
```
johnSchnake: No need to add this, if you don't set a `Run` or `RunE` it will show usage messages already.

main.go
```diff
@@ -20,13 +20,17 @@ import (
...
 
 	"github.com/heptio/sonobuoy/cmd/sonobuoy/app"
+	"github.com/heptio/sonobuoy/pkg/errlog"
 )
 
-// main entry point of the program
+// Main entry point of the program. Execute methods historically would log
+// errors and exit manually via os.Exit in which case the error handling
+// here never is invoked. We want to move towards commands that return errors
+// and use this generic log/exit logic.
```
johnSchnake: Let me roll this back; I didnt mean to introduce this yet.

pkg/client/testdata/default-plugins-via-nil-selection.golden
```diff
@@ -95,7 +95,7 @@ spec:
   - command:
     - /bin/bash
     - -c
-    - /sonobuoy master --no-exit=true -v 3 --logtostderr
+    - /sonobuoy aggregator --no-exit=true -v 3 --logtostderr
```
johnSchnake: This shouldn't be changing...

johnSchnake: Shouldnt change...

pkg/plugin/driver/daemonset/daemonset.go
```diff
@@ -107,7 +107,7 @@ func (p *Plugin) createDaemonSetDefinition(hostname string, cert *tls.Certificat
 		Labels:      labels,
 		Annotations: annotations,
 		OwnerReferences: []metav1.OwnerReference{
-			metav1.OwnerReference{
```
johnSchnake: Just a gofmt change; I have my editor set to use `gofmt -s` which simplifies more than the default/minimum.

pkg/plugin/interface.go
```diff
@@ -116,7 +116,7 @@ type AggregationConfig struct {
 
 // WorkerConfig is the file given to the sonobuoy worker to configure it to phone home.
 type WorkerConfig struct {
-	// MasterURL is the URL we talk to for submitting results
+	// MasterURL is the URL we talk to the aggregator pod on for submitting results
```
johnSchnake: Did not change this since it is part of an exported struct/config.

site/docs/v0.13.0/enhancements/v2-worker-master.md
```diff
@@ -17,7 +17,7 @@
 
 ## How it works now 
 
-Sonobuoy uses a worker-master model, where a master delegates tasks to worker
```
johnSchnake: This one line from some old docs had gotten updated; I don't really care about the older ones. If its no different to anyone else I'll just leave it rather than trying to fix up other sections in this (or other older docs)

cmd/sonobuoy/app/master.go
```diff
@@ -27,47 +28,57 @@ import (
...
 	errcount := discovery.Run(kcfg, cfg)
 
-	if noExit {
+	if input.noExit {
 		logrus.Info("no-exit was specified, sonobuoy is now blocking")
 		select {}
 	}
 
-	os.Exit(errcount)
+	return fmt.Errorf("%v errors encountered during execution", errcount)
```
zubron: By using `fmt.Errorf` here, we're always returning a non-nil error. We should be returning nil in the case where errcount is 0.

site/docs/v0.13.0/enhancements/v2-worker-master.md
zubron: Are these files even linked to from the site? I can't find any references to them.

johnSchnake: The enhancements aren't linked but they do actually exist if you manually write the URL.

They were really for internal discussion/documentation back then; I made the decision that I didn't really want/need them to be displayed in our online docs site though.

cmd/sonobuoy/app/master.go
johnSchnake: 👍 Fixed.

PR #851, opened 2019-08-21
Add junit support for non-e2e tests/multi-suite

### LINKED ISSUE #849: Valid junit fails to be parsed correctly
Plugin wrote results file with contents like:
```
<?xml version='1.0'?>
<testsuites>
  <testsuite name='...' tests='124' failed='37' failures='37'>
    <testcase name='...'>
```
But this failed to be parsed as valid junit:
```
- name: output.xml
  status: unknown
  meta:
    error: expected element type <testsuite> but have <testsuites>
    file: results/global/tmp/results/output.xml
```

Should have handled this just fine; my guess is that our junit has just been overfitted to the e2e results which is a single testsuite. Our junit parsing should probably add an extra level of depth so that each testsuite is its own branch on the tree.

**Environment:**

- Sonobuoy version: v0.15.2

johnSchnake:
The e2e tests report via junit but with the top-level item
in the xml being a `<testsuite>`. Other implementations seem
to default to allowing multiple testsuites and so they have a
higher level `testsuites` object.

This PR enables a new post-processing path and makes the older
junit parsing only apply to the result-type `e2e`. The new,
`junit` result-type will use this more generalized logic.

Fixes #849

TBD; need to discuss this because the above commit message isn't accurate, nor is the issue I originally filed. I thought we were setting the result format for `e2e` but we were not; we were using `junit`. So this change makes older `sonobuoy gen` output no longer get post-processed correctly.

That is a brand new feature so maybe not a huge deal, but we should consider not changing that.
Instead of `junit` being the 'new' format, maybe we continue to let `junit` be the e2e-junit/single-suite and we introduce a new `junit-suites` or something similar for the new format.

**Release note**:
```
JUnit output which containts a `testsuites` object consisting of multiple test suites can now be post-processed by setting the result-format to `TBD`
```


johnSchnake:
After discussion on this we decided that the preferred route would be to try and make the `junit` result-format be compatible with both.

There will be a bit more work to accomplish this but we want to try and avoid having N different flavors of junit. Instead, it would be preferable to have logic that will figure them all out and report on them as needed.

pkg/client/e2e.go
```diff
@@ -27,16 +27,15 @@ import (
 
 	"github.com/heptio/sonobuoy/pkg/client/results"
 	"github.com/heptio/sonobuoy/pkg/client/results/e2e"
-	"github.com/onsi/ginkgo/reporters"
```
johnSchnake: Now use our own type which is an amalgamation of the ginkgo/reporters and the jstemmer types. Paired with a custom unmarshaller, this allows us to handle both cases with a single type.

pkg/client/e2e.go
```diff
@@ -46,7 +45,7 @@ func (*SonobuoyClient) GetTests(reader io.Reader, show string) ([]reporters.JUni
 			}
 			// TODO(chuckha) consider reusing this function for any generic e2e-esque plugin results.
 			// TODO(chuckha) consider using path.Join()
-			if path == e2eJunitPath || path == legacye2eJunitPath {
+			if path == e2eJUnitPath || path == legacye2eJUnitPath {
```
johnSchnake: Updated a bunch of little lines for consistent naming: `JUnit` preferred over `Junit`

pkg/client/e2e.go
```diff
@@ -57,35 +56,35 @@ func (*SonobuoyClient) GetTests(reader io.Reader, show string) ([]reporters.JUni
...
 	if !found {
-		return nil, fmt.Errorf("failed to find results file %q in archive", e2eJunitPath)
+		return nil, fmt.Errorf("failed to find results file %q in archive", e2eJUnitPath)
 	}
 
-	out := make([]reporters.JUnitTestCase, 0)
+	out := make([]results.JUnitTestCase, 0)
 	if show == "passed" || show == "all" {
-		out = append(out, results.Filter(results.Passed, junitResults)...)
+		out = append(out, results.JUnitFilter(results.JUnitPassed, junitResults)...)
```
johnSchnake: Although not strictly requiring this prefix now that there isnt a `junit` and a `e2e` filter/passed/etc, it makes sense to me to leave a prefix since this _is_ type specific.

As we add other formats we will probably need similar named things. At that point we may also want to consider some interface to do this instead though.

pkg/client/gen.go
```diff
@@ -294,7 +294,7 @@ func e2eManifest(cfg *GenConfig) *manifest.Manifest {
 			PluginName:   "e2e",
 			Driver:       "Job",
 			ResultType:   "e2e",
-			ResultFormat: "junit",
+			ResultFormat: "e2e",
```
johnSchnake: This should go back to junit, but it really doesnt matter: they get processed on the same path.

pkg/client/results/junit.go
```diff
@@ -18,25 +18,104 @@ package results
...
-// Filter keeps only the tests that match the predicate function.
-func Filter(predicate func(testCase reporters.JUnitTestCase) bool, testSuite reporters.JUnitTestSuite) []reporters.JUnitTestCase {
-	out := make([]reporters.JUnitTestCase, 0)
+// junitResult is a wrapper around the suite[s] which enable results to
+// be either a single suite or a collection of suites. For instance,
+// e2e tests (which use the onsi/ginkgo reporter) report a single, top-level
+// testsuite whereas other tools report a top-level testsuites object
+// which may have 1+ testsuite children. Only one of the fields should be
+// set, not both.
+type junitResult struct {
```
johnSchnake: This type is "where the magic happens" since it has a custom unmarshal method which will handle either case based on the top level object.

pkg/client/results/junit.go
```diff
@@ -18,25 +18,104 @@ package results
...
+	suites JUnitTestSuites
+}
+
+// UnmarshalXML will unmarshal the document into either the 'Suites'
```
johnSchnake: Needs updated; in another implementation I had `Suites` and `Suite` and only one would get set. Instead, now there is just suites and we will just add the lone suite to the list if that is what we are given.

pkg/client/results/junit.go
```diff
@@ -18,25 +18,104 @@ package results
...
-func Filter(predicate func(testCase reporters.JUnitTestCase) bool, testSuite reporters.JUnitTestSuite) []reporters.JUnitTestCase {
-	out := make([]reporters.JUnitTestCase, 0)
+// junitResult is a wrapper around the suite[s] which enable results to
+// be either a single suite or a collection of suites. For instance,
+// e2e tests (which use the onsi/ginkgo reporter) report a single, top-level
+// testsuite whereas other tools report a top-level testsuites object
+// which may have 1+ testsuite children. Only one of the fields should be
+// set, not both.
+type junitResult struct {
+	suites JUnitTestSuites
```
johnSchnake: This is slightly awkward naming but I dont have a better way to name this really. It just leads to `junitResults.suites.Suites...`. I hate the repetition but I couldn't come up with something that would be clear/intutive/didnt repeat.

pkg/client/results/junit.go
```diff
@@ -18,25 +18,104 @@ package results
...
+	if start.Name.Local == "testsuites" {
+		returnErr = d.DecodeElement(&j.suites, &start)
+	} else {
+		var s JUnitTestSuite
+		returnErr = d.DecodeElement(&s, &start)
+		j.suites = JUnitTestSuites{Suites: []JUnitTestSuite{s}}
+	}
+	for i := range j.suites.Suites {
+		if j.suites.Suites[i].Name == "" {
+			j.suites.Suites[i].Name = fmt.Sprintf("testsuite-%03d", i+1)
```
johnSchnake: When processing junit results, each suite is on its own branch. As a result, if you dont have a suite name you get a weird branch that you can't target very well.

In particular, this happens for the e2e tests so I wanted to ensure that non-empty names were set.

pkg/client/results/junit.go
```diff
@@ -18,25 +18,104 @@ package results
...
+// JUnitTestCase is a single test case with its result.
+type JUnitTestCase struct {
+	XMLName     xml.Name             `xml:"testcase"`
+	Classname   string               `xml:"classname,attr"`
+	Name        string               `xml:"name,attr"`
+	Time        string               `xml:"time,attr"`
+	SkipMessage *JUnitSkipMessage    `xml:"skipped,omitempty"`
+	Failure     *JUnitFailureMessage `xml:"failure,omitempty"`
+	SystemOut   string               `xml:"system-out,omitempty"`
+	SystemErr   string               `xml:"system-err,omitempty"`
```
johnSchnake: New field; not really used right now but could be. Didnt add this into the metadata but could.

pkg/client/results/junit.go
```diff
@@ -45,84 +124,119 @@ func Filter(predicate func(testCase reporters.JUnitTestCase) bool, testSuite rep
...
 		return resultObj, errors.Wrapf(err, "opening file %v", currentFile)
 	}
 	defer infile.Close()
 
-	decoder := xml.NewDecoder(infile)
-	if err := decoder.Decode(&junitResults); err != nil {
-		resultObj.Status = StatusUnknown
-		resultObj.Metadata["error"] = err.Error()
-		return resultObj, errors.Wrap(err, "error decoding json into object")
+	resultObj, err = junitProcessReader(
```
johnSchnake: Removed all the file handling and put the rest of the logic into this new method.

pkg/client/results/processing.go
```diff
@@ -67,25 +68,6 @@ type postProcessor func(string, string) (Item, error)
 // file only or all files with a given suffix (for instance).
 type fileSelector func(string, os.FileInfo) bool
 
-// Item is the central format for plugin results. Various plugin
-// types can be transformed into this simple format and set at a standard
-// location in our results tarball for simplified processing by any consumer.
-type Item struct {
```
johnSchnake: Had moved this into `types.go` when I was playing around with refactoring into subpackages; though that went nowhere. I can add it back so it removes a bunch of unnecessary diff.

pkg/client/results/testdata/mockResults/plugins/job-junit-02/job-junit-02.golden.json
```diff
@@ -1 +1 @@
...
...nted] [Testpattern: Dynamic PV (default fs)] subPath should support existing directories when readOnly specified in the volumeSource","status":"skipped"},{"name":"[sig-storage] In-tree Volumes [Driver: rbd][Feature:Volumes] [Testpattern: Pre-provisioned PV (default fs)] subPath should support restarting containers using file as subpath [Slow]","status":"skipped"}]}]},{"name":"output2.xml","status":"failed","meta":{"file":"results/global/output2.xml"},"items":[{"name":"testsuite-001","status":"failed","items":[{"name":"[k8s.io] Pods should be submitted and removed [NodeConformance] [Conformance]","status":"passed"},{"name":"[sig-apps] Daemon set [Serial] should rollback without unnecessary restarts [Conformance]","status":"failed"},{"name":"[sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: dir-link-bindmounted] [Testpattern: Dynamic PV (default fs)] subPath should support existing directories when readOnly specified in the volumeSource","status":"skipped"},{"name":"[sig-storage] In-tree Volumes [Driver: rbd][Feature:Volumes] [Testpattern: Pre-provisioned PV (default fs)] subPath should support restarting containers using file as subpath [Slow]","status":"skipped"}]}]}]}
```
johnSchnake: The main thing that changes in all of these is that the entire 'testsuite' is put into a new level. This way when you have N test suites, each is on their own branch and can be filtered/walked appropriately.

pkg/client/testdata/ssh.golden
```diff
@@ -44,7 +44,7 @@ data:
     sonobuoy-config:
       driver: Job
       plugin-name: e2e
-      result-format: junit
```
johnSchnake: These all need put back.

johnSchnake:
This will fail CI now because nothing uses the ginkgo reporters and they can be removed; will leave that commit out for now just for clarity.

That diff will be `16 files changed, 1664 deletions(-)`

pkg/client/results/testdata/mockResults/plugins/job-junit-02/job-junit-02.golden.json
zubron: There are a number of these golden files where it looks like the failure messages aren't being recorded. They seem to get captured correctly in `junit_good_junit.xml` file though.

johnSchnake: 💯 thanks for the catch. When developing I do tend to be a little less than critical with my review of those goldenfiles.

Probably worth it to do something like change (for tests) to MarshalIndent (with 0 indent so changes in depth don't cascade to every sub item) so that the diffs are more clean.

pkg/client/results/junit.go
johnSchnake: Update: Went ahead and added the processing to support this now too. Just seems too intuitive a field to not include

johnSchnake:
After merging the easy-json-diff PR I can see I still have a small problem here. Going to fix up.

pkg/client/results/testdata/mockResults/plugins/ds-junit-02/ds-junit-02.golden.json
```diff
@@ -43,14 +49,17 @@
...
+"items": [
+{
 "name": "[k8s.io] Pods should be submitted and removed [NodeConformance] [Conformance]",
 "status": "passed"
 },
 {
 "name": "[sig-apps] Daemon set [Serial] should rollback without unnecessary restarts [Conformance]",
 "status": "failed",
 "details": {
-"failure": "/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/framework/framework.go:696\nConformance test suite needs a cluster with at least 2 nodes.\nExpected\n    \u003cint\u003e: 1\nto be \u003e\n    \u003cint\u003e: 1\n/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/apps/daemon_set.go:385",
```
johnSchnake: This should still be here; not sure why it got removed. I see a few other tests seem to show the failure stuck around... Should be an easy fix when I find it.

johnSchnake: The problem seems to be that I was looking for the `failure.Message` but that is 0, this is the `chardata` from the xml which (in our datatype) is called `Contents`.

The fix is fast (use the new field) but raises the question about what we should do with this new field. Should we just concat the message/contents or should we save them separately? Going to take a peek at some of the other junit (non-e2e) and see what makes sense there to compare.

johnSchnake: :) So it seems they just use the opposite on the other output I've tested against (chef/inspec with junit output). They only use the message for the failure message and https://github.com/inspec/inspec/blob/a35d37043e65ce5fd528d58cd48ecb369c74a432/lib/inspec/reporters/junit.rb#L48-L54 seems to show they never even use the contents.

johnSchnake: Settled on joining message/contents together (space separated) so the data is all present. Users could still grep the data and have one main location to look still.  I thought this was better than adding a new `Details` entry for every type of attribute/field. Open to discussion though.

pkg/client/results/junit.go
```diff
@@ -45,84 +128,126 @@ func Filter(predicate func(testCase reporters.JUnitTestCase) bool, testSuite rep
...
+				status = StatusSkipped
+			}
+			testItem := Item{Name: t.Name, Status: status, Details: map[string]string{}, Metadata: map[string]string{}}
+
+			// Different JUnit implementations build the objects in slightly different ways.
+			// Some will only use contents, some only the message attribute. Here we just concat
+			// the values, separated with a space.
+			hasFailureContents := (t.Failure != nil && (t.Failure.Message != "" || t.Failure.Contents != ""))
+			if hasFailureContents {
+				testItem.Details[JUnitFailureKey] = strings.TrimSpace(t.Failure.Message + " " + t.Failure.Contents)
```
johnSchnake: This (and the lines above it) where the most recent change. Just checks that either value exists and joins them (trimming spaces at the end).

johnSchnake:
@zubron , rerequesting review since I found issues/resolved them again. Open to discussion if you dont agree with my approach.

pkg/client/results/junit.go
zubron: I think this is fine. You could also build the string up gradually by checking whether either `Message` or `Contents` was empty and appending if they aren't but I don't feel particularly strongly about it. I'm happy for you to leave it as is.

PR #857, opened 2019-08-23
Remove plugin.ResultType due to confusion/lack of need

### LINKED ISSUE #830: Plugin name/type are redundant and misleading
Run a plugin with a different name vs type.

Who knows. Sometimes we seem to use `Name` sometimes `Type` to display data to the user or report results. At worst, this leads to actual errors/failures to save files. For instance, the results are actually saved using the `Type` as a directory. Then the postprocessing uses `Name`; if the values aren't the same you'll get errors since the directory doesnt exist.

I feel like the right solution is just to remove `type` altogether. The one catch I can think of is that the `Results` type that the plugins use to report their results with doesn't even have a `Name` and only a `Type`. I think we'll just live with the mismatch there; the name/type overlap is more confusing and the `plugin.Name` is a much more natural name than `plugin.Type`.

**Environment:**

- Sonobuoy version: v0.15.1

johnSchnake:
There was redundancy between three different fields of the plugin:
name, resultType, and resultFormat. The first two existed prior
to the third and were the same in almost every case, leading
to confusion about why there were two different values at all.

The problem with this is that the aggregator server would be primarily
concerned with resultType and use it through its code when referring
to plugins. As a result, the plugin name, despite having the more
intuitive title, was actually almost never used.

This all led to a few confusing situations:
 - If a plugin name != resultType it was unclear on which query/screen
they would see which value
 - Uniqueness was only enforced on resultType and not plugin name, leading
to the confusing requirement that they provide a name but not using it often
and it not being required to be unique
 - The plugin name/type were almost always 100% the same even though their
names don't imply that, whereas plugin type/format were never meant to be
related despite sounding like similar ideas.

By removing the plugin result type (and relying on plugin name for the
unique identifier) we are left with two fields (name and format) which
seem much more clear and unique.

Internally, the aggregator server still uses the "ResultType" naming
convention in its own types since those types did not have the 'Name'
field to conflict with. In addition, that is not really a user-facing
field whereas plugin definitions (where type/name both exist) are.
Fixes #830

**Release note**:
```
Plugin definitions (e.g. output from `sonobuoy gen plugin` and input for the `--plugin` flag) no longer have a `--plugin-type` field. This was usually identical to the `plugin-name` field and only led to confusing users or causing issues if those values were not identical.
```

### DIFF
--- a/cmd/sonobuoy/app/gen_plugin_def.go
+++ b/cmd/sonobuoy/app/gen_plugin_def.go
@@ -116,7 +116,7 @@ func defaultManifest() manifest.Manifest {
 	m.Spec.Name = defaultPluginName
 	m.SonobuoyConfig.Driver = defaultPluginDriver
 	m.Spec.VolumeMounts = []v1.VolumeMount{
-		v1.VolumeMount{
+		{
 			MountPath: plugin.ResultsDir,
 			Name:      defaultMountName,
 			ReadOnly:  defaultMountReadOnly,
@@ -142,9 +142,6 @@ func genPluginDefWrapper(cfg *GenPluginDefConfig) func(cmd *cobra.Command, args 
 // genPluginDef returns the YAML for the plugin which Sonobuoy would expect as
 // a configMap in order to run/gen a typical run.
 func genPluginDef(cfg *GenPluginDefConfig) ([]byte, error) {
-	// Result type just duplicates the name in most cases.
-	cfg.def.SonobuoyConfig.ResultType = cfg.def.SonobuoyConfig.PluginName
-
 	// Copy the validated value to the actual field.
 	cfg.def.SonobuoyConfig.Driver = cfg.driver.String()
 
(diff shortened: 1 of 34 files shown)


johnSchnake:
I want to run some manual tests, but I am pleasantly surprised by how easy this turned out to be and how simple a review this looks to be 💯

zubron:
Nice to see this removed. I'm assuming that for any existing plugins, the `result-type` field will just be ignored?

johnSchnake:
Right; if someone has a saved plugin def it will be ignored and in most cases the value is identical to the plugin-name which will remain.

johnSchnake:
👍 I'll still do a few more tests but I wanted to see where approval was too. I see that I put WIP remarks in the notes so I realize now why you didnt finalize your review.

johnSchnake:
Just the known pid issue; rerunning ci for safety then merging. Manual testing went well. I used a mix of plugins which provided the old field and some which did not. Was able to run and process results as expected for all of them.

PR #862, opened 2019-08-30
Update version and add new docs for v0.15.3

zubron:


zubron:
I applied the same changes mentioned here: https://github.com/heptio/sonobuoy/pull/846/files#issuecomment-522128532.

I might take a quick look at that later this afternoon.

PR #866, opened 2019-09-04
Bubble up unknown results just like failures

### LINKED ISSUE #848: Unknown results should bubble up like failures
I made a plugin that produced xml output but if failed to be parsed as xml.

Instead of marking that as a failure, we mark it as `unknown` since we dont know about the underlying tests at all. However, you can see the output here:
```
name: myPlugin
status: passed
items:
- name: output.xml
  status: unknown
  meta:
    error: expected element type <testsuite> but have <testsuites>
    file: results/global/tmp/results/output.xml
```

The unknown does not bubble up and so the top level item defaults to passed (no failures).

We should have unknowns bubble up as well, but failures would obviously override that and lead to failure.

Results to lead to `unknown` at the top level.

The bug listed there also is being reported in another ticket. Our junit processing is too specific to the output produced by `e2e` and so this (valid) junit failed to get parsed correctly.

**Environment:**

- Sonobuoy version: v0.15.2

johnSchnake:
When doing post processing if a plugin (or file) results
in an 'unknown' status, it should bubble up according
to the following rules:
 - if any failures exist, failure bubbles up
 - if any unknowns exist (and no failures), unknown bubbles up
 - pass otherwise

Fixes #848

**Release note**:
```
When looking at the results of plugins, Sonobuoy will now bubble up unknown values. If one result is unknown and others pass, the aggregate is unknown. If one result is unknown and any others failure, the aggregate result is still failure.
```

### DIFF
--- a/pkg/client/results/processing.go
+++ b/pkg/client/results/processing.go
@@ -97,6 +97,7 @@ func aggregateStatus(items ...Item) string {
 		return StatusUnknown
 	}
 
+	unknownFound := false
 	for i := range items {
 		// Branches should just aggregate their leaves and return the result.
 		if len(items[i].Items) > 0 {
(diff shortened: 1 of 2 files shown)


johnSchnake:
I've deleted the `signed-off-by` webhook (in favor of the DCO one) but it still seems to be used on this PR. Will monitor this.

Issue #735, opened 2019-05-31
Customizable webhooks for status updates


johnSchnake:
**Describe the solution you'd like**
Allow the user to specify a series of webhooks/subscriptions which certain events will post back too. Examples:
 - subscribe to overall status events (plugin starts/stops/fails or the run is over in general) (similar to #585 )
 - subscribe to each progress update for each plugin (if we complete #354 )
 - subscribe to get the results tarball at the very end (similar to use for #39 )
 - a way to subscribe to get not just the status but maybe even the "results" of each plugin if such a thing is possible.

**Anything else you would like to add:**
Webhooks we be a pretty generic way to allow other tools to be built in/around Sonobuoy without having to vendor Sonobuoy itself and use the internal methods.

It would be great to do something like:
 - launch a run in CI
 - the results automatically get put into S3
 - the final status gets passed to a slackbot and posted into a team channel

Implementing 1 webhook isn't much easier than supporting multiple, we just have to allow the user to specify them and then fire off the info as we get it.

vladimirvivien:
cc @wilsonehusin this maybe worth revisiting as well.

PR #867, opened 2019-09-06
Adjust MASTER_URL to be just the hostname

johnSchnake:
In preparation of supporting status updates, I want to clarify
the meaning of the MASTER_URL env var on the worker.

Up until now the only URL of importance was the results URL so
that is what the MASTER_URL was set to. Soon, we will have results
and status updates, so it makes sense to have the MASTER_URL simply
be the hostname and add the path on the fly depending on which
endpoint we are trying to hit.

ref #735

**Release note**:
```
MASTER_URL set on the Sonobuoy worker container now refers to just `https://<hostname>` instead of the full results URL in order to more easily support manipulating the path used. This should have no impact on users since this is just a value used internally.
```

### DIFF
--- a/cmd/sonobuoy/app/worker.go
+++ b/cmd/sonobuoy/app/worker.go
@@ -21,14 +21,17 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"net/http"
+	"net/url"
 	"os"
 	"os/signal"
+	"path"
 	"strings"
 	"syscall"
 	"time"
 
 	"github.com/heptio/sonobuoy/pkg/errlog"
 	"github.com/heptio/sonobuoy/pkg/plugin"
+	"github.com/heptio/sonobuoy/pkg/plugin/aggregation"
 	"github.com/heptio/sonobuoy/pkg/worker"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


pkg/plugin/driver/base.go
```diff
@@ -31,6 +31,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// EnvKey is the env value expected to be set in order to inform the worker env about what mode it is in.
+	// Technically this is also discernable via the command invoked, but this places the
+	// information in the standard env location.
+	EnvKey = "DRIVER"
```
johnSchnake: I'm not certain I really need this. I originally thought I needed it but when starting up the sonobuoy worker we use a different command for each type of (global/node-specific) and _that_ is what triggers the different logic, not anything abou the URL.

I don't think this value is bad to have, but it does complicate the PR a bit. I think I'll remove it.

cmd/sonobuoy/app/worker.go
```diff
@@ -35,6 +37,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	pathResultsByNode = "/api/v1/results/by-node"
+	pathResultsGlobal = "/api/v1/results/global"
```
johnSchnake: I could have tried to export these from the aggregator's package, but the value that those really need has a parameterized suffix (e.g. `/api/v1/results/by-node/{node}/{type}`) and it felt a bit forced to add the new values, export them, and use them to build the existing values by adding back the suffix.

At the time I thought this was a bit overkill, but and it would be normal for a client to reference the paths of things like that.

Update pending...

PR #868, opened 2019-09-09
Show logs for a specific plugin

### LINKED ISSUE #702: Sonobuoy logs should allow specification of a single plugin
We should add an optional selection argument that enables the user to get just the data they want or perhaps even default the plugin to e2e... --plugin=e2e

From splitting #301

zubron:
Add a new flag to the `logs` command to show the logs only for a
specific plugin. If not provided, or no pods are found for the plugin,
default to showing all logs.

We didn't have a way to uniquely identify pods for a plugin, so this
change adds a new label when creating Pods and DaemonSets which can
be used to filter the pods. This means that the flag will not work
for older versions of Sonobuoy, but it will still default to showing
all logs which was the original behaviour.

- Fixes #702 

**Release note**:
```
Sonobuoy can now show only the logs for a specific plugin by using the flag `--plugin <plugin_name>` when using the logs command.
```

### DIFF
--- a/cmd/sonobuoy/app/logs.go
+++ b/cmd/sonobuoy/app/logs.go
@@ -35,6 +35,7 @@ const (
 var logFlags struct {
 	namespace  string
 	follow     bool
+	plugin     string
 	kubeconfig Kubeconfig
 }
 
@@ -52,6 +53,7 @@ func NewCmdLogs() *cobra.Command {
 	)
 	AddKubeconfigFlag(&logFlags.kubeconfig, cmd.Flags())
 	AddNamespaceFlag(&logFlags.namespace, cmd.Flags())
+	cmd.Flags().StringVarP(&logFlags.plugin, pluginFlag, "p", "", "Show logs for a specific plugin")
 	return cmd
 }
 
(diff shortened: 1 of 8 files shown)


johnSchnake:
Seems like that `sonobuoy-plugin` label got removed accidentally when we moved away from template files. https://github.com/heptio/sonobuoy/pull/834/files#diff-90fec45294302b8f99ef89e826ce7e9bL30

LGTM but I'm going to run it manually to double check and then I'll approve.

zubron:
> Seems like that `sonobuoy-plugin` label got removed accidentally when we moved away from template files. https://github.com/heptio/sonobuoy/pull/834/files#diff-90fec45294302b8f99ef89e826ce7e9bL30

The `sonobuoy-plugin` line shown in that commit was for an annotation. We still have those annotations (https://github.com/heptio/sonobuoy/blob/master/pkg/plugin/driver/daemonset/daemonset.go#L88) but the docs for annotations say that they shouldn't be used for identifying objects so that's why we have to now include the label as well. Still means that this new client won't work with the older versions :(

PR #869, opened 2019-09-09
Status updates

### LINKED ISSUE #354: RFE: Implement progress for e2e tests
Something simple we could do is parse the e2e logs (they tell us how many things will run and when things finish). The thing could update the status metadata with some extra % information.

A more long-term solution would be to integrate something directly with ginkgo. This would be good for the extended tests as well.

johnSchnake:
Adds functionality to the aggregator and the workers in order to
allow plugins to report their progress.

 - A new configuration value ProgressUpdatesPort is added to the config
which controls which port the plugins and the worker should
communicate on (via localhost)
 - The worker launches a server listening on that port and relays
the status updates to the aggregator
 - The aggregator puts that information onto the aggregators pod
annotation so that users can see it via `sonobuoy status --json`
or via kubectl

Fixes #354

**Release note**:
```
Adds progress update support for plugins. Plugins communicate their progress to the Sonobuoy worker which is running as a sidecar with it via localhost on a port specified by the environment variable `SONOBUOY_PROGRESS_PORT` (default 8099). This port can be modified via the config value 
```


pkg/plugin/aggregation/aggregator.go
```diff
@@ -157,13 +163,13 @@ func (a *Aggregator) isComplete() bool {
 	return true
 }
 
-func (a *Aggregator) isResultExpected(result *plugin.Result) bool {
-	_, ok := a.ExpectedResults[result.ExpectedResultID()]
+func (a *Aggregator) isResultExpected(obj keyer) bool {
```
johnSchnake: Should rename this to just `isExpected` since now it applies to status updates as well.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -157,13 +163,13 @@ func (a *Aggregator) isComplete() bool {
...
+	_, ok := a.ExpectedResults[obj.Key()]
 	return ok
 }
 
-func (a *Aggregator) isResultDuplicate(result *plugin.Result) bool {
-	_, ok := a.Results[result.ExpectedResultID()]
+func (a *Aggregator) isResultDuplicate(obj keyer) bool {
```
johnSchnake: This name is fine as-is since it really is specific to results; I don't care about duplicate status updates. Probably just need to revert these changes on this method; I think I made them just so it would build at some point.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -198,21 +204,47 @@ func (a *Aggregator) processResult(result *plugin.Result) error {
...
+func (a *Aggregator) processStatus(status plugin.StatusUpdate) error {
+	a.resultsMutex.Lock()
+	defer a.resultsMutex.Unlock()
+
+	// jschnake need to define how this expected stuff works. Either preload another
+	// expected status stuff or utilize the results work. but the results work expects
+	// the 'global' to be known due to a different URL path which we dont have.
+	// Seems like we'd have to say either the key (name+node) or it is (name+global)? Kind of wonky though
+	// and what if you get name/node and name/node2 but expected name+global? Or do we push that
+	// understanding down into the worker? Have a new env that says it is part of a job/ds?
```
johnSchnake: Old comment; can be removed. It now . will check about `expected` but the URL the plugins write to has the node/global in it so there is no ambiguity. I just followed the existing pattern used for results.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -244,6 +276,29 @@ func (a *Aggregator) HandleHTTPResult(result *plugin.Result, w http.ResponseWrit
...
+		switch t := err.(type) {
+		case *httpError:
+			logrus.Errorf("Status processing error (%v): %v", t.HttpCode(), t.Error())
+			http.Error(
+				w,
+				t.Error(),
+				t.HttpCode(),
+			)
+		default:
+			logrus.Errorf("status processing error (%v): %v", http.StatusInternalServerError, t.Error())
```
johnSchnake: nit: capitalization `Status`

pkg/worker/config.go
```diff
@@ -52,6 +57,7 @@ func LoadConfig() (*plugin.WorkerConfig, error) {
 	viper.BindEnv("cacert", "CA_CERT")
 	viper.BindEnv("clientcert", "CLIENT_CERT")
 	viper.BindEnv("clientkey", "CLIENT_KEY")
+	viper.BindEnv("statusport", "SONOBUOY_STATUS_PORT")
```
johnSchnake: We should probably start prefixing our env vars with `SONOBUOY` for clarity/consistency. This is something that should be able to be set by the user to avoid potential conflicts so the worker and the plugin need to get that value from their env.

Makefile
```diff
@@ -24,7 +24,7 @@ REGISTRY ?= gcr.io/heptio-images
 IMAGE = $(REGISTRY)/$(TARGET)
 DIR := ${CURDIR}
 DOCKER ?= docker
-LINUX_ARCH := amd64 arm64
+LINUX_ARCH := amd64
```
johnSchnake: dont commit this; just speeding up local builds

johnSchnake:
TODO:
 - [ ] Do something with the given status updates; currently, we just log them but we want to surface them to the user
 - [ ] Add an example plugin which submits these status updates. I have a one for personal testing, but I should make one that is more focused on just the status updates (like a ticker/countdown plugin or something)

johnSchnake:
Codecov gets a bit confused with the force pushes; the link to the most recent commit is here: https://codecov.io/gh/heptio/sonobuoy/commit/d9ba95a9ac626ccaca2f4e8a26cafdf1c2080811
which shows a diff coverage of 42% (not the 28% shown). It reduces the overall coverage by 0.11%. I may add a bit more to ensure we are moving coverage upward but, consistent with our past practices, I don't want to get this stuck making trivial tests just for coverage sake. I'm just using it as a baromter of whether or not I'm testing the reasonable logic centers of the code.

pkg/plugin/aggregation/update_test.go
```diff
@@ -210,3 +211,114 @@ func TestGetAggregatorPodName(t *testing.T) {
...
+				Status: "running",
+			},
+			results: map[string]*plugin.Result{
+				"type1/node1": {NodeName: "node1", ResultType: "type1"},
+			},
+			updates: map[string]*plugin.StatusUpdate{
+				"type1/node1": {PluginName: "type1", Node: "node1", Status: "new status"},
+			},
+		}, {
+			desc: "Non-overlapping updates and results",
```
johnSchnake: TODO: 
 - [ ] test with failing result
 - [ ] test with failures and errors in the status update

pkg/plugin/aggregation/aggregator.go
```diff
@@ -198,21 +212,45 @@ func (a *Aggregator) processResult(result *plugin.Result) error {
...
 	// Upon success, we no longer want to keep processing duplicate results.
-	delete(a.FailedResults, result.ExpectedResultID())
+	delete(a.FailedResults, result.Key())
+
+	return nil
+}
+
+// processStatus is the main aggregator logic for handling the status updates from plugins.
+// We first
+func (a *Aggregator) processStatus(status plugin.StatusUpdate) error {
```
johnSchnake: Add a simple test or two for this; ensuring what/how we check for expected values and handle global nodes, erroring, etc

pkg/plugin/aggregation/aggregator.go
```diff
@@ -244,6 +282,30 @@ func (a *Aggregator) HandleHTTPResult(result *plugin.Result, w http.ResponseWrit
 	}
 }
 
+// HandleHTTPStatus wraps the aggregators processStatus method in such a way as to respond
+// with appropriate logging and HTTP codes.
+func (a *Aggregator) HandleHTTPStatus(status plugin.StatusUpdate, w http.ResponseWriter) {
```
johnSchnake: Will add a processstatus test already; may want to add a test here to ensure we properly continue to report the errors/codes as intended.

pkg/plugin/aggregation/handler.go
```diff
@@ -97,6 +126,19 @@ func resultFromRequest(r *http.Request, muxVars map[string]string) *plugin.Resul
 	return result
 }
 
+func statusFromRequest(r *http.Request, muxVars map[string]string) (plugin.StatusUpdate, error) {
+	var update plugin.StatusUpdate
+	dec := json.NewDecoder(r.Body)
+	err := dec.Decode(&update)
+	update.Node = muxVars["node"]
```
zubron: In the case where the decode fails and there is an error, is the resulting `update` object used?

pkg/plugin/aggregation/handler_test.go
```diff
@@ -232,3 +239,68 @@ func TestFilenameFromHeader(t *testing.T) {
...
 	}
 }
+
+func TestStatusFromRequest(t *testing.T) {
+	tcs := []struct {
+		desc        string
+		input       io.Reader
+		vars        map[string]string
+		expected    *plugin.StatusUpdate
+		expectedErr string
```
zubron: Will more test cases get added here to check for error cases?

pkg/plugin/aggregation/status.go
```diff
@@ -123,3 +139,28 @@ func GetStatus(client kubernetes.Interface, namespace string) (*Status, error) {
 
 	return &status, nil
 }
+
+// SetStatus sets the status on the pod via an annotation. It will overwrite the
+// existing status.
+func SetStatus(client kubernetes.Interface, namespace string, status *Status) error {
```
zubron: Perhaps this should include Pod in its name, `SetPodStatus`?

pkg/plugin/aggregation/status.go
```diff
@@ -49,6 +51,8 @@ type PluginStatus struct {
 
 	ResultStatus       string         `json:"result-status"`
 	ResultStatusCounts map[string]int `json:"result-counts"`
+
+	Progress *plugin.StatusUpdate
```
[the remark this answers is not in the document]
johnSchnake: 👍 Thanks for catching this; I ran across seeing that in the json but didnt remember to take a note to fix it.

pkg/plugin/aggregation/handler.go
johnSchnake: It shouldn't be (by convention if you get the err the value is not very meaningful), but the method is just a few lines long and avoids an unnecessary indentation by  just returning the object/wrapped error. We have to return an object anyways (even on the error case) so it doesn't matter if we do/dont set these values.

pkg/plugin/aggregation/status.go
johnSchnake: 👍 I'll fix up the name. Yes the term status is very overloaded. I had just copied this from elsewhere and I think it may be able to be removed from the diff now; I'll double check. Either way I'll clarify the name.

johnSchnake:
@zubron I agree with all the naming concerns. I started calling them statusUpdates but then as I kept working on it I kept running more and more into things like `updater.Status.Status` and so I started using `Progress` to differentiate it. This is accurate and distinct so I think I should go back and update them all.

So there is `Status` which is the aggregate pod status, `PluginStatus` objects which are each individual status of the plugin which includes the `Progress` object we're creating.

Somewhat confusing is the `PluginStatus` because it originally (in the long long ago time) had `status` which meant completed/running/failed but then we added the concept of _actual_ results such as passed/failed so we called that `resultStatus`. In addition, the `Progress` object we're adding has a catchall string which we called `status`. Maybe I'll just update that to be `msg` since it really is just a catchall for plugin authors to use as they want really.

pkg/plugin/aggregation/update_test.go
johnSchnake: Glad I did these; found a bug in the code that prevented some data from being reported.

pkg/plugin/aggregation/aggregator.go
```diff
@@ -198,21 +212,45 @@ func (a *Aggregator) processResult(result *plugin.Result) error {
...
-	delete(a.FailedResults, result.ExpectedResultID())
+	delete(a.FailedResults, result.Key())
+
+	return nil
+}
+
+// processStatus is the main aggregator logic for handling the status updates from plugins.
+// We first
+func (a *Aggregator) processStatus(status plugin.StatusUpdate) error {
+	a.resultsMutex.Lock()
```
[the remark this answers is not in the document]
johnSchnake: I wouldn't jump at the chance to worry about this right now. The other logic that exists for results handling locks/unlocks the mutex at a higher level so that while it is handling a given result (not just checking for duplicates/etc) no other results "cut in line".

If I moved mutex handling down into those low level calls then we lose the ability to do higher level locking without hitting conflicts.

pkg/plugin/aggregation/handler_test.go
johnSchnake: Added one 👍

johnSchnake:
Lots of renames and a fair number of new tests added. Ready for a final review.

pkg/plugin/aggregation/handler.go
```diff
@@ -37,47 +39,74 @@ const (
 	PathResultsByNode = "/api/v1/results/by-node"
 
 	// PathResultsGlobal is the path for global (non-node-specific) results to be PUT to. Callers should
-	// add one path elements as a suffix to this to specify the plugin name (e.g. `<path>/plugin`)
+	// add one path element as a suffix to this to specify the plugin name (e.g. `<path>/plugin`)
 	PathResultsGlobal = "/api/v1/results/global"
 
+	// PathStatusByNode is the path for node-specific status to be POSTed to. Callers should
+	// add two path elements as a suffix to this to specify the node and plugin (e.g. `<path>/node/plugin`)
+	PathStatusByNode = "api/v1/status/by-node"
```
zubron: I thought that the API path for this would also have been changed. Did you want to keep it as `status`?

examples/plugins/progress-reporter/run.sh
```diff
@@ -0,0 +1,43 @@
...
+
+results_dir="${RESULTS_DIR:-/tmp/results}"
+progress_port="${SONOBUOY_PROGRESS_PORT:-8099}"
+total_updates=60
+sleep_seconds=1
+
+# saveResults prepares the results for handoff to the Sonobuoy worker.
+# See: https://github.com/heptio/sonobuoy/blob/master/docs/plugins.md
+saveResults() {
+    #cd ${results_dir}
```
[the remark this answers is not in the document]
johnSchnake: 👍 Thanks I must not have carefully looked at this file. It was copied from the easy-cmd example and just tweaked to work.

pkg/plugin/aggregation/handler.go
johnSchnake: No lets change that as well.

johnSchnake:
Updated based on comments, will check back tomorrow and do a self review too. Thanks @zubron for all the comments so far.

pkg/plugin/aggregation/handler.go
```diff
@@ -109,6 +151,23 @@ func (h *Handler) resultsHandler(w http.ResponseWriter, r *http.Request) {
 	r.Body.Close()
 }
 
+func (h *Handler) statusHandler(w http.ResponseWriter, r *http.Request) {
```
zubron: Sorry to point out another naming thing, but for consistency I think we should use `progress` in this file. Especially in the log messages.

pkg/config/config.go
```diff
@@ -161,6 +163,9 @@ type Config struct {
 	ImagePullPolicy   string            `json:"ImagePullPolicy" mapstructure:"ImagePullPolicy"`
 	ImagePullSecrets  string            `json:"ImagePullSecrets" mapstructure:"ImagePullSecrets"`
 	CustomAnnotations map[string]string `json:"CustomAnnotations,omitempty" mapstructure:"CustomAnnotations"`
+
+	// ProgressUpdatesPort is the port on which the Sonobuoy worker will listen for status updates from its plugin.
+	ProgressUpdatesPort string `json:"ProgressUpdatesPort,omitempty" mapstructure:"StatusUpdates"`
```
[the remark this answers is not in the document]
zubron: This conversation was resolved but it looks like this still needs to be updated. There was a commit in between which changed this line. Did the change get removed again?

johnSchnake:
@zubron Thanks for all the reviews; I appreciate all the naming finds you pointed out. `status` is a hard word to grep against to find meaningful updates.

I've since updated the places you pointed out and also update the example a bit more and fixed some typos in the example README.

Rebased as well and so I'll let this go green on CI and then merge.

pkg/config/config.go
```diff
@@ -161,6 +163,9 @@ type Config struct {
 	ImagePullPolicy   string            `json:"ImagePullPolicy" mapstructure:"ImagePullPolicy"`
 	ImagePullSecrets  string            `json:"ImagePullSecrets" mapstructure:"ImagePullSecrets"`
 	CustomAnnotations map[string]string `json:"CustomAnnotations,omitempty" mapstructure:"CustomAnnotations"`
+
+	// ProgressUpdatesPort is the port on which the Sonobuoy worker will listen for status updates from its plugin.
+	ProgressUpdatesPort string `json:"ProgressUpdatesPort,omitempty" mapstructure:"ProgressUpdatesPort"`
```
johnSchnake: The line has been updated here; not sure why it wasn't in the line you commented on. I'm unable to reply to your comment as well since I think it is on an old commit that isn't referenced now.

PR #874, opened 2019-09-12
Update contributor copyright

johnSchnake:


johnSchnake:
@emuetzel It was originally `"the Heptio Sonobuoy project contributors".` and I just removed the `Heptio`.

For a while it seemed we marked a handful of files as `the Sonobuoy contributors` but I don't think there was strong, clear direction other than not to say `Heptio` any longer and to say `contributors`

emuetzel:
@timh can you confirm that "the Sonobuoy project contributors" is the correct copy? If so, this looks good to me.

johnSchnake:
Looking into SPDX headers as well; trying to do the right thing and do it just once instead of thrashing on this license detail. In a holding pattern for the moment until we decide how to proceed.

johnSchnake:
I think the spdx issue is somewhat tangential, though I'd still like to do them at the same time.

See https://github.com/torvalds/linux/commit/acc962c5140071f251c62f360cb6668c1962ff31#diff-35ffec5ead8756cae8d1726695eef91dR1-R2
for an example. We'd get rid of the header text in each file, but we'd still need to specify the copyright holders/year.

johnSchnake:
```
Copyright notices
SPDX IDs are intending to express information about licenses. Copyright notices ‐ statements about who owns the copyright in a file or project ‐ are outside the scope of SPDX short-form IDs.

Therefore, you should not remove or modify existing copyright notices in files when adding an SPDX ID.
```
https://spdx.org/ids-how

johnSchnake:
That page also says 

> Like copyright notices, existing license texts and notices should be retained, not replaced ‐ especially a third party's license notices

but it seems within our right to replace them all for our own files. Just a bit of housekeeping IMO but guidance of norms would be great

jonasrosland:
This is interesting, I like the SPDX approach to clean things up.

PR #878, opened 2019-09-13
Add a disruptive taint to list of default tolerations

### LINKED ISSUE #876: Sonobuoy must tolerate certain test taints
See https://github.com/kubernetes/kubernetes/pull/82664 for some context; lots of conversations in slack.

TL;DR; we may allow some conformance tests to be also disruptive; specifically there are some w.r.t. taints that were causing evictions on a node and impacting sonobuoy. We need to tolerate these taints (or all taints?) in order to be sure that we do not get evicted during a test run.

Sonobuoy MUST be able to run all conformance tests (and, ideally, all of the e2e suite); not get evicted due to a taint added during a test.

**Environment:**

- Sonobuoy version: v.1.15.3
- Kubernetes version: (use `kubectl version`): v1.16.0.beta.3

johnSchnake:
A new conformance test in v1.16 will add a taint to a node which
can impact the sonobuoy aggregator pod as well as the e2e pod.

This change adds the toleration to both the default job podspec
and the aggregator. The default daemonset podspec already tolerates
all taints.

Fixes #876

I created a v1.16 cluster which may not be strictly necessary. You may just need to target the newer tests:

```
sonobuoy run --e2e-focus "removing taint cancels eviction" --kube-conformance-image-version=v1.16.0-rc.2 --sonobuoy-image schnake/sonobuoy:testTaint
```

**Release note**:
```
Added a new toleration to jobs plugins and the Sonobuoy aggregator in order to avoid disruptive effects of some E2E tests.
```

### DIFF
--- a/cmd/sonobuoy/app/testdata/pluginDef-default-podspec.golden
+++ b/cmd/sonobuoy/app/testdata/pluginDef-default-podspec.golden
@@ -8,6 +8,8 @@ podSpec:
     operator: Exists
   - key: CriticalAddonsOnly
     operator: Exists
+  - key: kubernetes.io/e2e-evict-taint-key
+    operator: Exists
 sonobuoy-config:
   driver: ""
   plugin-name: "n"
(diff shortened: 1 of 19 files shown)


zubron:
CI is failing due to a unit test which is checking the tolerations: https://github.com/heptio/sonobuoy/blob/master/pkg/plugin/driver/job/job_test.go#L209-L214

johnSchnake:
Thanks @zubron Fixed that test, I hadn't noticed it. Will wait until ci goes green and merge.

PR #885, opened 2019-09-17
Add certified conformance mode and non-disruptive-conformance mode

### LINKED ISSUE #875: Consider removal of `E2E_SKIP` default value
Upstream made an effort to ensure all conformance tests were properly tagged so that you didnt need to skip any tests to be a conformance run.

To support the older set of tests we delayed updating this so we still provide unnecessary env on the e2e test run that can be misleading (i.e. it implies that there are potentially Alpha conformance tests, or flakey conformance tests).

We will have to double-check what versions of k8s we support and when the upstream 'fixed' all the tests so conformance w/o a SKIP value was appropriate.

### LINKED ISSUE #877: Sonobuoy must handle conformance+live cluster differently than full conformance
The CNCF conformance instructions do tell people to run `sonobuoy run` to get results for certification. 

Historically sonobuoy ran the conformance tag but skipped alpha, disruptive, flakey, etc. That was all fine since they _should_ have been exclusive. Now that those overlap some (if https://github.com/kubernetes/kubernetes/pull/82664 is updated based on the peer feedback), we'll have to accommodate both flows:
 - if a user wants results for certification, do they run `sonobuoy run` and get potentially destructive tests? These wouldn't be completely safe to run in a prod cluster.
 - if a user wants to test their running cluster, what do they run?

Sonobuoy will make this pretty simple since we have "modes" for the e2e plugin. Currently, we have the Conformance and Quick modes.

The new mode naming will need some thought:
 - LiteConformance vs Conformance
 - NonDestructive vs Conformance

And we'll need to decide which is the default

johnSchnake:
With K8s v1.16, disruptive tests may not be part of the conformance
suite. As a result, we need to ensure that user workloads are
protected by default while still allowing for CNCF certification
test runs.

In this PR we:
 - rename the default `conformance` mode to be `non-disruptive-conformance`
and modified the skip list appropriately.
 - added a `certified-conformance` mode which does not provide a skip list
and will run even the disruptive tests
 - tweaked the naming to use lower case values by default to be more
consistent with our other flags/naming

Fixes #877
Fixes #875

I could have written the skip list in a slightly different way, naming the tests individually, but instead chose a higher level term that filtered both and, IMO, seems more clear why they may be skipped.

However, we need to talk about this issue in general because we are actually in a situation now where the `E2E_SKIP` value is dependent on the version of the cluster :( I think we can get away with this because the only real risks are:
 - someone uses this version of Sonobuoy with an older cluster
    - the non-disruptive-conformance run should hit all the conformance tests anyways since the NoExecuteTaintManager tests were not part of conformance on older clusters
 - someone uses this version of Sonobuoy with a newer cluster
    - the non-disruptive-conformance run may skip the NoExecuteTaintManager tests, even if the tests get updated somehow to not be disruptive or other NoExecuteTaintManager tests which are non-disruptive get added to Conformance. The only downside here is that a few tests don't get run, but since these are non-certification runs it doesn't matter much.

If there is a situation I've failed to think of, they can still set the e2e-focus/e2e-skip values on the CLI or the env vars in the YAML directly so we are not a blocker.

**Release note**:
```
Default `sonobuoy run` commands will run a subset of the `Conformance` tests which are safe to run on clusters with production workloads (avoids running known disruptive tests). To generate results for CNCF certification, you must run `sonobuoy run --mode=certified-conformance`
```

### DIFF
--- a/pkg/client/mode.go
+++ b/pkg/client/mode.go
@@ -31,17 +31,26 @@ type Mode string
 
 const (
 	// Quick runs a single E2E test and the systemd log tests.
-	Quick Mode = "Quick"
+	Quick Mode = "quick"
 
-	// Conformance runs all of the E2E tests and the systemd log tests.
-	Conformance Mode = "Conformance"
+	// NonDisruptiveConformance runs all of the `Conformance` E2E tests which are not marked as disuprtive and the systemd log tests.
+	NonDisruptiveConformance Mode = "non-disruptive-conformance"
+
+	// CertifiedConformance runs all of the `Conformance` E2E tests and the systemd log tests.
+	CertifiedConformance Mode = "certified-conformance"
 )
 
-const defaultSkipList = `Alpha|\[(Disruptive|Feature:[^\]]+|Flaky)\]`
+// nonDisruptiveSkipList should generally just need to skip disruptive tests since upstream
+// will disallow the other types of tests from being tagged as Conformance. However, in v1.16
+// two disruptive tests were  not marked as such, meaning we needed to specify them here to ensure
+// user workload safety. See https://github.com/kubernetes/kubernetes/issues/82663
+// and https://github.com/kubernetes/kubernetes/issues/82787
+const nonDisruptiveSkipList = `\[Disruptive\]|NoExecuteTaintManager`
 
 var modeMap = map[string]Mode{
-	string(Conformance): Conformance,
-	string(Quick):       Quick,
+	string(NonDisruptiveConformance): NonDisruptiveConformance,
+	string(Quick):                    Quick,
+	string(CertifiedConformance):     CertifiedConformance,
 }
 
 // ModeConfig represents the sonobuoy configuration for a given mode.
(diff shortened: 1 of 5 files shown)


pkg/client/mode.go
```diff
@@ -31,17 +31,25 @@ type Mode string
...
+
+	// CertifiedConformance runs all of the `Conformance` E2E tests and the systemd log tests.
+	CertifiedConformance Mode = "certified-conformance"
 )
 
-const defaultSkipList = `Alpha|\[(Disruptive|Feature:[^\]]+|Flaky)\]`
+// nonDisruptiveSkipList should generally just need to skip disruptive tests since upstream
+// will disallow the other types of tests from being tagged as Conformance. However, in v1.16
+// two disruptive tests were  not marked as such, meaning we needed to specify them here to ensure
+// user workload safety.
```
zubron: A link to the relevant issues/PRs/discussions here would be useful for coming back to this in the future.

johnSchnake: 👍 Added.

PR #886, opened 2019-09-17
Bump version and add v0.15.4 docs

johnSchnake:


johnSchnake:
@zubron appreciate just a basic sanity check on this, notably the links in the readme.

I created #887 since I found I was still having to update readme links (and I updated some in the older versions of the docs I found too that needed updating).

Issue #668, opened 2019-04-10
Expand integration test suite


johnSchnake:
When the upstream image for e2e was broken on the unhappy path (failed to report results) would not report results and Sonobuoy would hang.

1) We should fix Sonobuoy so that we detect the container failed and handle it appropriately (ref #667)
2) We should test the e2e plugin specifically for this unhappy path, ensuring the situation is handled properly.

IMO "handled properly" means that if the donefile is missing but the container stopped; we should tar up anything in the results directory and report that as the results along with an error message stating that the container exited and providing any information we can about it (exit code, status message, etc)

johnSchnake:
Also useful for testing queries filter/gather data correctly and items we create in the cluster are related appropriately (e.g. for cascading deletions)

johnSchnake:
Would be useful to have a go integration suite that can launch a series of sonobuoy runs to test various behaviors.

johnSchnake:
Adding to the next milestone; main priority is setting up the framework and documenting what we _want_ to test in this way. Even if, for now, those tests are noops or commented out, lets shoot for being able to run them.

johnSchnake:
Breadcrumb while its on my mind: we can slightly incorporate our examples into this. By that I mean that we have a sample command runner plugin (which returns a tarball). I would imagine we should also make a plugin that is similar that returns a single file (with known time, e.g. json).

This way we can make a big test run and see all (or as many as possible) of the combinations:
job vs daemonset
tarball vs single file results
passing/skipped/failing tests (junit)
plugins failing to launch/complete/report results

johnSchnake:
First PR has merged; now we can spend some time to actually make some useful tests :) Actually looking forward to getting this extra coverage

johnSchnake:
The test image has been created as well as a basic test using sonobuoy and a live cluster. Going to close this for now with the understanding that more tickets may be created for specific sets of tests that we want to add.

PR #894, opened 2019-09-18
Add integration test to check version information

zubron:
This adds a new package, `integration`, which is the starting point
for adding integration testing for Sonobuoy. The package only contains
one test so far which runs `sonobuoy version` and checks that all fields
in the output from `version` are non-empty.

To separate these tests, a new Makefile target has been added which
sets the appropriate flags for running these tests. They are currently
run conditionally with the tag `integration` under the assumption that
as this package grows, we may not want to run these tests every time.

- Part of #668 

You can verify this test by building the sonobuoy executable with your
local go toolchain then running it:
```
go build
go test test/integration/sonobuoy_integration_test.go
```
This test should fail since the GitSHA information is not set.

To see it pass, build the executable using the Makefile and then run
the test again.
```
rm sonobuoy && make build_sonobuoy
go test test/integration/sonobuoy_integration_test.go
```

### DIFF
--- a/test/integration/sonobuoy_integration_test.go
+++ b/test/integration/sonobuoy_integration_test.go
@@ -1,0 +1,71 @@
+// +build integration
+
+package integration
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+)
+
+const (
+	defaultSonobuoyPath = "../../sonobuoy"
+)
+
+var (
+	// Path to the Sonobuoy CLI
+	sonobuoy string
+)
+
+func findSonobuoyCLI() (string, error) {
+	sonobuoyPath := os.Getenv("SONOBUOY_CLI")
+	if sonobuoyPath == "" {
+		sonobuoyPath = defaultSonobuoyPath
+	}
+	if _, err := os.Stat(sonobuoyPath); os.IsNotExist(err) {
+		return "", err
+	}
+
+	return sonobuoyPath, nil
+}
+
+// TestSonobuoyVersion checks that all fields in the output from `version` are non-empty
+func TestSonobuoyVersion(t *testing.T) {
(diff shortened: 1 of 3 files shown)


johnSchnake:
The other PR merged so this can be rebased

test/integration/sonobuoy_integration_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func TestMain(m *testing.M) {
+	var err error
+	sonobuoy, err = findSonobuoyCLI()
+	if err != nil {
+		fmt.Printf("Skipping integration tests: failed to find sonobuoy CLI: %v\n", err)
+		os.Exit(1)
+	}
+
+	flag.StringVar(&gitSHA, "git-sha", os.Getenv("GIT_SHA"), "SHA of HEAD Git commit (git rev-parse --verify HEAD)")
```
johnSchnake: nit: Can we define all these env vars at the top in a const section? That way we can know what env vars may impact the test run?

test/integration/sonobuoy_integration_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// Details from the git repository that should be available in the Sonobuoy version information
+	gitSHA     string
+	gitVersion string
+)
+
+func findSonobuoyCLI() (string, error) {
+	sonobuoyPath := os.Getenv("SONOBUOY_CLI")
+	if sonobuoyPath == "" {
+		sonobuoyPath = "../../sonobuoy"
```
johnSchnake: I'd const this as well as a default value.

zubron:
Thanks for the review! Yeah, I'm fine to just check that each field has been set, and you're right that it would make this test a lot cleaner/tidier as those Git values aren't needed otherwise. I'll update it!

johnSchnake:
Travis is green but not reporting back to CI for some reason. Merging.

PR #895, opened 2019-09-19
Update images for k8s v1.16

johnSchnake:
With each new minor version of k8s we update the image list
so that commands like `sonobuoy images` pull the right set of images
for testing airgapped envs.

Prepares for v0.16 release

We don't have a great way to test this exactly. The best thing we can do is really just manually check the values against the upstream.

See the values for v1.16 here: https://github.com/kubernetes/kubernetes/blob/v1.16.0/test/utils/image/manifest.go

Note that there are a number of new registries added. Also, in the upstream there is another value `e2eGcRegistry` which was just set to a constant rather than a value from the registry list. This is inappropriate and I've made the intuitive decision to map that value to the corresponding registry list value (which defaults to the same value as the upstream)

### DIFF
--- a/pkg/image/manifest.go
+++ b/pkg/image/manifest.go
@@ -25,12 +25,17 @@ import (
 
 // RegistryList holds public and private image registries
 type RegistryList struct {
-	DockerLibraryRegistry string `yaml:"dockerLibraryRegistry"`
-	E2eRegistry           string `yaml:"e2eRegistry"`
-	EtcdRegistry          string `yaml:"etcdRegistry"`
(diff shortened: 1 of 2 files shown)


johnSchnake:
Force pushed because I forgot to add the new branch on the switch statement to return this value.

zubron:
Sorry, I missed this PR and ended up adding this as part of the release process. Happy to merge this in first though and rebase onto it.

johnSchnake:
No thats OK @zubron , lets just put it into your PR.

PR #896, opened 2019-09-19
Prepare v0.16.0 release

zubron:
Adds the new docs and images for v0.16.0 which corresponds
to the v1.16.0 Kubernetes release.

This release of Kubernetes introduces a number of new registries.
One of the new registry variables (`e2eGcRegistry`) has the same
value as another existing registry variable (`e2eRegistry`). To keep
the code as similar to what is taken from upstream as possible, I
left this in and set it to the same value.


johnSchnake:
So @zubron did you take the code for the images directly from my PR or had you done it yourself separately?

zubron:
I had done it separately. I think the changes were pretty much the same except for a few comments that I had added regarding the registries.

PR #898, opened 2019-09-19
Remove unused dependency from Gopkg.toml

zubron:
This change removes the unused dependency
`k8s.io/apiextensions-apiserver` from Gopkg.toml which was causing a
warning when running `dep ensure`.

This package was no longer needed after #689 and the package was
removed from `vendor/` in #739.


johnSchnake:
👍

PR #903, opened 2019-09-25
Add a small test which runs a basic plugin

zubron:
This change adds a simple test and mostly provides scaffolding to be
able to write further tests using the changes introduced in #907.

It adds a new script `run_integration_tests.sh` which creates a separate
cluster if necessary and then runs the tests on it.

### DIFF
--- a/scripts/run_integration_tests.sh
+++ b/scripts/run_integration_tests.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -e
+
+SCRIPTS_DIR="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
+DIR=$(cd $SCRIPTS_DIR; cd ..; pwd)
+
+cluster="integration"
+testImage="sonobuoy/testimage:v0.1"
+
+if ! kind get clusters | grep -q "^$cluster$"; then
+    kind create cluster --name $cluster
+    # Although the cluster has been created, not all the pods in kube-system are created/available
+    sleep 20
+fi
+
+# Build and load the test plugin image
+make -C $DIR/test/integration/testImage
+kind load docker-image --name $cluster $testImage
+
+# Build and load the sonobuoy image and run integration tests
+make -C $DIR KIND_CLUSTER=$cluster deploy_kind
+KUBECONFIG="$(kind get kubeconfig-path --name="$cluster")" VERBOSE=true make -C $DIR int
--- a/test/integration/sonobuoy_integration_test.go
+++ b/test/integration/sonobuoy_integration_test.go
@@ -4,11 +4,13 @@ package integration
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
 	"strings"
 	"testing"
+	"time"
 )
 
 const (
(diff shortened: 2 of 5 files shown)


test/integration/sonobuoy_integration_test.go
```diff
@@ -32,6 +34,40 @@ func findSonobuoyCLI() (string, error) {
...
 
+// TestRunAndDelete runs a simple plugin to check that it runs successfully and can be deleted
+func TestRunAndDelete(t *testing.T) {
+	var stdout, stderr bytes.Buffer
+
+	ns := randomNamespace()
+	ctx, cancel := context.WithTimeout(context.Background(), 45*time.Second)
+	defer cancel()
+
+	runCommand := exec.CommandContext(ctx, sonobuoy, "run", "--image-pull-policy", "IfNotPresent", "--wait", "-p", "testdata/basic_plugin.yaml", "-n", ns)
```
johnSchnake: FWIW if you don't require special spacing in your values it can be easier to write/read if you just do `strings.Fields("run --image-pull-policy=IfNotPresent ...")`

zubron: I didn't know about that. Thanks! :)

test/integration/testdata/basic_plugin.yaml
```diff
@@ -0,0 +1,13 @@
+sonobuoy-config:
```
johnSchnake: FYI I'm working on the testImage and plugin definitions now. Going to get a few that can run and produce interesting output (job/ds junit/raw passing/failing/etc) and we can expand on it as we want afterwards.

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,22 @@
...
+
+if ! kind get clusters | grep -q "^$cluster$"; then
+    kind create cluster --name $cluster
+    # Although the cluster has been created, not all the pods in kube-system are created/available
+    sleep 20
+fi
+
+# Build and load the test plugin image
+make -C $DIR/test/integration/testImage
+kind load docker-image --name $cluster sonobuoy/testimage:v0.1
```
johnSchnake: nit; I'd specify this at the top of the file too for visibility.

test/integration/sonobuoy_integration_test.go
```diff
@@ -32,6 +34,42 @@ func findSonobuoyCLI() (string, error) {
 	return sonobuoyPath, nil
 }
 
+func cleanup(t *testing.T, namespace string) {
+	var stdout, stderr bytes.Buffer
+	deleteCommand := exec.Command(sonobuoy, "delete", "--wait", "-n", namespace)
```
johnSchnake: If we have various namespaces for different tests (so we wont have collisions) then I'd say we can not do `--wait`, what do you think?

It would speed tests up and avoid issues where a namespace fails to terminate promptly (which isn't uncommon for k8s)

test/integration/sonobuoy_integration_test.go
```diff
@@ -32,6 +34,42 @@ func findSonobuoyCLI() (string, error) {
...
+	deleteCommand.Stdout = &stdout
+	deleteCommand.Stderr = &stderr
+
+	t.Logf("Cleaning up. Running %q\n", deleteCommand.String())
+	if err := deleteCommand.Run(); err != nil {
```
johnSchnake: I think as part of expanding this suite we need to make a handful of helper functions that handle this sort of thing:
 - run, log output if error & error a test
 - run, return output

It feels like you're having to do too much right now each time you want to execute some sonobuoy action.

I'd rather have a test that looks like
```
func TestFoo(...){
  cmd:="run --plugin foo1 --plugin foo2"
  expectInResults := []string{"s1","s2"}
  output := mustExecuteCommandAndGetResults(t, cmd)
  for _,s:=range expectInResults{
    if !strings.Contains(s1, output) {...}
  }
}
```

Once we centralize more of the boilerplate of dealing namespaces, buffers, etc into a few of these helpers we can have tests that look almost like unit tests (e.g. table driven tests to just check for strings/files/etc).

test/integration/utils.go
```diff
@@ -0,0 +1,21 @@
...
+func stringWithCharset(length int, charset string) string {
+	b := make([]byte, length)
+	for i := range b {
+		b[i] = charset[seededRand.Intn(len(charset))]
+	}
+	return string(b)
+}
+
+func randomNamespace() string {
+	charset := "abcdefghijklmnopqrstuvwxyz"
```
johnSchnake: We can do random namespaces, but I know it is a frustration sometimes in tests to not know what test created the ns.

Would it be better to just use `"sonobuoy-"+t.Name()` as a namespace if we needed it?

test/integration/sonobuoy_integration_test.go
```diff
@@ -32,6 +34,42 @@ func findSonobuoyCLI() (string, error) {
...
+	defer cleanup(t, ns)
+
+	args := fmt.Sprintf("run --image-pull-policy IfNotPresent --wait -p testImage/yaml/job-junit-passing-singlefile.yaml -n %v", ns)
+	runCommand := exec.CommandContext(ctx, sonobuoy, strings.Fields(args)...)
+	runCommand.Stdout = &stdout
+	runCommand.Stderr = &stderr
+
+	t.Logf("Running %q\n", runCommand.String())
+	if err := runCommand.Run(); err != nil {
+		t.Errorf("Sonobuoy exited with an error: %v", err)
```
johnSchnake: I guess for our first test, just seeing that it ran and exited 0 is OK. Feels a bit empty but it is just a starting place.

test/integration/sonobuoy_integration_test.go
zubron: You're right, it doesn't need to wait.

test/integration/sonobuoy_integration_test.go
zubron: Yes, I definitely wanted to do that. I had started to do that, then got a bit apprehensive because I was basing the design on one test. I can add some of this now though and refactor later.

test/integration/sonobuoy_integration_test.go
zubron: Sorry, I should have followed up on this PR with a comment. Yes, this is a bit of dull and uninteresting test that doesn't tell us anything more than what we currently have in our CI scripts. I think I just got a bit frustrated with trying to get things working and decided that once it _was_ working, I should just try to get that wrapped up and merged. Admittedly, it was fairly straightforward in the end, I was just frustrated with some silly mistakes I made :upside_down_face:

test/integration/sonobuoy_integration_test.go
johnSchnake: Valid concern; we can follow up with a handful of useful tests and the helpers they need next even in the same PR.

You're right not to want to prematurely make 100 hepler funcs which we dont really need.

PR #904, opened 2019-09-25
Adds the results format as a flag parameter for gen plugin

johnSchnake:
When creating a plugin, one of the most common things you may
want to do is leverage the new result processing capabilities by
specifying if the results should be treated as raw files or as
structured junit (and eventually other formats).

This PR adds the new flag; no testing updates were required
since this does not change default behavior and any tests would
really just test the fact that the flag was added at all, being
trivial.

Related to #668 

**Release note**:
```
Added a new flag `--format` to the `sonobuoy gen plugin` command which allows you to specify if the plugin generates structured junit results. Defaults to `raw` meaning no structure is assumed.
```

### DIFF
--- a/cmd/sonobuoy/app/gen_plugin_def.go
+++ b/cmd/sonobuoy/app/gen_plugin_def.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/heptio/sonobuoy/pkg/client/results"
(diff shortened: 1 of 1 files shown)


johnSchnake:
I am putting this up because our integration suite will reuse a single image in many ways; this was a piece of functionality I needed to generate lots of varying plugin definitions.

PR #907, opened 2019-09-25
Adds a testImage for our integration tests

johnSchnake:
Adds the following:
 - a Dockerfile to build a testImage
 - a Makefile to build the image
 - a script to generate YAML plugin definitions for some test cases
 - the generated plugin definitions from the above script
 - source code to run a test image in two modes: single file or
tarball results
 - raw data which the test image will report to Sonobuoy as its
test results

The test image is meant to be invoked with either single-file
or tar-file mode. The arguments after that are the file[s] which
to report as the plugins results.

This allows us to generate passing/failing/skipped tests and
report on all the combinations with just having some static
results files and various plugin definitions which invoke the
image in different ways.

The path to the files to report, when invoking the plugin,
should be stated as absolute paths, rooted at /resources (analogous
to how they are currently put into the resources directory).

Related to #668

This doesn't fully tie it into CI or anything but the intent was for you to be able to leverage this in your PRs.  This should be a no-op really for the repo if it isn't utilized so I don't see a problem with it merging as you tinker with it.

I was building it via `make build` then loading it manually into my kind cluster via `kind load docker-image sonobuoy/testimage:v0.1`

**Release note**:
```
NONE
```


test/integration/testImage/src/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/heptio/sonobuoy/test/integration/testImage/src
+
+go 1.12
```
zubron: Can this use go 1.13? That's what is being used everywhere else.

test/integration/testImage/src/singleFileResults.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func reportSingleFile(cmd *cobra.Command, args []string) error {
+	targetFile := args[0]
+	resultsFile := filepath.Join(resultsDir, filepath.Base(targetFile))
+
+	// Copy file to location Sonobuoy can get it.
+	_, err := copyFile(targetFile, resultsFile)
+	if err != nil {
+		return errors.Wrap(err, "failed to write to done file")
```
zubron: This message should indicate that it failed to copy the file rather than write the done file.

test/integration/testImage/src/tarResults.go
```diff
@@ -0,0 +1,63 @@
...
+	defer os.RemoveAll(outpath)
+
+	for _, f := range args {
+		targetFile := f
+		resultsFile := filepath.Join(outpath, filepath.Base(targetFile))
+
+		// Copy file to location Sonobuoy can get it.
+		_, err := copyFile(targetFile, resultsFile)
+		if err != nil {
+			return errors.Wrap(err, "failed to write to done file")
```
zubron: Like the file above, it should mention failing to copy the file.

test/integration/testImage/yaml/generate.sh
```diff
@@ -0,0 +1,41 @@
...
+--format="raw" > job-raw-singlefile.yaml
+
+sonobuoy gen plugin \
+--name=ds-junit-passing-tar \
+--image=sonobuoy/testimage:v0.1 \
+--cmd="testImage" \
+--arg="tar-file" \
+--arg="/resources/hello-world.txt" \
+--arg="/resources/junit-multi-suite-single-failure.xml" \
+--format="junit" > ds-junit-passing-tar.yaml
```
zubron: Including `ds` in the name here and below suggest that these should be DaemonSet plugins but they are Job plugins.

test/integration/testImage/src/go.mod
johnSchnake: 👍 I dont know how much it matters; it had to do with the fact that I ran `go mod init` locally with go 1.12 but after updating go and such to get this to be the go 1.13, the go.sum/go.mod doesnt change in any way (just FYI, the requested change was made)

johnSchnake:
I added `testImage` to the gitignore, I wonder why it isn't being respected. I must have made a mistake in how I listed it?

zubron:
Hmm, any chance you staged/added the file before creating the .gitignore file?

johnSchnake:
That could be too

test/integration/testImage/yaml/ds-junit-passing-raw.yaml
```diff
@@ -0,0 +1,18 @@
+sonobuoy-config:
```
zubron: Is this a leftover file? It's not managed by the `generate.sh` script.

johnSchnake: Removed; thanks. Old file that should have gotten cleaned up.

PR #909, opened 2019-09-26
Prints all plugins when invoking `sonobuoy results` by default

### LINKED ISSUE #905: Sonobuoy results command should list all plugins
`sonobuoy results $tarball` should give the same report it currently does for all plugins if `--plugin` is not provided.

Current functionality would be maintained if users just provided `--plugin=e2e` so any breakage to existing scripts is minimal.

The problem this tries to solve is if you are given a tarball it is still impossible to find out _what was run_ without actually opening up the tarball. This is one of the problems we were trying to solve with this generic `results` command.

johnSchnake:
If given a tarball, there is not a way to detect which plugins
were actually run without opening it up first. This means that the
`sonobuoy results` command is less useful since you have to have
preexisting knowledge about the tarball/run. You'd also have to
manually execute the command numerous times if you want all the
summaries of the plugins.

This change:
 -  makes the default value for the `--plugin` flag the
empty string
 - when a string value is given, it will print a single report
like the existing behavior
 - when the default, empty string is used, we will list the report
for each plugin
 - adds a new file, meta/info.json which will contain the list of
plugins that were actually run so that it is easier to grab this
list (rather than more complicated logic using the other data in
the tarball)

Fixes #905

If you target an old tarball with the new CLI you'll get a message about the plugins not being discernable from either the flags or the `meta/info.json`. I could fallback to trying the `e2e` by default at that point but I think it is a bit confusing to twist the default behavior since the error message in some cases may not reflect what was the intended behavior (e.g. they want to list all plugins and then get an error msg about e2e)

If you target a new tarball with an old CLI, no difference in behavior.

If you target a new tarball (e.g. a tarball made by the new image) with the new CLI then you'll get the new behavior.

**Release note**:
```
`sonobuoy results results.tar.gz` now defaults to printing information on all the plugins that were run. You may still print just the e2e plugin information by setting the existing flag `--plugin e2e`.
```

### DIFF
--- a/cmd/sonobuoy/app/results.go
+++ b/cmd/sonobuoy/app/results.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	"github.com/heptio/sonobuoy/pkg/client/results"
+	"github.com/heptio/sonobuoy/pkg/discovery"
 	"github.com/heptio/sonobuoy/pkg/errlog"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -66,7 +67,11 @@ func NewCmdResults() *cobra.Command {
 		},
 		Args: cobra.ExactArgs(1),
 	}
-	AddPluginFlag(&data.plugin, cmd.Flags())
+
+	cmd.Flags().StringVarP(
+		&data.plugin, "plugin", "p", "",
+		"Which plugin to show results for. Defaults to printing them all.",
+	)
 	cmd.Flags().StringVarP(
 		&data.mode, "mode", "m", resultModeReport,
 		`Modifies the format of the output. Valid options are report, detailed, or dump.`,
@@ -100,6 +105,9 @@ func getReader(filepath string) (*results.Reader, func(), error) {
 	return r, func() { gzr.Close(); f.Close() }, nil
 }
 
+// result takes the resultsInput and tries to print the requested infromation from the archive.
+// If there is an error printing any individual plugin, only the last error is printed and all plugins
+// continue to be processed.
 func result(input resultsInput) error {
 	r, cleanup, err := getReader(input.archive)
 	defer cleanup()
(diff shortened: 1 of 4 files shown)


cmd/sonobuoy/app/results.go
```diff
@@ -66,7 +67,11 @@ func NewCmdResults() *cobra.Command {
 		},
 		Args: cobra.ExactArgs(1),
 	}
-	AddPluginFlag(&data.plugin, cmd.Flags())
```
johnSchnake: Also, FYI, this was inappropriate to use here. The values `--plugin -p` were ok, but the description of the flags meaning was specific to the `sonobuoy images` command.

pkg/config/config.go
```diff
@@ -154,7 +154,7 @@ type Config struct {
 	PluginSelections []plugin.Selection       `json:"Plugins" mapstructure:"Plugins"`
 	PluginSearchPath []string                 `json:"PluginSearchPath" mapstructure:"PluginSearchPath"`
 	Namespace        string                   `json:"Namespace" mapstructure:"Namespace"`
-	LoadedPlugins    []plugin.Interface       `json:"-"` // this is assigned when plugins are loaded.
+	LoadedPlugins    []plugin.Interface       `json:"LoadedPlugins" mapstructure:"Namespace"` // this is assigned when plugins are loaded.
```
johnSchnake: Oops; this was just debug to check things. Removing it.

pkg/discovery/discovery.go
```diff
@@ -207,11 +216,20 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
...
+		runInfo.LoadedPlugins = append(runInfo.LoadedPlugins, p.GetName())
 		trackErrorsFor("saving plugin info")(
 			dumpPlugin(p, outpath),
 		)
 	}
 
+	// Dump extra metadata that may be useful to postprocessors or analysis.
+	if blob, err := json.Marshal(runInfo); err == nil {
+		if err = ioutil.WriteFile(path.Join(metapath, "info.json"), blob, 0644); err != nil {
+			errlog.LogError(errors.Wrap(err, "could not write config.json file"))
```
johnSchnake: Fixing up this whole section; wrong log lines (copypasta) and return path.

pkg/client/results/reader.go
```diff
@@ -53,6 +53,10 @@ const (
 	defaultNodesFile          = "Nodes.json"
 	defaultServerVersionFile  = "serverversion.json"
 	defaultServerGroupsFile   = "servergroups.json"
+
+	// InfoFile contains data not that isn't strictly in another location
+	// but still relevent to post-processing or understanding the run in some way.
+	InfoFile = "info.json"
```
zubron: Given that a method is provided to return the info file path, does it need to be exported?

johnSchnake: So you raise a good question which I definitley touched on when coding this: who is in charge of what information and what is the right way to grab it?

A few different packages hardcode these string values (results.json, errors.json, results/, etc), some of the values here are exported, some arent. Some have methods to return their values, some dont.

I'm going to make a new ticket specifically for fixing this up and clarifying where to find all the results structure (folders/names/etc) and how to use them.

johnSchnake: https://github.com/heptio/sonobuoy/issues/912

TIL: there is a quick 'Reference comment in new issue' button which will create a new issue with the comment body.

PR #911, opened 2019-09-27
Allow the --plugin flag to specify a directory of plugins

### LINKED ISSUE #906: Sonobuoy CLI should allow specifying a directory of plugins
We currently allow `sonobuoy run --plugin` but if you have numerous plugins this can get tedious and is more frustrating to script around.

I'd like to be able to specify something like `sonobuoy run --plugin-dir` or `--plugins` which may take a directory.

I think this would just be a shortcut for adding all those plugins one-by-one and wouldn't require much effort.

I don't think this has to be overthought or have complex interaction with the existing `--plugin` flag. I think it is reasonable to allow both to be set and they are all just additive (i.e. it would be valid to say `--plugins lotsOfPluginsInThisDir --plugin alsoAddThisOne.yaml`

I am hitting this because of the integration suite we are putting together and I am currently creating lots of yaml plugin definitions and want to run them all.

johnSchnake:
If someone specifies `sonobuoy run|gen --plugin X` then X ought to
be allowed to be either a single plugin file or a directory of plugins.

To avoid dropping errors intentionally or processing lots of unnecessary
files, when processing a directory, we only consider *.yaml files. We
also do not recurse deeper into the directory.

Fixes #906

This entire diff is almost just test updates. I can split it out so that it is more clear if you want.

I originally wasn't going to have tests for this since there are no tests for flags->GenConfig generation. However I decided that it should be covered and the test was _almost_ doing that anyways.

However, I found a bug when making the test so I had to resolve that too (a one word change). Made comments throughout to help clarify.

**Release note**:
```
The --plugin flag (for specifying custom plugins on the command line) can now accept a directory name. If a directory is specified, all *.yaml files will be processed as plugins and loaded as if they had been specified individually.
```

### DIFF
--- a/cmd/sonobuoy/app/gen.go
+++ b/cmd/sonobuoy/app/gen.go
@@ -104,7 +104,7 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
 
 	// the Enabled and Disabled modes of rbacmode don't need the client, so kubeclient can be nil.
 	// if kubeclient is needed, ErrRBACNoClient will be returned and that error can be reported back up.
-	rbacEnabled, err := genflags.rbacMode.Enabled(kubeclient)
+	rbacEnabled, err := g.rbacMode.Enabled(kubeclient)
 	if err != nil {
 		if errors.Cause(err) == ErrRBACNoClient {
 			return nil, errors.Wrap(err, kubeError.Error())
--- a/cmd/sonobuoy/app/gen_test.go
+++ b/cmd/sonobuoy/app/gen_test.go
@@ -14,6 +14,8 @@ limitations under the License.
 package app
 
 import (
+	"io/ioutil"
+	"os"
 	"reflect"
 	"strings"
 	"testing"
@@ -22,6 +24,10 @@ import (
 	"github.com/heptio/sonobuoy/pkg/client"
 	"github.com/heptio/sonobuoy/pkg/config"
 	"github.com/heptio/sonobuoy/pkg/plugin"
+	"github.com/heptio/sonobuoy/pkg/plugin/manifest"
+
+	"github.com/kylelemons/godebug/pretty"
+	v1 "k8s.io/api/core/v1"
 )
 
 const (
@@ -85,19 +91,30 @@ func TestResolveConformanceImage(t *testing.T) {
 func TestResolveConfig(t *testing.T) {
 	defaultPluginSearchPath := config.New().PluginSearchPath
 	defaultAggr := plugin.AggregationConfig{TimeoutSeconds: 10800}
+	dynamicConfigFileName := "*determinedAtRuntime*"
 
 	tcs := []struct {
-		name     string
-		input    *genFlags
+		name string
+
+		// CLI input to parse.
+		input string
+
+		// Not every field of this will be tested at this time.
 		expected *config.Config
-		cliInput string
+
+		// If specified, will write the config to a temp file then append
+		// the `--config=tmpfile` to the input. This way we can keep the config files
+		// small and with the test rather than in a testdata file.
+		configFileContents string
+
+		// TODO(jschnake): This test previously was just testing the config.Config
+		// and only certain fields. We may consider expanding this to testing the entire
+		// genConfig object but for now I am expanding this just to ensure proper plugin loading.
+		expectedGenConfigPlugins *client.GenConfig
 	}{
 		{
-			name: "NonDisruptiveConformance mode when supplied config is nil (nothing interesting happens)",
-			input: &genFlags{
-				mode:           client.NonDisruptiveConformance,
-				sonobuoyConfig: SonobuoyConfig{},
-			},
+			name:  "NonDisruptiveConformance mode when supplied config is nil (nothing interesting happens)",
+			input: "",
 			expected: &config.Config{
 				Namespace:       "sonobuoy",
 				WorkerImage:     "gcr.io/heptio-images/sonobuoy:" + buildinfo.Version,
@@ -110,19 +127,13 @@ func TestResolveConfig(t *testing.T) {
 				Aggregation:      defaultAggr,
 				Resources:        config.DefaultResources,
 			},
-		}, {
-			name: "Quick mode and a non-nil supplied config",
-			input: &genFlags{
-				mode: client.Quick,
-				sonobuoyConfig: SonobuoyConfig{
-					Config: config.Config{
-						Aggregation: plugin.AggregationConfig{
-							BindAddress: "10.0.0.1",
-						},
-					},
-					raw: rawInput,
-				},
+			expectedGenConfigPlugins: &client.GenConfig{
+				DynamicPlugins: []string{"e2e", "systemd-logs"},
 			},
+		}, {
+			name:               "Quick mode and a non-nil supplied config",
+			input:              "--mode quick --config=" + dynamicConfigFileName,
+			configFileContents: `{"Server":{"bindaddress":"10.0.0.1"}}`,
 			expected: &config.Config{
 				Namespace:       "sonobuoy",
 				WorkerImage:     "gcr.io/heptio-images/sonobuoy:" + buildinfo.Version,
@@ -138,19 +149,13 @@ func TestResolveConfig(t *testing.T) {
 				PluginSearchPath: defaultPluginSearchPath,
 				Resources:        config.DefaultResources,
 			},
-		}, {
-			name: "NonDisruptiveConformance mode with plugin selection specified",
-			input: &genFlags{
-				mode: client.NonDisruptiveConformance,
-				sonobuoyConfig: SonobuoyConfig{
-					Config: config.Config{
-						PluginSelections: []plugin.Selection{
-							{Name: "systemd-logs"},
-						},
-					},
-					raw: "not empty",
-				},
+			expectedGenConfigPlugins: &client.GenConfig{
+				DynamicPlugins: []string{"e2e"},
 			},
+		}, {
+			name:               "NonDisruptiveConformance mode with plugin selection specified",
+			input:              "--plugin systemd-logs --config=" + dynamicConfigFileName,
+			configFileContents: `{"Plugins":[{"name":"systemd-logs"}]}`,
 			expected: &config.Config{
 				Namespace:       "sonobuoy",
 				WorkerImage:     "gcr.io/heptio-images/sonobuoy:" + buildinfo.Version,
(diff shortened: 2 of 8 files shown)


cmd/sonobuoy/app/gen.go
```diff
@@ -104,7 +104,7 @@ func (g *genFlags) Config() (*client.GenConfig, error) {
 
 	// the Enabled and Disabled modes of rbacmode don't need the client, so kubeclient can be nil.
 	// if kubeclient is needed, ErrRBACNoClient will be returned and that error can be reported back up.
-	rbacEnabled, err := genflags.rbacMode.Enabled(kubeclient)
+	rbacEnabled, err := g.rbacMode.Enabled(kubeclient)
```
johnSchnake: Tiny, mean bug that I found while coding these changes. It is referencing a global variable instead of the value it it was invoked on. This is why it happens to work for actual runs of Sonobuoy (since the global value gets used/set) but not in the test.

Very frustrating to find and another reason why we shouldn't be relying on them. One day we'll get around to fixing all the globals for these cmds...one day...

cmd/sonobuoy/app/gen_test.go
```diff
@@ -85,19 +91,31 @@ func TestResolveConformanceImage(t *testing.T) {
 func TestResolveConfig(t *testing.T) {
 	defaultPluginSearchPath := config.New().PluginSearchPath
 	defaultAggr := plugin.AggregationConfig{TimeoutSeconds: 10800}
+	dynamicConfigFileName := "*determinedAtRuntime*"
+	defaultConfig := `{"Server":{"bindaddress":"10.0.0.1"}}`
```
johnSchnake: Doesn't need a default value; I'll just use this where appropriate.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -85,19 +91,31 @@ func TestResolveConformanceImage(t *testing.T) {
 func TestResolveConfig(t *testing.T) {
 	defaultPluginSearchPath := config.New().PluginSearchPath
 	defaultAggr := plugin.AggregationConfig{TimeoutSeconds: 10800}
+	dynamicConfigFileName := "*determinedAtRuntime*"
```
johnSchnake: I use this to hot-swap out the name since I write the config to a temp file but want the cmdline input to be specified explicitly.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -85,19 +91,31 @@ func TestResolveConformanceImage(t *testing.T) {
...
+		configFileContents string
+
+		// TODO(jschnake): This test previously was just testing the config.Config
+		// and only certain fields. We may consider expanding this to testing the entire
+		// genConfig object but for now I am expanding this just to ensure proper plugin loading.
+		expectedGenConfigPlugins *client.GenConfig
 	}{
 		{
-			name: "NonDisruptiveConformance mode when supplied config is nil (nothing interesting happens)",
-			input: &genFlags{
```
johnSchnake: The test used to be a bit confusing since it could be given a genFlags object or it could be given cmdline input.

Now the test is a bit more consistent by always parsing cmdline input and relying on actual config files.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -244,44 +257,105 @@ func TestResolveConfig(t *testing.T) {
...
+				WorkerImage:      "gcr.io/heptio-images/sonobuoy:" + buildinfo.Version,
+				ImagePullPolicy:  "IfNotPresent",
+				PluginSelections: nil,
+				PluginSearchPath: defaultPluginSearchPath,
+				Aggregation:      defaultAggr,
+				Resources:        config.DefaultResources,
+			},
+			expectedGenConfigPlugins: &client.GenConfig{
+				StaticPlugins: []*manifest.Manifest{
+					{SonobuoyConfig: manifest.SonobuoyConfig{Driver: "Job", PluginName: "plugin1", SkipCleanup: false, ResultFormat: "raw"}, Spec: manifest.Container{Container: v1.Container{Name: "plugin", Image: "foo/bar:v1", Command: []string{"./run.sh"}, VolumeMounts: []v1.VolumeMount{{ReadOnly: false, Name: "results", MountPath: plugin.ResultsDir}}}}},
```
johnSchnake: These static plugin lines are long but shouldn't really change. I just had to grab one and then a bit of copy/paste. Details arent too important here, the point is that the right set of plugins got loaded (really just need to look at name/image)

cmd/sonobuoy/app/gen_test.go
```diff
@@ -244,44 +257,105 @@ func TestResolveConfig(t *testing.T) {
...
+				}
+				defer os.Remove(tmpFile.Name())
+				err = ioutil.WriteFile(tmpFile.Name(), []byte(tc.configFileContents), 0666)
+				if err != nil {
+					t.Fatalf("Failed to write temp config file for test: %v", err)
 				}
+				tc.input = strings.Replace(tc.input, dynamicConfigFileName, tmpFile.Name(), -1)
 			}
 
-			conf := tc.input.resolveConfig()
```
johnSchnake: The old test was using `resolveConfig` which generates the `config.Config` object. However, this isn't sufficient to see if the GenConfig would be appropriate so we were missing the ability to test large sections of input processing.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -244,44 +257,105 @@ func TestResolveConfig(t *testing.T) {
...
 
-			if conf.Namespace != tc.expected.Namespace {
-				t.Errorf("Expected namespace %v but got %v", tc.expected.Namespace, conf.Namespace)
+			if genConfig.Config.ImagePullPolicy != tc.expected.ImagePullPolicy {
+				t.Errorf("Expected image pull policy %v but got %v", tc.expected.ImagePullPolicy, genConfig.Config.ImagePullPolicy)
 			}
 
-			if conf.WorkerImage != tc.expected.WorkerImage {
-				t.Errorf("Expected worker image %v but got %v", tc.expected.WorkerImage, conf.WorkerImage)
+			if genConfig.Config.Aggregation.TimeoutSeconds != tc.expected.Aggregation.TimeoutSeconds {
```
johnSchnake: Now that we are generating the 'higher level' genConfig object, these checks just needed tweaked to look at the same value.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -244,44 +257,105 @@ func TestResolveConfig(t *testing.T) {
...
 
-			if conf.Aggregation.TimeoutSeconds != tc.expected.Aggregation.TimeoutSeconds {
-				t.Errorf("Expected timeout %v but got %v", tc.expected.Aggregation.TimeoutSeconds, conf.Aggregation.TimeoutSeconds)
+			if !reflect.DeepEqual(genConfig.Config.Resources, tc.expected.Resources) {
+				t.Errorf("expected resources %v but got %v", tc.expected.Resources, genConfig.Config.Resources)
 			}
 
-			if !reflect.DeepEqual(conf.PluginSelections, tc.expected.PluginSelections) {
-				t.Errorf("expected PluginSelections %v but got %v", tc.expected.PluginSelections, conf.PluginSelections)
+			if diff := pretty.Compare(genConfig.StaticPlugins, tc.expectedGenConfigPlugins.StaticPlugins); diff != "" {
```
johnSchnake: Added 2 checks here just for the static/dynamic plugins. Consistent with the existing test which is just checking for key bits of data.

If we wanted to change these tests to check the entire genconfig we should probably just serialize the values and use golden files since the objects are so big.

cmd/sonobuoy/app/pluginList.go
```diff
@@ -71,19 +74,53 @@ func (p *pluginList) Set(str string) error {
 	case pluginSystemdLogs:
 		p.DynamicPlugins = append(p.DynamicPlugins, str)
 	default:
-		b, err := ioutil.ReadFile(str)
+		finfo, err := os.Stat(str)
```
johnSchnake: Since I have to stat to check if it is a directory or not, I had to have a new possible error message and so some tests needed updated for this reason as well. Tiny change but impacts a handful of tests.

cmd/sonobuoy/app/pluginList.go
```diff
@@ -71,19 +74,53 @@ func (p *pluginList) Set(str string) error {
...
+				return errors.Wrapf(err, "failed to load plugin in file %q", file.Name())
+			}
+		}
+	}
+
+	return nil
+}
+
+// loadSinglePlugin loads a single plugin located at the given path.
+func (p *pluginList) loadSinglePlugin(filepath string) error {
```
johnSchnake: This entire section is just moving code around. The existing `load plugin` logic was moved here and then the `load dir of plugins` just gets a list of files and leverages this same method.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -186,10 +190,13 @@ func TestResolveConfig(t *testing.T) {
...
+			expectedGenConfigPlugins: &client.GenConfig{
+				DynamicPlugins: []string{"e2e", "systemd-logs"},
+			},
 		}, {
-			name:     "Flags shouldn't override the config settings unless set",
-			input:    &genFlags{},
-			cliInput: "--sonobuoy-image=flagImage --config testdata/sonobuoy.conf",
+			name:               "Flags shouldn't override the config settings unless set",
+			input:              "--sonobuoy-image=flagImage --config " + dynamicConfigFileName,
+			configFileContents: `{"Namespace":"configNS","WorkerImage":"configImage","ImagePullPolicy":"Never","Server":{"timeoutseconds":500}}`,
```
johnSchnake: This `sonobuoy.conf` was used just for a few of these tests and since I already added the logic to make/use a temp one I just deleted the file and copied the contents to a few tests.

Much more flexible and now it is clear what the conf file is for each test instead of having to go look it up.

johnSchnake:
Test failing in CI but passing locally due to me having my KUBECONFIG set and pointed to a real cluster. Will update the tests.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -202,10 +209,13 @@ func TestResolveConfig(t *testing.T) {
...
 				Resources:        config.DefaultResources,
 			},
+			expectedGenConfigPlugins: &client.GenConfig{
+				DynamicPlugins: []string{"e2e", "systemd-logs"},
+			},
 		}, {
-			name:     "Flags shouldn't override the config settings unless set",
-			input:    &genFlags{},
-			cliInput: "--config testdata/sonobuoy.conf",
+			name:               "Flags shouldn't override the config settings unless set",
```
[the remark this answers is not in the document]
johnSchnake: Its not clear to me what one of those is supposed to test; the name indicates resolving conflicts with the flag/config but doesnt actually give any flags other than the config. I'll have it check against some other field so it does something reasonable before merging (and update the name)

PR #917, opened 2019-09-30
Tweak readme with info about results.

johnSchnake:
A recent issue made me take a look at our docs and I noticed that
we were still referencing the `e2e` command instead of the
`results` command. It was actually hard to find a link to that page
which should be more visible.

N/A

N/A

**Release note**:
```
NONE
```


johnSchnake:
Small docs change; merging without review/CI finishing. Docs look good.

PR #923, opened 2019-09-30
Fix blockquote rendering

johnSchnake:
Will not be linked to the docs site but will be accessible by URL.
Just useful for devs who may have questions about format.

### DIFF
--- a/site/README-JEKYLL.md
+++ b/site/README-JEKYLL.md
@@ -20,7 +20,7 @@ If you are running a build on Ubuntu you will need the following packages:
 1. Install Jekyll and plug-ins in one fell swoop. `gem install github-pages`
 This mirrors the plug-ins used by GitHub Pages on your local machine including Jekyll, Sass, etc.
 2. Clone down your own fork, or clone the main repo `git clone https://github.com/vmware-tanzu/sonobuoy` and add your own remote.
-3. `cd velero/site`
+3. `cd sonobuoy/site`
 4. `rbenv local 2.6.3`
 5. `bundle install`
 6. Serve the site and watch for markup/sass changes `jekyll serve --livereload`. You may need to run `bundle exec jekyll serve --livereload`.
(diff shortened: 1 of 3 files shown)


johnSchnake:
From the page https://deploy-preview-923--sonobuoy.netlify.com/docs/master/markdown-cheatsheet/

We can see problems with:
 -  underlined header styles
 - tables
 - syntax highlighting in code blocks
 - blockquotes
 - some issue (not sure how much) with inline html

johnSchnake:
Trying to change the `markdown` processor from redcarpet to kramdown did not fix this. Both syntax seem to support the same blockquote syntax (`> text`) but neither is working so I'm assuming it is something else higher in the stack.

That led me to look at the source html and there is actually a `<blockquote>` tag, so the css must be involved in removing the look we want. Taking a look...

johnSchnake:
I can see that it doesn't seem to have the border-left that I expect but adding it doesn't seem to have the effect I intend. I also tried to toggle lots of css off to see if it would look as expected, but I can't seem to track it down.

johnSchnake:
See https://deploy-preview-923--sonobuoy.netlify.com/docs/master/markdown-cheatsheet/#blockquotes for the effect.

It was just had some css for the blockquotes that didnt have the borders set

johnSchnake:
If you are opposed to the file let me know; I can remove it. Having it go forward would just make it easier to check our rendering/css. It came from https://github.com/adam-p/markdown-here; I'm just copying a file of sample markdown so I'm not under the impression that I have to cite the MIT license due to the size/value of the data. If you disagree I can add another custom one that just spits out that sort of stuff.

johnSchnake:
Done tweaking and happy with the look now (on the sample page and throughout other docs). Feel free to review now.

johnSchnake:
Yea it seems bootstrap just doesn't do much with the blockquote: https://getbootstrap.com/docs/4.0/content/typography/#blockquotes

I didnt even realize when I jumped to those files they were vendored at first. That isn't a great approach since the changes are easily lost.

Maybe I should just adjust the PR to make all those `note` blockquotes italic'd or something.

johnSchnake:
Ended up removing the cheatsheet; since we have the netlify preview it shouldn't be a big deal really. Having the cheatsheet wouldnt have helped unless we know what to look for in the first place.

Having the custom css doc will make it going forward if we want to tweak things.

PR #927, opened 2019-10-01
Update our registry to dockerhub and use our latest kind-node

### LINKED ISSUE #913: Update cluster verison we test with kind to v1.16
Images just arent available yet: https://hub.docker.com/r/kindest/node/tags?page=1&name=v1.16

However, once they get posted then we will also be able to leverage the go-based runner.

This will:
 - give us better logging in our own test runs
 - avoid the PID issue lots of our CI runs have (this is also fixed in the bash runner though)
 - gives the go runner some testing so we can see if it has issues
 - allow us to experiment with other configurations more easily if need-be

johnSchnake:
New Dockerhub org for sonobuoy so we will start moving our
image publishing there. As of now we do not have our main images
published there (another PR will do that) but I've manually
built/pushed kind-node:v1.16.0 and so we can use it in our CI.

Fixes #913

**Release note**:
```
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ BINARY = sonobuoy
 TARGET = sonobuoy
 GOTARGET = github.com/vmware-tanzu/$(TARGET)
 GOPATH = $(shell go env GOPATH)
-REGISTRY ?= gcr.io/heptio-images
+REGISTRY ?= sonobuoy
 IMAGE = $(REGISTRY)/$(TARGET)
 DIR := ${CURDIR}
 DOCKER ?= docker
@@ -72,6 +72,10 @@ LINT = golint $(GOLINT_FLAGS) $(TEST_PKGS)
 DOCKER_FLAGS =
 DOCKER_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(BUILDMNT) $(DOCKER_FLAGS) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c
 GO_BUILD ?= CGO_ENABLED=0 $(GO_SYSTEM_FLAGS) go build -o $(BINARY) $(VERBOSE_FLAG) -ldflags="-s -w -X $(GOTARGET)/pkg/buildinfo.Version=$(GIT_VERSION) -X $(GOTARGET)/pkg/buildinfo.GitSHA=$(GIT_REF_LONG)" $(GOTARGET)
+
+# Kind images
+K8S_PATH ?= $(GOPATH)/src/github.com/kubernetes/kubernetes
+KIND_K8S_TAG ?= $(shell cd $(K8S_PATH) && git describe)
 
 .PHONY: all container push clean test local-test local generate plugins int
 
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -73,6 +73,10 @@ DOCKER_FLAGS =
 DOCKER_BUILD ?= $(DOCKER) run --rm -v $(DIR):$(BUILDMNT) $(DOCKER_FLAGS) -w $(BUILDMNT) $(BUILD_IMAGE) /bin/sh -c
 GO_BUILD ?= CGO_ENABLED=0 $(GO_SYSTEM_FLAGS) go build -o $(BINARY) $(VERBOSE_FLAG) -ldflags="-s -w -X $(GOTARGET)/pkg/buildinfo.Version=$(GIT_VERSION) -X $(GOTARGET)/pkg/buildinfo.GitSHA=$(GIT_REF_LONG)" $(GOTARGET)
 
+# Kind images
+K8S_PATH ?= $(shell cd ~/go/src/github.com/kubernetes/kubernetes && pwd)
```
zubron: Perhaps this could use `GOPATH` (it's defined above) for referencing the Kubernetes path?

Makefile
```diff
@@ -188,5 +192,24 @@ clean:
...
+
+# check-kind-env will show you what will be built/tagged before doing so with kind_images
+check-kind-env:
+ifndef K8S_PATH
+	$(error K8S_PATH is undefined)
+endif
+ifndef KIND_K8S_TAG
+	$(error KIND_K8S_TAG is undefined)
+endif
+	echo --kube-root=$(K8S_PATH) tagging as --image $(REGISTRY)/kind-node:$(KIND_K8S_TAG)
```
zubron: Was this a debugging line? I think the output from make is verbose and will show the `kind build node-image...` command details. If you wanted to include it, it might be better as part of the `kind_images` target as that is where the building/tagging takes place.

johnSchnake: Well it was partially a debug line and partially a helper since I thought it might make sense in our steps to do something like `make check-kind-env` and verify the path and versions are what we want. The output from the build starting isn't actually that great at the start; once things are tagged all is clear, but its not that optimal to either waste that time or even start/stop that mid build IMO.

johnSchnake:
Kind kept failing in CI but I couldn't repro locally. I just updated the version of kind we download to see if that fixes it (I am on v0.6.0, manually built from github) and CI was on v0.4.0. I updated to v0.5.1, which is the latest actual release.

johnSchnake:
Hm so the issue right now is that I updated the tagging in the scripts for the main sonobuoy image so that gets loaded into the cluster, but then the codebase itself is looking for the `gcr.io` image.

I'll have to simultaneously update the code that specifies the `sonobuoy` image to use for the aggregator/worker.

johnSchnake:
The only consequence of this is that if you run from master, you'll have to specify `sonobuoy-image` until we get a release pushed there. I think thats ok, especially if we copy over the latest release to dockerhub. We can do that manually at some point. Just need to ensure the manifest is accurate.

johnSchnake:
Asking for re-review since I had to update some sonobuoy defaults and tests since last approved.

PR #930, opened 2019-10-01
Use created namespace name during `run --wait`

### LINKED ISSUE #929: Segfault during `run --wait` when reading manifest from stdin/file
A segfault occurs when the output from `sonobuoy gen` is passed via file or stdin to `sonobuoy run`.

```
./sonobuoy gen --mode quick | ./sonobuoy run --wait -f -                                                                                                                                                                              0|1 ↵  7513  15:19:59
INFO[0000] created object                                name=sonobuoy namespace= resource=namespaces
INFO[0000] created object                                name=sonobuoy-serviceaccount namespace=sonobuoy resource=serviceaccounts
INFO[0000] created object                                name=sonobuoy-serviceaccount-sonobuoy namespace= resource=clusterrolebindings
INFO[0000] created object                                name=sonobuoy-serviceaccount namespace= resource=clusterroles
INFO[0000] created object                                name=sonobuoy-config-cm namespace=sonobuoy resource=configmaps
INFO[0000] created object                                name=sonobuoy-plugins-cm namespace=sonobuoy resource=configmaps
INFO[0000] created object                                name=sonobuoy namespace=sonobuoy resource=pods
INFO[0000] created object                                name=sonobuoy-master namespace=sonobuoy resource=services
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x180 pc=0x10a7e5e]

goroutine 1 [running]:
github.com/vmware-tanzu/sonobuoy/pkg/client.(*SonobuoyClient).RunManifest.func1(0xc0007e9a18, 0xc0007e99c8, 0x2)
	/go/src/github.com/vmware-tanzu/sonobuoy/pkg/client/run.go:104 +0x3e
github.com/vmware-tanzu/sonobuoy/vendor/k8s.io/apimachinery/pkg/util/wait.WaitFor(0xc0003dca40, 0xc0007e9c48, 0xc000628420, 0x0, 0x0)
	/go/src/github.com/vmware-tanzu/sonobuoy/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:405 +0x155
github.com/vmware-tanzu/sonobuoy/vendor/k8s.io/apimachinery/pkg/util/wait.pollInternal(0xc0003dca40, 0xc0007e9c48, 0x0, 0x0)
	/go/src/github.com/vmware-tanzu/sonobuoy/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:293 +0x9f
github.com/vmware-tanzu/sonobuoy/vendor/k8s.io/apimachinery/pkg/util/wait.Poll(0x4a817c800, 0x4e94914f0000, 0xc0007e9c48, 0x7, 0xffffffffffffffff)
	/go/src/github.com/vmware-tanzu/sonobuoy/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:287 +0x4d
github.com/vmware-tanzu/sonobuoy/pkg/client.(*SonobuoyClient).RunManifest(0xc000029e00, 0xc000456180, 0xc000314000, 0x11c2, 0x1e00, 0x0, 0x0)
	/go/src/github.com/vmware-tanzu/sonobuoy/pkg/client/run.go:130 +0x791
github.com/vmware-tanzu/sonobuoy/pkg/client.(*SonobuoyClient).Run(0xc000029e00, 0xc000456180, 0x23433e0, 0x0)
	/go/src/github.com/vmware-tanzu/sonobuoy/pkg/client/run.go:163 +0x99
github.com/vmware-tanzu/sonobuoy/cmd/sonobuoy/app.submitSonobuoyRun(0xc000139b80, 0xc00002e900, 0x0, 0x3)
	/go/src/github.com/vmware-tanzu/sonobuoy/cmd/sonobuoy/app/run.go:109 +0x1d5
github.com/vmware-tanzu/sonobuoy/vendor/github.com/spf13/cobra.(*Command).execute(0xc000139b80, 0xc00002e870, 0x3, 0x3, 0xc000139b80, 0xc00002e870)
	/go/src/github.com/vmware-tanzu/sonobuoy/vendor/github.com/spf13/cobra/command.go:830 +0x2aa
github.com/vmware-tanzu/sonobuoy/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc0000de000, 0x0, 0x1364c80, 0xc0000ae058)
	/go/src/github.com/vmware-tanzu/sonobuoy/vendor/github.com/spf13/cobra/command.go:914 +0x2fb
github.com/vmware-tanzu/sonobuoy/vendor/github.com/spf13/cobra.(*Command).Execute(...)
	/go/src/github.com/vmware-tanzu/sonobuoy/vendor/github.com/spf13/cobra/command.go:864
main.main()
	/go/src/github.com/vmware-tanzu/sonobuoy/main.go:26 +0x27

```

This is happening because the wait logic tries to get the status of the aggregator pod using the [`config.Config.Namespace`](https://github.com/vmware-tanzu/sonobuoy/blob/master/pkg/client/run.go#L104). This config is only set in the case where we aren't using a file: https://github.com/vmware-tanzu/sonobuoy/blob/master/cmd/sonobuoy/app/run.go#L68.

For the run to complete successfully with no segfault.

**Environment:**
- Sonobuoy version: Sonobuoy Version: v0.16.1-2-g3290552

zubron:
When using a manifest from a file and the `--wait` flag with `run`, a
segfault would happen because the value is retrieved from the
`config.Config` which is part of the `GenConfig` within the `RunConfig`.
When using a file, this config is nil, so an attempt to retrieve the
namespace value resulted in a segfault. Instead, we can capture the
namespace from the objects within the manifest file when they are being
created by the Kubernetes client.

- Fixes #929

I've written added an integration level test to check this and confirmed that the segfault exists before the change and is fixed by it.

**Release note**:
```
Fixed a bug where a segfault would occur with `sonobuoy run --wait` when providing a manifest with a file.
```

### DIFF
--- a/pkg/client/run.go
+++ b/pkg/client/run.go
@@ -57,6 +57,7 @@ func (c *SonobuoyClient) RunManifest(cfg *RunConfig, manifest []byte) error {
 	buf := bytes.NewBuffer(manifest)
 	d := yaml.NewYAMLOrJSONDecoder(buf, bufferSize)
 
+	var createdNamespace string
 	for {
 		ext := runtime.RawExtension{}
 		if err := d.Decode(&ext); err != nil {
@@ -84,6 +85,14 @@ func (c *SonobuoyClient) RunManifest(cfg *RunConfig, manifest []byte) error {
 		if err != nil {
 			return errors.Wrap(err, "could not get object namespace")
 		}
+
+		// The namespace value is required if Wait is enabled in order to check the status.
+		// It may not be available within the RunConfig but we can retrieve it from the
+		// namespace of objects within the manifest.
+		if createdNamespace == "" && namespace != "" {
+			createdNamespace = namespace
+		}
+
 		// err is used to determine output for user; but first extract resource
 		_, err = c.dynamicClient.CreateObject(obj)
 		resource, err2 := c.dynamicClient.ResourceVersion(obj)
@@ -101,7 +110,7 @@ func (c *SonobuoyClient) RunManifest(cfg *RunConfig, manifest []byte) error {
 		seenStatus := false
 		runCondition := func() (bool, error) {
 			// Get the Aggregator pod and check if its status is completed or terminated.
-			status, err := c.GetStatus(&StatusConfig{cfg.Config.Namespace})
+			status, err := c.GetStatus(&StatusConfig{Namespace: createdNamespace})
 			switch {
 			case err != nil && seenStatus:
 				return false, errors.Wrap(err, "failed to get status")
--- a/Makefile
+++ b/Makefile
@@ -90,6 +90,7 @@ stress: sonobuoy
 
 # Integration tests
 int: DOCKER_FLAGS=-v $(KUBECONFIG):/root/.kube/kubeconfig --env KUBECONFIG=/root/.kube/kubeconfig --network host
+int: TESTARGS= $(VERBOSE_FLAG) -timeout 3m
 int: sonobuoy
 	$(DOCKER_BUILD) 'CGO_ENABLED=0 $(INT_TEST)'
 
--- a/test/integration/sonobuoy_integration_test.go
+++ b/test/integration/sonobuoy_integration_test.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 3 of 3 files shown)


pkg/client/run.go
```diff
@@ -84,6 +85,14 @@ func (c *SonobuoyClient) RunManifest(cfg *RunConfig, manifest []byte) error {
 		if err != nil {
 			return errors.Wrap(err, "could not get object namespace")
 		}
+
+		// The namespace value is required if Wait is enabled in order to check the status.
+		// It may not be available within the RunConfig but we can retrieve it from the
+		// namespace of objects within the manifest.
+		if createdNamespace == "" && namespace != "" {
+			createdNamespace = namespace
```
johnSchnake: Why not just use `namespace`? I'm not clear on what `createdNamespace` adds since by default it is the empty string, and it is set to `namespace` if it isn't the empty string. So won't it just always match `namespace`?

zubron: This loop is iterating over all the resources in the yaml output. Some of those are cluster-scoped (e.g. [ClusterRole](https://github.com/vmware-tanzu/sonobuoy/blob/master/pkg/templates/manifest.go#L51-#L71) which means that they do not have an associated namespace. If the yaml was edited to have these resources be the last ones to be iterated over, namespace would be empty at the end of the loop. You can see this in the logs from Sonobuoy run, some of the resources it logs don't have the namespace field set in the log. Storing the first non empty namespace was an attempt to combat that. This approach isn't fool proof, but seemed good enough for now as it avoids the need to fully parse the yaml and find the Namespace resource, or the contents of the config map (we do this in gen_test).

Issue #914, opened 2019-09-28
Push images to Dockerhub via CircleCI


johnSchnake:
**Describe the solution you'd like**
We'd like the analytics in dockerhub; we want to move way from heptio-images gcr bucket anyways and now would be a good time to kill 2 birds with one stone. In addition we'd may consider changing our CI at the same time. Considering circleCI. Doing both at the same time would make it one central effort and get rid of some of the rework.

**Anything else you would like to add:**
Will probably do one more push to gcr.io and then have a few weeks to transition.

johnSchnake:
Going to split this into 2 tickets; this will remain for dockerhub, but another ticket will be made for circleci.

Once CircleCI has been configured to run appropriately, we can work on getting images pushed to dockerhub via CircleCI.

johnSchnake:
Dockerhub images now being pushed via travis successfully. Closing this ticket as I think the conversion to CircleCI will be pretty trivial on this one issue and is covered by #926

PR #932, opened 2019-10-02
Get dockerhub image pushing to work on Travis

johnSchnake:
I know we're moving away from Travis but I'm working out the
details of getting our scripts to push appropriately. Moving
this to CircleCI probably wont add much overhead.

Related to #914

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ services:
 - docker
 install: true
 script:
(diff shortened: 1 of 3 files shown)


johnSchnake:
Going to merge and just see how it goes; may need to follow up but to get the pushing logic to run we need to get the change onto master.

Issue #883, opened 2019-09-16
Sonobuoy timeout needs to be more clear


johnSchnake:
**What steps did you take and what happened:**
A test run which takes a long time and times out; e.g. `sonobuoy run --e2e-focus=sig-storage` will timeout after 3h but it is hard to diagnose the problem.

For instance, the # of tests to run, that have run, etc by ginkgo don't add up as expected. In addition, the timeout lines are buried in the middle of Sonobuoy/e2e logs because there are numerous cleanup/followup actions that take place.

**What did you expect to happen:**
If there is a Sonobuoy timeout we need to make sure that it is:
 - highly visible; preferably the last thing logged as it exits
 - the status reflects the timeout if you run `sonobuoy status`
 - the log lines for the timeout are errors (looks like they are info lines now)

**Anything else you would like to add:**
xref #880

johnSchnake:
After investigating this there are also some inconsistencies with how we handle these timeouts that I want to remedy. As a bonus, it simplifies some of the logic.

To fixup the timeout behavior I want to roll back part of https://github.com/heptio/sonobuoy/commit/4e4513dbdc84e742b1274ab0fde9f21eb317d77b which added the graceful shutdown of plugins.

The issue it was trying to solve was related to ensuring that _when_ things get stalled/wedged we want to be able to get diagnostic information.

The problem now is that if you look at the timeout flow, the graceful shutdown causes the pods to get cleaned up (i.e. deleted) before the typical cluster queries take place. This means that we lose out on the json data of the pods as well as the podlogs themselves.

In addition, it is unintuitive to any caller of `Run` that the pods in questino will get deleted if and only if there is a timeout.

The graceful shutdown also is handled by the conformance test image in such a way that it will stop running the tests and report back what it already has done. This has some consequences:
 - you do end up getting the `e2e.log` file as part of the results from the plugin (a good thing)
 - you also get the `junit.xml` file with actual junit results in it which _seems_ like a good thing until you realize the the junit doesn't properly report this timeout behavior. So you get in a situation where the logs say (for instance) 276 tests will run but at the end of the logs only 100 ran with no failures. Then the junit just says 100 tests ran and passed. This false positive is very confusing to users and the fact that it was a timeout is buried in the logs and will surely be missed by automation.


IMO the appropriate solution here is:
 - do away with the graceful shutdown. If you miss the deadline, you miss the deadline.
 - since we dont have graceful shutdown
[...]

PR #936, opened 2019-10-03
Results aggregation should consider the errors directory

### LINKED ISSUE #897: Result aggregation doesn't take into account the /errors directory
If you do something like
```
sonobuoy run --timeout=10 --plugin do-nothing-plugin --wait
sonobuoy status --json
```

You'll see that the plugin gets reported as unknown because it didnt have any results. The issue is that the result post-processing looks in the results directory for actual results but doesn't look into the `errors` directory to process the Sonobuoy reported issues like timeouts or failure to start a plugin.

If Sonobuoy is reporting an error and saving it into the tarball, our own post-processing should pick up on that and mark the plugin status as `failed` or `errored` with the details copied from the `errors.json` file.

Currently, the e2e plugin, when timing out, is told to stop and so it still reports results (even if they are partial results). This is changing as part of #883 though.

johnSchnake:
Whenever the aggregator itself deems that a plugin has failed,
(e.g. pods didnt start, timeouts, etc) then it writes its results
into an errors directory. This should be considered when doing
results aggregation so that they can be resulted as failures.

Otherwise, the plugin is just reported as having 0 results and
having an unknown status even though Sonobuoy may have been the one
to mark it as failure internally.

Related to #883
Fixes #897

**Release note**:
```
When inspecting plugin results (for reporting via `sonobuoy status` and `sonobuoy results`) errors saved in the `<plugin>/errors` directory which are reported by Sonobuoy itself are now considered. This should help make certain failure modes such as timeout more clear.
```

### DIFF
--- a/pkg/client/results/processing.go
+++ b/pkg/client/results/processing.go
@@ -17,6 +17,7 @@ limitations under the License.
 package results
 
 import (
+	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path"
@@ -26,6 +27,7 @@ import (
 	"github.com/vmware-tanzu/sonobuoy/pkg/plugin"
 	"github.com/vmware-tanzu/sonobuoy/pkg/plugin/driver/daemonset"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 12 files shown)


pkg/client/results/processing.go
```diff
@@ -146,48 +148,120 @@ func PostProcessPlugin(p plugin.Interface, dir string) (Item, error) {
 	return i, err
 }
 
+// processNodesWithProcessor is called to invoke processDir on each node-specific directory contained
+// underneath the given dir. The directory is assumed to be either the results directory or errors directory
+// which should have the nodes as subdirectories. It returns an item for each node processed and an error
+// only if it couldn't open the original directory. Any errors while processing a specific node are logged
+// but not returned.
+func processNodesWithProcessor(p plugin.Interface, baseDir, dir string, processor postProcessor, selector fileSelector) ([]Item, error) {
```
johnSchnake: This method is just code moved code, not new code. Extracted from the `processPluginWithProcessor` method.

pkg/client/results/processing.go
```diff
@@ -146,48 +148,120 @@ func PostProcessPlugin(p plugin.Interface, dir string) (Item, error) {
...
 	results := Item{
 		Name: p.GetName(),
 	}
 
 	if isDS {
-		nodeDirs, err := ioutil.ReadDir(pResultsDir)
+		items, err := processNodesWithProcessor(p, baseDir, pResultsDir, processor, selector)
 		if err != nil {
-			return Item{}, err
+			logrus.Warningf("Error processing results entries for plugin %v: %v", p.GetName(), err)
```
johnSchnake: TODO: rather than log here, I want to return a slice of errors and just return them all and log at the parent if we want. Awkward to have lower level methods not pass up their errors.

I really just wanted to continue processing after errors so on my first pass I added logging rather than returning.

pkg/client/results/processing.go
```diff
@@ -146,48 +148,120 @@ func PostProcessPlugin(p plugin.Interface, dir string) (Item, error) {
...
+		return resultObj, errors.Wrapf(err, "opening file %v", currentFile)
+	}
+	defer infile.Close()
+
+	dec := json.NewDecoder(infile)
+	result := map[string]string{}
+	if err := dec.Decode(&result); err != nil {
+		return resultObj, errors.Wrapf(err, "decoding file %v", currentFile)
+	}
+	logrus.Infof("SCHNAKE got errors.json file with %#v", result)
```
johnSchnake: Debug line.

pkg/client/results/testdata/mockResults/plugins/ds-errors-02/ds-errors-02.golden.json
```diff
@@ -0,0 +1,41 @@
...
+"details": {
+"data": "foo",
+"error": "testerr"
+}
+}
+]
+},
+{
+"name": "node2",
+"status": "",
```
johnSchnake: This shouldn't be empty. #937 had seen this before and understood it to be for unknown nodes but here it is something that has a failed leaf node. Need to identify why this isnt bubbling up.

johnSchnake: Found the other bug; it was that `aggregateStatuses` was doing early exit on failures. That would bubble up the right values, but wasn't ensuring that every node was visited. #937 has been updated.

johnSchnake:
Green tests, reasonable coverage. Going to just merge and move forward. Can fixup if desired but this is a blocker for the #883 which is a slightly larger change and I want ready for review soon.

PR #937, opened 2019-10-03
Process all nodes and treat empty status as unknown

### LINKED ISSUE #900: Rolling up junit results on daemonset plugin incorrect
I had a plugin that ran a junit producing plugin on a 2 node cluster as a daemonset.

One of the nodes, which had test failures, did not get a status at all:

```
- name: kind-worker
  status: ""
  items:
  - name: output.xml
    status: failed
    meta:
      file: results/kind-worker/tmp/results/output.xml
    items:
    - name: benchmark
      status: failed
      items:
      - name: benchmark-1.7.1
        status: skipped
```

The other node in the cluster did get a properly bubbled-up `failed` status.

Each nodes results should be aggregated appropriately.

At this time I don't want to share the plugin I was using as it is a WIP but I think the bug should be fairly easy to repro/identify in the code.

johnSchnake:
If a node isn't marked with a status, it should be treated as
being unknown and bubble up accordingly.

Fixes #900

**Release note**:
```
Fixed a bug where some plugins may have reported a final status which was empty to `sonobuoy status` or `sonobuoy results`. These are now treated as "unknown".
```

### DIFF
--- a/pkg/client/results/processing.go
+++ b/pkg/client/results/processing.go
@@ -97,24 +97,31 @@ func aggregateStatus(items ...Item) string {
 		return StatusUnknown
 	}
 
-	unknownFound := false
+	failedFound, unknownFound := false, false
 	for i := range items {
 		// Branches should just aggregate their leaves and return the result.
 		if len(items[i].Items) > 0 {
 			items[i].Status = aggregateStatus(items[i].Items...)
 		}
 
-		// Any failures immediately fail the parent.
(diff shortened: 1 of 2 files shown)


johnSchnake:
Small simple bug fix with 100% coverage. Merging after green in CI. Will rebase 
#936 onto this since I noticed it impacted some of the golden file tests that it introduced.

PR #940, opened 2019-10-07
Add initial CircleCI config

zubron:
This initial config replicates some of the steps from our TravisCI
config. It introduces a new workflow that contains 3 jobs. It currently
only handles the steps that take place on PR branches:

* Checking that the READMEs are in sync
* Checking the go module state
* Running unit, stress, integration tests and a single Sonobuoy run

It does not yet handle image pushing. That will be addressed in a later
PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,68 @@
+version: 2
+jobs:
+  check_go_mod:
+    docker:
+      - image: golang:1.13
+    steps:
+      - checkout
+      - run:
+          name: "Check go module up to date"
+          command: ./scripts/ci/check_go_modules.sh
+
+  check_readme_sync:
+    docker:
+      - image: golang:1.13
+    steps:
+      - checkout
+      - run:
+          name: "Check README in sync"
+          command: ./scripts/ci/check_readme_in_sync.sh
+
+  build:
+    machine:
+      enabled: true
+      docker_layer_caching: true
+    steps:
+      - checkout
+      - run:
+          name: "Install kind"
+          command: |
+            curl -L https://github.com/kubernetes-sigs/kind/releases/download/v0.5.1/kind-linux-amd64 --output kind
+            chmod +x ./kind
+            sudo mv ./kind /usr/local/bin
+      - run:
+          name: "Create kind cluster"
+          command: |
+            kind create cluster --config kind-config.yaml
+            echo 'export KUBECONFIG="$(kind get kubeconfig-path --name="kind")"' >> $BASH_ENV
+      - run:
+          name: "Build Sonobuoy container image and add to kind cluster"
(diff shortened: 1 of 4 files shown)


scripts/ci/check_e2e_tarball.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+CI_SCRIPTS_DIR="$( cd "$( dirname "$0" )" >/dev/null 2>&1 && pwd )"
```
zubron: The contents of this scripts were taken as is from the existing [travis-ci.sh script](https://github.com/vmware-tanzu/sonobuoy/blob/master/travis-ci.sh#L30-L72). It still uses the `e2e` command below. I would like to change that to use the new results command but thought that that should be part of a later change.

.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
...
+            ./scripts/run_integration_tests.sh
+      - run:
+          name: "Start Sonobuoy run, wait, and gather results"
+          command: |
+            ./sonobuoy run --image-pull-policy=IfNotPresent -m Quick --wait
+            tarball=$(./sonobuoy retrieve)
+            ./scripts/ci/check_e2e_tarball.sh $tarball
+            mv $tarball /tmp/sonobuoy-tarball.tar.gz
+      - store_artifacts:
+          path: /tmp/sonobuoy-tarball.tar.gz
```
johnSchnake: Are you seeing the artifacts actually get uploaded? I dont see where to view them at
image

johnSchnake: Still curious about this, but I just opted-in to the new UI on CircleCI and the artifacts were visible:
image

zubron: Are you looking at the `build` job? The artifacts are only available within that one job. It's not very intuitive but they are being uploaded: https://circleci.com/gh/vmware-tanzu/sonobuoy/95#artifacts/containers/0

I don't know if there's a way to see all the artifacts from the workflow overview. I can take a look into that to see if there's a way to make that a bit clearer.

zubron: I used the `store_artifacts` step. I guess I should use `store_test_results` instead: https://circleci.com/docs/2.0/configuration-reference/#store_test_results

zubron: Never mind, that seems to only be for specific test formats (JUnit, Cucumber etc)

johnSchnake: Still dont see them there for some reason. /shrug I'd think it was a permission issue but on the new view I can see them so it probably isnt that.

Devs are the only ones that really need to see them so if we both have _some_ way of seeing it its fine with me. Wanted to make sure they were getting uploaded was all. I'll just use the updated UI

Issue #926, opened 2019-10-01
Move from Travis to CircleCI


johnSchnake:
**Describe the solution you'd like**
Move from Travis CI to CircleCI.

Most of our scripts should be usable but may need to be tweaked slightly.

Once this is done we can configure it to push images from master builds and tagged releases (See #914)

PR #945, opened 2019-10-09
Move image pushing into CircleCI

johnSchnake:
 - Moves/updates the travis deploy script to work on circleCI.
 - Added DOCKERHUB_TOKEN via web UI to circleCI
 - runs publishing of images on master branch and on tags like v.*
 - minor change to naming to clarify job purposes

Ref: #926

I tried to use circleCI on my own repo to show this behavior to some extent. You can see here: https://circleci.com/workflow-run/9fdeb6cc-04ae-4e11-9173-bc2f06c14f40 that the publishing waits for the tests to pass. In addition, you can see https://circleci.com/workflow-run/9a9a49a3-5279-40ab-8638-9803a92063e1 doesn't show that the publishing would have occurred at all since it wasn't on the master branch.

**Release note**:
```
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,10 +18,16 @@ jobs:
           name: "Check README in sync"
           command: ./scripts/ci/check_readme_in_sync.sh
 
-  build:
+  build_and_publish:
     machine:
       enabled: true
-      docker_layer_caching: true
+    steps:
+      - checkout
+      - run: ./scripts/publish.sh
+
+  build_and_test:
+    machine:
+      enabled: true
     steps:
       - checkout
       - run:
@@ -61,8 +67,16 @@ jobs:
 
 workflows:
   version: 2
-  build_and_test:
+  checks:
     jobs:
       - check_go_mod
       - check_readme_sync
-      - build
+      - build_and_test
+      - build_and_publish:
+          requires:
+            - build_and_test
+          filters:
+            branches:
+              only: master
+            tags:
+              only: /^v.*/
--- a/scripts/ci/publish.sh
+++ b/scripts/ci/publish.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# Don't fail silently when a step doesn't succeed
+set -e
+
+if [ -z "$CIRCLECI" ]; then
+    echo "this script is intended to be run only on travis" >&2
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -61,8 +67,16 @@ jobs:
...
       - check_go_mod
       - check_readme_sync
-      - build
+      - build_and_test
+      - build_and_publish:
+          requires:
+            - build_and_test
+          filters:
+            branches:
+              only: master
```
zubron: I was reading about the filters earlier and came across this post:  https://discuss.circleci.com/t/dont-run-deploy-workflow-step-for-forked-pr/23606
It suggests that a PR from the master branch of a fork would pass this filter. We might want to consider expanding the check in the publish script to check for one of the environment variables that are only set for forked PRs: https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

What do you think?

johnSchnake: Great question.

So you're right on the approach (using env vars) and we do this though by checking the CIRCLE_BRANCH. For pull requests it is `pull/945` or whatnot.

However, as I was typing my response I realized that for the _tagged_ portion of the bash script we could hit a problem. E.g. If I make a PR that includes a tag on my commit I may be able to trigger that part of the build.

We can make this check a bit more verbose by checking for another env var value about the PR, but we should acknowledge that a malicious user could still manipulate the scripts/env vars to trigger these. I dont know if we can prevent that.

zubron: I just checked the `CIRCLE_BRANCH` in a test PR and you're right, it sets the branch correctly to be a pull request branch:
```
#!/bin/bash -eo pipefail
echo $CIRCLE_BRANCH
echo $CIRCLE_PR_NUMBER
echo $CIRCLE_PR_REPONAME
echo $CIRCLE_PR_USERNAME
pull/946
946
sonobuoy
zubron
```

I'm not sure if there is anything that we can do to prevent malicious PRs. But, given that we've prevented the secrets being passed to forked PRs, they could bypass the checks but the steps shouldn't work because the secrets aren't available.

johnSchnake: Yes, excellent point. The config settings in the UI prevent image pushing from ever actually occurring.

PR #949, opened 2019-10-10
Unity integration testing and remove Travis

johnSchnake:
 - Removes travis scripts
 - Uses a single kind cluster
 - Removes the integration test that originally existed, running
just a quick mode run
 - Adds the functionality from the above test into an integration
test
 - Changes the integration suite to use the 'kind' cluster
 - Modifies/adds various test methods to run commands with less
boilerplate code
 - Adds a test method to move artifacts to a desired location for
saving in CircleCI

Ref #926 

**Release note**:
```
NONE
```


johnSchnake:
Green right now and shaved 1-2m off the execution time but it has a bug regarding the artifact uploading. Will figure it out this morning and should be ready to merge.

johnSchnake:
`invalid cross-device link` when trying to move the tarball via golang. I'll probably just have to shell out and call `mv` which is what it was doing before.

go.sum
```diff
@@ -179,18 +188,23 @@ github.com/pmezard/go-difflib v0.0.0-20151028094244-d8ed2627bdf0/go.mod h1:iKH77
...
 github.com/prometheus/common v0.4.0/go.mod h1:TNfzLD0ON7rHzMJeJkieUDPYmFC7Snx/y86RQel1bk4=
 github.com/prometheus/procfs v0.0.0-20181005140218-185b4288413d/go.mod h1:c3At6R/oaqEKCNdg8wHV1ftS6bRYblBhIjjI8uT2IGk=
+github.com/prometheus/procfs v0.0.0-20190507164030-5867b95ac084 h1:sofwID9zm4tzrgykg80hfFph1mryUeLRsUfoocVVmRY=
 github.com/prometheus/procfs v0.0.0-20190507164030-5867b95ac084/go.mod h1:TjEm7ze935MbeOT/UhFTIMYKhuLP4wbCsTZCD3I8kEA=
 github.com/prometheus/tsdb v0.7.1/go.mod h1:qhTCs0VvXwvX/y3TZrWD7rabWM+ijKTux40TwIPHuXU=
 github.com/rifflock/lfshook v0.0.0-20180920164130-b9218ef580f5 h1:mZHayPoR0lNmnHyvtYjDeq0zlVHn9K/ZXoy17ylucdo=
 github.com/rifflock/lfshook v0.0.0-20180920164130-b9218ef580f5/go.mod h1:GEXHk5HgEKCvEIIrSpFI3ozzG5xOKA2DVlEX/gGnewM=
 github.com/rogpeppe/fastuuid v0.0.0-20150106093220-6724a57986af/go.mod h1:XWv6SoW27p1b0cqNHllgS5HIMJraePCO15w5zCzIWYg=
 github.com/rogpeppe/go-charset v0.0.0-20180617210344-2471d30d28b4/go.mod h1:qgYeAmZ5ZIpBWTGllZSQnw97Dj+woV0toclVaRGI8pc=
+github.com/russross/blackfriday v1.5.2 h1:HyvC0ARfnZBqnXwABFeSZHpKvJHJJfPz81GNueLj0oo=
```
johnSchnake: I'm not exactly sure why the go.sum got changed in all of this. No meaningful imports were changed.

johnSchnake: `go mod tidy` didnt do anything meaningful here but when I reverted these changes then ran `go mod tidy` and related commands they did not get added back. Maybe during a test iteration they got added and the `go.sum` file doesnt get cleaned up at all when doing tidy I guess.

test/integration/sonobuoy_integration_test.go
```diff
@@ -46,15 +49,43 @@ func runSonobuoyCommandWithContext(ctx context.Context, t *testing.T, args strin
...
+func runSonobuoyCommand(t *testing.T, args string) (bytes.Buffer, bytes.Buffer, error) {
 	return runSonobuoyCommandWithContext(context.Background(), t, args)
 }
 
+// getNamespace returns the namespace to use for the current test and a function to clean it up
+// asynchronously afterwards.
+func getNamespace(t *testing.T) (string, func()) {
+	ns := "sonobuoy-" + strings.ToLower(t.Name())
+	return ns, func() { cleanup(t, ns) }
+}
```
johnSchnake: Expanded the helper; cleanup was being used throughout but the boilerplate for each test was a bit verbose. This simplifies the calls.

test/integration/sonobuoy_integration_test.go
```diff
@@ -46,15 +49,43 @@ func runSonobuoyCommandWithContext(ctx context.Context, t *testing.T, args strin
...
+	return stdout, stderr, command.Run()
+}
+
+func mustRunSonobuoyCommand(t *testing.T, args string) bytes.Buffer {
+	return mustRunSonobuoyCommandWithContext(context.Background(), t, args)
+}
+
+// mustRunSonobuoyCommandWithContext runs the Sonobuoy CLI with the given context and arguments.
+// It returns stdout and fails the test immediately if there are any errors.
+func mustRunSonobuoyCommandWithContext(ctx context.Context, t *testing.T, args string) bytes.Buffer {
```
johnSchnake: Helper added since we already were repeatedly calling sonobuoy functions and erroring/logging on failure. This just makes the tests easier to read/write if we want to just be able to code along the happy path.

Unhappy path testing is still supported via the original methods.

test/integration/sonobuoy_integration_test.go
```diff
@@ -113,31 +138,129 @@ func TestUseNamespaceFromManifest(t *testing.T) {
...
+	origFile := filepath.Join(pwd(t), filepath.Base(p))
+
+	if err := os.MkdirAll(artifactsDir, 0755); err != nil {
+		t.Logf("Error creating directory %v: %v", artifactsDir, err)
+		return p
 	}
 
+	var stdout, stderr bytes.Buffer
+
+	cmd := exec.CommandContext(context.Background(), bash, "-c", fmt.Sprintf("mv %v %v", origFile, artifactFile))
```
johnSchnake: Need to add a comment on here about why we do this instead of `os.Rename`; it wont be clear to readers.

test/integration/sonobuoy_integration_test.go
```diff
@@ -9,13 +9,16 @@ import (
...
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
 )
 
 const (
 	defaultSonobuoyPath = "../../sonobuoy"
+	bash                = "/bin/sh"
```
zubron: My preference here would be to use `/bin/bash` here to avoid unexpected differences between the shells. I think we've been caught out before by differences between these two. I think it's typically symlinked to different shells on different systems, dash on Linux (and most of the container images we use) and bash on macOS.

test/integration/sonobuoy_integration_test.go
```diff
@@ -104,7 +129,7 @@ func TestUseNamespaceFromManifest(t *testing.T) {
 
 	// Pass the gen output to sonobuoy run
 	runArgs := fmt.Sprintf("run --wait -f %v", tmpfile.Name())
-	err, _, runStderr := runSonobuoyCommandWithContext(ctx, t, runArgs)
+	_, runStderr, err := runSonobuoyCommandWithContext(ctx, t, runArgs)
```
zubron: I think we can use the `mustRun` helper here?

test/integration/sonobuoy_integration_test.go
```diff
@@ -113,31 +138,130 @@ func TestUseNamespaceFromManifest(t *testing.T) {
...
 	defer cancel()
 
-	ns := "sonobuoy-" + strings.ToLower(t.Name())
-	defer cleanup(t, ns)
+	ns, cleanup := getNamespace(t)
+	defer cleanup()
 
 	args := fmt.Sprintf("run --image-pull-policy IfNotPresent --wait -p testImage/yaml/job-junit-passing-singlefile.yaml -n %v", ns)
-	err, _, stderr := runSonobuoyCommandWithContext(ctx, t, args)
+	_, stderr, err := runSonobuoyCommandWithContext(ctx, t, args)
```
zubron: Again, we can use the helper here.

test/integration/sonobuoy_integration_test.go
```diff
@@ -113,31 +138,130 @@ func TestUseNamespaceFromManifest(t *testing.T) {
...
+	switch {
+	case failCount == 0:
+		expectVals = []string{
+			`"status":"complete","result-status":"passed"`,
+			`"passed":1`,
+		}
+	case failCount > 0:
+		expectVals = []string{
+			`"status":"complete","result-status":"failed"`,
+			fmt.Sprintf(`"passed":%v`, failCount),
```
zubron:
```suggestion
			fmt.Sprintf(`"failed":%v`, failCount),
```

.circleci/config.yml
```diff
@@ -36,6 +36,9 @@ jobs:
             curl -L https://github.com/kubernetes-sigs/kind/releases/download/v0.5.1/kind-linux-amd64 --output kind
             chmod +x ./kind
             sudo mv ./kind /usr/local/bin
+            curl -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 --output jq 
+            chmod +x ./jq
+            sudo mv ./jq /usr/local/bin
       - run:
```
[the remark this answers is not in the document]
johnSchnake: I'd rather create it sooner and leave it as is for now. We can see if it doesnt work well that way.

The proximity of this comment to the `jq` lines made me realize I wasn't using `jq` yet though. I do think having that tool will make some future tests easier so I'm inclined to just leave it for now so we dont have to re-add those later.

johnSchnake:
Waiting to go green again then we'll commit this and update circleci to be the default/required checks and disable travis.

PR #953, opened 2019-10-11
Update badge to be from CircleCI not Travis

johnSchnake:


johnSchnake:
FYI I got a weird error on my first push with the check_readme_sync script. Locally it seemed clean; in the logs for circleCI I couldnt tell what the problem was. In the end the resolution was to ensure that each file had the ending newline. Not sure why that would cause the problem, but it seemed to resolve it for me FYI in case we see it again.

johnSchnake:
Went ahead and updated the old docs too since it was very simple/procedural. I didnt want folks to see failing on travis and get the wrong idea.

PR #956, opened 2019-10-11
Update release process docs to reference new systems

zubron:
We recently moved our CI to CircleCI and our image pushing to Docker
Hub. This updates our docs to reference these new systems.

Also, this removes the link from the side bar as it's only really useful
for the development team.

### DIFF
--- a/site/_data/shortlinks.yml
+++ b/site/_data/shortlinks.yml
@@ -22,6 +22,3 @@
 - title: Make your own plugins
   key: examples
   destination: examples
-- title: Cut a release
-  key: release
-  destination: release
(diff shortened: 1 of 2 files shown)


site/_data/shortlinks.yml
```diff
@@ -22,6 +22,3 @@
 - title: Make your own plugins
   key: examples
   destination: examples
-- title: Cut a release
```
zubron: I'm removing this but I don't think it has an effect since the side bar has already been reorganised.

zubron: docs/master/release still works. I'm not really sure what this file is for.

PR #960, opened 2019-10-11
Update github org in one of our scripts

johnSchnake:
Without this update, the config.yml doesnt get updated which
cases the versioned docs to not be formatted properly.


zubron:
Can you wait until after the release PR goes through before merging this so I can avoid a rebase and another CI run?

johnSchnake:
Yup absolutely

PR #961, opened 2019-10-11
Adjust our CircleCI to list tags on required job

johnSchnake:
Our latest release didnt trigger a tagged release in circleCI; I think this was the cause.


johnSchnake:
I want to try this out before the weekend. The release isn't really a rush, but if we can get it tagged/published/solved sooner is better. Then monday we can announce and fixup the release notes.

johnSchnake:
Config validated; this change wouldnt impact the actual tests. Merging and re-attempting.

PR #968, opened 2019-10-18
Start time status

### LINKED ISSUE #587: `Status` start time
It would be neat to see the start time of plugins when looking at `sonobuoy status`. For example, when E2E are running, they may take a while, but when did they start?

```
$ sonobuoy status                                                                                                            
PLUGIN		STATUS		COUNT
e2e		running		1
systemd_logs	complete	1

```

spiffcs:
Update status to show duration for a plugin

- Fixes #587 

I'm looking for feedback on how to hook into and measure the `e2e` plugin

**Release note**:
```
release-note
```

### DIFF
--- a/cmd/sonobuoy/app/status.go
+++ b/cmd/sonobuoy/app/status.go
@@ -176,7 +176,7 @@ func printSummary(w io.Writer, status *aggregation.Status) error {
 	}
 
 	statusTimeKey := func(p aggregation.PluginStatus) string {
-		return p.StartTime + "-" + p.Duration
+		return p.StartTime.String() + "-" + p.Duration.String()
 	}
 
 	for _, pStatus := range status.Plugins {
@@ -194,13 +194,14 @@ func printSummary(w io.Writer, status *aggregation.Status) error {
 	for pluginName, results := range totals {
 		for statusAndResult, pluginStats := range results {
 			for startAndCurrent, count := range pluginStats {
+				fmt.Printf("%v", startAndCurrent)
 				summaries = append(summaries, pluginSummary{
 					plugin:    pluginName,
 					status:    strings.Split(statusAndResult, ":")[0],
 					result:    strings.Split(statusAndResult, ":")[1],
 					count:     count,
-					startTime: strings.Split(startAndCurrent, "-")[0],
-					duration:  strings.Split(startAndCurrent, "-")[1],
+					startTime: strings.Split(startAndCurrent, "")[0],
+					duration:  strings.Split(startAndCurrent, "")[1],
 				})
 			}
 		}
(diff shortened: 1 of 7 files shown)


pkg/time/time.go
```diff
@@ -8,3 +8,15 @@ import (
...
+// to pointers for the json STATUS
+func Time(time time.Time) *time.Time {
+	return &time
+}
+
+// Duration is a function that helps convert test duration values
+// to pointers for the json STATUS
+func Duration(duration time.Duration) *time.Duration {
+	return &duration
+}
```
johnSchnake: I'd remove these; either move them into the test if you really want those helpers but it looks like you just need 2 test values for reuse:
```
testTime := time.Date(...)
testDuration := time.Duration(100) * time.Second
...
...
StartTime:=&testTime
Duration := &testDuration
...
```

If something is mutating those and you need new values each time, then I'd just move these helpers.

johnSchnake:
So along with that screenshot it is interesting to see this is the underlying json:
```
{"plugins":[{"plugin":"e2e","node":"global","status":"running","result-status":"","result-counts":null,"start_time":"2019-10-24T14:33:59.9497703Z"},{"plugin":"systemd-logs","node":"kind-control-plane","status":"complete","result-status":"","result-counts":null,"start_time":"2019-10-24T14:33:59.9497703Z","duration":553846686300},{"plugin":"systemd-logs","node":"kind-worker","status":"complete","result-status":"","result-counts":null,"start_time":"2019-10-24T14:33:59.9497703Z","duration":553846676600}],"status":"running","tar-info":{"name":"","created":"0001-01-01T00:00:00Z","sha256":"","size":0}}
```

So the e2e plugin has a start time but isn't having the duration calculated or printed properly.

cmd/sonobuoy/app/status.go
```diff
@@ -165,37 +169,50 @@ func printAll(w io.Writer, status *aggregation.Status) error {
 
 func printSummary(w io.Writer, status *aggregation.Status) error {
 	tw := defaultTabWriter(w)
-	totals := map[string]map[string]int{}
+	totals := map[string]map[string]map[string]int{}
```
johnSchnake: I already had thought this was hard to follow; can we avoid adding the extra level of map? I'd like to not show each plugins duration either, just the max of the values in the group. If someone adds the `--show-all` then we would show duration of each.

Speaking of `--show-all`, that doesn't hav duration at all which it should now.
image

johnSchnake:
@spiffcs Are you still able to finish this off? I know I came up with a few pieces of feedback that may be nontrivial. If you're open to it, I can even take/squash your commits and finish it off on my own. You'd still be marked as the author on our set of the work then.

Let me know.

PR #989, opened 2019-11-10
Allow plugins to be specified via a URL

### LINKED ISSUE #986: Load plugins from a given URL
When someone specifies the `--plugin` flag it can currently be:
 - a known word like `e2e` or `systemd-logs`
 - a name of a local file which holds the plugin definition

I think a 3rd option should be supported to aid in sharing/experimenting with plugins: loading from a URL.

I think this is pretty trivial to do; we just need to check the string and see if it parses as a URL. If it does, do a GET to the URL and the response body is the io.Reader that we can pass around.

This would not be the most secure feature since the user _could_ be targeting a URL they haven't verified the contents of or they could be subject to MITM attacks but I think this is still a useful, feature to make experimentation/sharing more simple.

johnSchnake:
We currently allow loading plugins from the local file system
but for sharing of plugins it makes sense to allow users to
specify a URL.

Fixes #986

**Release note**:
```
The `--plugin` flag now supports URLs in addition to paths to local files or directories.
```

### DIFF
--- a/cmd/sonobuoy/app/pluginList.go
+++ b/cmd/sonobuoy/app/pluginList.go
@@ -17,15 +17,20 @@ limitations under the License.
 package app
 
 import (
+	"fmt"
+	"io"
 	"io/ioutil"
+	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/vmware-tanzu/sonobuoy/pkg/plugin/manifest"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
-
 	kuberuntime "k8s.io/apimachinery/pkg/runtime"
 )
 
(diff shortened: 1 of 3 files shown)


johnSchnake:
Not sure why but the 2 circleci checks are hung; they seemed to have pass but just not get reported to Github

cmd/sonobuoy/app/pluginList.go
```diff
@@ -108,16 +125,42 @@ func (p *pluginList) loadPluginsDir(dirpath string) error {
...
+	defer r.Close()
+	b, err := ioutil.ReadAll(r)
+	if err != nil {
+		return errors.Wrap(err, "failed to read data for plugin")
+	}
 
 	newPlugin, err := loadManifest(b)
 	if err != nil {
-		return errors.Wrapf(err, "failed to load plugin file %q", filepath)
+		return errors.Wrap(err, "failed to load plugin")
```
zubron: Minor usability nit: If the user specifies multiple plugins and one fails to decode, it seems like it won't be as easy to determine which one had the issue. Instead, could you return the unwrapped error here and wrap where this function is called in `loadSinglePluginFromURL` and `loadSinglePluginFromFile`? That way both of those functions could include the location information of the plugin in the error message.

johnSchnake: Instead of just returning the value of `loadSinglePlugin` I wrapped that value with the context necessary to describe the URL or file. 👍

PR #990, opened 2019-11-11
Allow daemonset plugins running on subset of nodes

### LINKED ISSUE #988: Improve daemonset monitoring logic
Currently we assume a deamonset plugin will run on every node and therefore we expect one from every node, block until we get results from them all, and error if we fail to see a pod startup on a given node.

However, a daemonset can set affinity and other options which impact scheduling which should be allowed. E.g. If I want to run a daemonset plugin that will run on every non-master node. Currently the non-master nodes will work fine but errors will be reported for all the master nodes since pods did not start up there.

Some details that may be utilized are in in the daemonsetStatus field: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#daemonsetstatus-v1-apps

This indicates the # of intended pods and the # that are actually scheduled. This would allow us to not-fail if everything was going well. However, if we had 5 nodes, wanted 3 pods, but only had 1: what do we report the error for? What results we we "wait" for? Currently the logic in the aggregator uses `node name + plugin name` as a key; if we dont know _which_ nodes should be scheduled for the plugin we run into a problem. I'm not sure if we can determine that info easily.

johnSchnake:
For daemonsets, node affinity can be set to limit/select the nodes
which the daemonset will run on. Since this is a native k8s feature
we should support it.

However, we use 'node-plugin' as a key in a few places which need
to understand that not every node will be considered.

The upstream logic for this node selection is in the core scheduling
packages and not readily consumable as a module. As a result, the
most reasonable path forward is to implement a small bit of the
desired functionality on our end.

For now, we just care about the RequiredDuringSchedulingIgnoredDuringExecution
field and only for node labels. This allows picking out nodes based on
labels that may satisfy most use cases.

Fixes #988

**Release note**:
```
Daemonset plugins may now specify node affinity in their podSpec without extraneous errors. Until now, this could be specified but Sonobuoy would record errors due to "missing" plugins on the non-targeted nodes. Currently support In/NotIn/Exists/DoesNotExists operators only.
```

### DIFF
--- a/pkg/plugin/driver/daemonset/daemonset.go
+++ b/pkg/plugin/driver/daemonset/daemonset.go
@@ -70,6 +70,7 @@ func NewPlugin(dfn manifest.Manifest, namespace, sonobuoyImage, imagePullPolicy,
 
 // ExpectedResults returns the list of results expected for this daemonset.
 func (p *Plugin) ExpectedResults(nodes []v1.Node) []plugin.ExpectedResult {
+	nodes = p.filterByNodeSelector(nodes)
 	ret := make([]plugin.ExpectedResult, 0, len(nodes))
 
 	for _, node := range nodes {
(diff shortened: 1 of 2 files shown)


johnSchnake:
I manually tested this slightly with just a `DoesNotExist` case to run a plugin just on my worker nodes and it worked well. It should be quite easy to test these functions.

PR #998, opened 2019-11-15
Add an FAQ

zubron:
This adds an FAQ to our documentation site. It mostly covers questions
related to the e2e plugin, but some of the information applies to
plugins in general.


site/docs/master/faq.md
```diff
@@ -0,0 +1,11 @@
+# Frequently Asked Questions
+
+## Conformance testing
+### Why were so many tests skipped?
+<answer>
+
+### Why did my test fail?
+<answer>
+
+### How can I configure settings when running the E2E tests?
```
johnSchnake: I really like how you're first just figuring out the questions instead of spending more time on the answers right away. I'll try and suggest some in the PR too

johnSchnake:
Brainstorm of faq items:
 - how can i run with certain test framework options set? What options are there?
 - Some of the registries necessary for tests is blocked from our test infra; can I still run tests?
 - We have some nodes with custom taints in our cluster and the tests wont start, how can I run the tests?
 - What tests can I run? How can I figure out what tests/tags I can select?

site/docs/master/faq.md
```diff
@@ -0,0 +1,216 @@
...
+To do this, you can make use of the `--e2e-focus` flag when using the `run` command.
+This flag accepts a regex which will be used to find and run only the tests matching that regex.
+For example, you can provide the name of a test to run only that test:
+
+```
+sonobuoy run --e2e-focus "should update pod when spec was updated and update strategy is RollingUpdate"
+```
+
+If the test continues to fail and it appears to be a genuine failure, the next step would be to read the logs to understand why the test failed.
+To read the logs for a test failure, you can fnd the log file within the results tarball from Sonobuoy (`plugins/e2e/results/global/e2e.log`) or you can use the `results` command to show details of test failures.
```
johnSchnake: s/fnd/find

site/docs/master/faq.md
```diff
@@ -0,0 +1,216 @@
...
+To view the available options that you can set when running the tests, you can run the test executable for the conformance image you will be using as follows:
+
+```
+KUBE_VERSION=<Kubernetes version you are using>
+docker run -it gcr.io/google-containers/conformance:$KUBE_VERSION ./e2e.test --help
+```
+
+If you are running Kubernetes v1.16.0 or greater, a new feature was included in this release which makes it easier to specify your own options.
+This new feature allows arbitrary options to be specified when the tests are invoked.
+To use this, you must enable the new feature with the flag `--plugin-env=e2e.E2E_USE_GO_RUNNER=true` and then you can provide your options with the flag `--plugin-env=e2e.E2E_EXTRA_ARGS`.
```
johnSchnake: `E2E_USE_GO_RUNNER` is the default now; we should make that clear instead of making it seem like that has to be manually set.

Maybe something like
```
... you must ensure the env var `E2E_USE_GO_RUNNER=true` is set (this is the default from the CLI and manually setting this only required if working from a saved YAML file...
```

Feel free to tweak the wording.

site/docs/master/faq.md
```diff
@@ -0,0 +1,216 @@
...
+- NodeConformance
+- Slow
+- Serial
+- Disruptive
+- Flaky
+- LinuxOnly
+- Feature:* (there are numerous feature tags)
+
+There are also specific tags for tests that belong to a particular [Special Interest Group (SIG)][sig-list].
+As of writing this document, we were able to see the following SIG tags within the E2E tests:
```
johnSchnake: I'd just say "the tests included the following SIG tags:" instead of qualifying it with "we were able to see...within the E2E tests"

site/docs/master/faq.md
```diff
@@ -0,0 +1,219 @@
...
+
+For details on the two different approaches that you can take, please refer to [our blog post][custom-e2e-image] which describes in more detail how to use the new v1.16.0 Go test runner and how to build your own custom images.
+
+
+### Some of the registries required for the tests are blocked with my test infrastructure. Can I still run the tests?
+Yes! Sonobuoy can be configured to use custom registries so that you can run the tests in airgapped environments.
+
+For more information and details on how to configure your environment, please refer to [our documentation for custom registries and air-gapped environments][airgap].
+
+### We have some nodes with custom taints in our cluster and the tests won't start. How can I run the tests?
```
johnSchnake: I'm dont think the podspec customization is the answer to this question.

So there are two use cases:
 - User has some nodes tainted and keep seeing messages about how the e2e tests are waiting for all the nodes to be schedulable (which doesnt happen) and then it times out.
 - User has all the nodes tainted so the same issue as the above happens, but even if they solve that issue, then all the jobs that the e2e tests try and make will not be schedulable and so tests are still not runnable.

In the first case, the upstream PR https://github.com/kubernetes/kubernetes/pull/81043 added a flag to allow the e2e framework to ignore certain taints when checking 'are all nodes ready to start tests'

In the 2nd case, an issue is still open to address this but will require that the e2e framework create a mutating webhook so that all jobs will tolerate certain taints when launched.

I think the podspec answer you gave answers a separate set of issues:
 - I have a tainted environment and I can't get my plugin to launch on any nodes, can I still run it?

and obviously it also answers the questions:
 - I need certain podspec options (like hostPID) set on my plugin, is that possible?
 - I need to run my plugin only on certain nodes, is that possible?

zubron: Yep, you're totally right. Sorry, I forgot about that PR and was only thinking about it launching the actual plugin pods, not all the other things that e2e framework launches itself.

zubron: I've updated this answer but it's likely that I am missing some information/context that would be helpful to mention here. Let me know what you think.

PR #1007, opened 2019-11-20
Update to log logic in order to stream from new containers as created

### LINKED ISSUE #703: Sonobuoy logs should start following new pods when they start

johnSchnake:
- Adds a method which adds a watch for new pods in the target namespace
 - Removes fallback/retry logic which mutates the query from a specific
plugin to all plugins.
 - Adds a ceiling for the max polling time when backing off due to a container
not being in a running state
 - Adjusts tests as necessary

Fixes #703

A simple way to test this is to do:
```
sonobuoy run && sonobuoy logs -f
```

You'll now see that the e2e logs get streamed when the container is created; that would not happen with the older logic.

**Release note**:
```
Sonobuoy will now begin tailing logs from new containers as they are created while running `sonobuoy logs -f`
```

### DIFF
--- a/pkg/client/logs.go
+++ b/pkg/client/logs.go
@@ -23,14 +23,18 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 )
 
 const (
 	bufSize = 4096
+
+	// maxBackoffSeconds is the maximum time to backoff when waiting for pods to startup. Prevents
+	// unhelpfully large noop periods.
+	maxBackoffSeconds = 32
 )
 
 // Reader provides an io.Reader interface to a channel of bytes. The first error
@@ -41,12 +45,12 @@ const (
 type Reader struct {
 	bytestream chan []byte
 	errc       chan error
-	done       chan struct{}
 
 	// Used for when one message is too large for the input buffer
 	// TODO(chuckha) consider alternative data structures here
 	overflowBuffer []byte
 	err            error
+	empty          bool
 }
 
 // NewReader returns a configured Reader.
@@ -64,11 +68,6 @@ func NewReader(bytestream chan []byte, errc chan error) *Reader {
 // Read manages the message overflow ensuring no bytes are missed.
 // If an error is set on the reader it will return the error immediately.
 func (r *Reader) Read(p []byte) (int, error) {
-	// Always return the error if it is set.
-	if r.err != nil {
-		return 0, r.err
-	}
-
 	// Send any overflow before grabbing new messages.
 	if len(r.overflowBuffer) > 0 {
 		// If we need to chunk it, copy as much as we can and reduce the overflow buffer.
@@ -84,20 +83,50 @@ func (r *Reader) Read(p []byte) (int, error) {
 		return n, nil
 	}
 
-	data, ok := <-r.bytestream
-	// If the bytestream is done then save the error for future calls to Read.
-	if !ok {
-		r.err, ok = <-r.errc
-		// Assume EOF if error channel is closed.
-		if r.err == nil && !ok {
+	// Return the error if set only if byte channel is empty.
+	if r.err != nil && r.empty {
+		return 0, r.err
+	}
+
+	// Be sure that we are draining any errors put on the errc and unblock senders.
+	var err error
+	var data []byte
+	var ok bool
+	select {
+	case err, ok = <-r.errc:
+		// Treat the errors as data as well. This ensures that the user reading the messages
+		// is aware of the errors as they happen and not only after all messages are processed.
+		// This is important in the case of tailing logs where one stream may never finish at all.
+		if err != nil && err != io.EOF {
+			data = []byte(fmt.Sprint(err))
+		}
+
+		// Save only the first error. Assume io.EOF if closed/nil.
+		switch {
+		case err != nil:
+			r.err = err
+		case err == nil || !ok:
 			r.err = io.EOF
 		}
-		if r.err != nil {
-			return 0, r.err
+
+		// If the channel is closed, set to nil to avoid reading this channel again.
+		if !ok {
+			r.errc = nil
+		}
+	case data, ok = <-r.bytestream:
+		if !ok {
+			r.empty = true
+			// Ensure we don't choose this path again in the select.
+			r.bytestream = nil
 		}
 	}
 
-	// TODO(chuckha) this code and the code above in the overflow buffer is identical. Might be an indication of a cleaner way to do this.
+	// Avoid the case where a channel was closed and we don't have data to process. This would end up returning
+	// (0, nil) which is discouraged but also ends up leaving `p` unmodified which is against the documentation
+	// of io.Reader.
+	if len(data) == 0 {
+		return r.Read(p)
+	}
 
 	// The incoming data is bigger than size of the remaining size of the buffer. Save overflow data for next read.
 	if len(data) > len(p) {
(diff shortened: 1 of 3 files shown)


johnSchnake:
Doing a draft; I was getting some unexpected behavior locally where the stress test was failing but none of my logic should be impacting that. I just updated my macOS so it could be something local.

pkg/client/logs.go
```diff
@@ -41,12 +45,12 @@ const (
 type Reader struct {
 	bytestream chan []byte
 	errc       chan error
-	done       chan struct{}
```
johnSchnake: This was just never used

pkg/client/logs.go
```diff
@@ -64,11 +68,6 @@ func NewReader(bytestream chan []byte, errc chan error) *Reader {
 // Read manages the message overflow ensuring no bytes are missed.
 // If an error is set on the reader it will return the error immediately.
 func (r *Reader) Read(p []byte) (int, error) {
-	// Always return the error if it is set.
-	if r.err != nil {
-		return 0, r.err
-	}
```
johnSchnake: We really want to always drain the overflow buffer; otherwise we could have messages which were received but only partially printed.

This is especially of interest now that the message queue could be empty but we receive an error which may need to be broken into parts/printed _then_ returned.

pkg/client/logs.go
```diff
@@ -84,20 +83,43 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+
+	// Be sure that we are draining any errors put on the errc and unblock senders.
+	var err error
+	var data []byte
+	var ok bool
+	select {
+	case err, ok = <-r.errc:
+		// Treat the errors as data as well. This ensures that the user reading the messages
+		// is aware of the errors as they happen and not only after all messages are processed.
+		// This is important in the case of tailing logs where one stream may never finish at all.
```
johnSchnake: This is the contentious part of this PR; printing these as we get them. I think it is the most logical thing to do with the errors since we may be tailing 10+ streams. Even in the default case on a 2 node cluster we have the aggregator, e2e tests + sonobuoy sidecar, 2 containers to get logs and their 2 sidecars. So 7 is a minimum.

If we just hold all errors until the end; 6/7 could be failing for some meaningful reason but the user would see a blank screen just seeing the logs of the 7th container doing (maybe) nothing.

pkg/client/logs.go
```diff
@@ -111,30 +128,65 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+		for _, p := range podList.Items {
+			podCh <- &p
 		}
+		close(podCh)
+	}()
+
+	return nil
+}
 
-		logrus.Warningf("failed to find pods for plugin %q, defaulting to all pods", cfg.Plugin)
```
johnSchnake: The diff here looks really strange; I removed this "defaulting to all pods" logic since it really made little since to me why we would want to do that if we were explicitly calling a method to look for a particular pod.  Let me know if you disagree; I did have to remove a testcase or 2 as a result of this but it didn't seem like intuitive behavior.

johnSchnake: Hm; looking in the blame/history I see you added this and it wasn't to really fallback in general but it was for a backwards compatibility reason. Let me know what your thoughts are on this now; I can and add it back if needed. (PR that added this here: https://github.com/vmware-tanzu/sonobuoy/pull/868)

johnSchnake: And it looks like that PR went in around the time of Sonobuoy 0.15.3 so it has been numerous versions since then. I just want to be sure we understand the risk/incompatibility we're actually facing.

zubron: I decided to fall back to the default behaviour because that seemed like the safest thing to do, but it's still not what the user wants if they specified a plugin argument. My guess is that the warning about falling back to show all pods gets lost in the stream anyway, so perhaps it's just better to give an explicit error and bail out.

pkg/client/logs_test.go
```diff
@@ -297,24 +301,6 @@ func TestPodsForLogs(t *testing.T) {
...
-			pluginName:             "unknown-plugin",
-			expectedCallCount:      2,
-			expectedLabelSelectors: []string{"sonobuoy-plugin=unknown-plugin", ""},
-			podListErrors:          []error{nil, errors.New("error")},
-			expectedPodCount:       2,
-			expectedError:          errors.New("failed to list pods: error"),
-		},
 	}
 
 	for _, tc := range testCases {
```
zubron: Huh, I can't seem to add a comment to folded lines in the diff so I will add it here :confused: 

These test cases can be simplified a lot now given that we only ever call `List` once. Anything related to `callCount` can be removed, which means that `expectedLabelSelectors` and `podListErrors` no longer need to be arrays, they can just be a string and error.

pkg/client/logs.go
```diff
@@ -151,47 +208,62 @@ func (s *SonobuoyClient) LogReader(cfg *LogConfig) (*Reader, error) {
...
+			return nil, errors.Wrap(err, "failed to watch for new pods")
+		}
+	} else {
+		if err := getPodsToStreamLogs(client, cfg, podCh); err != nil {
+			return nil, errors.Wrap(err, "failed to list pods")
+		}
 	}
 
-	errc := make(chan error, numContainers)
+	// TODO figure out this error channel logic; we dont want to get a deadlock and we dont know
```
zubron: Are you happy that we avoid deadlock here and could remove this TODO comment?

johnSchnake:
Great comments @zubron ; everyone one of them is spot on and reminded me that I did <insert silly thing> to try and debug a random behavior while developing this maze of a small ticket. Will fixup each.

pkg/client/logs.go
johnSchnake: Yes only applied when we weren't reading from the `errc` at the same time; the old logic just drained the bytestream before ever reading from the errc once.

pkg/client/logs_test.go
```diff
@@ -191,15 +193,17 @@ func TestLogReaderNoError(t *testing.T) {
...
-				if i >= len(tc.expectedReads) {
-					t.Fatalf("Too many actual reads, not enough expected reads. BUF: %v", mybuf[:n])
-				}
-				if len(mybuf[:n]) != len(tc.expectedReads[i]) {
-					t.Errorf("Expected to read %v bytes, got %v buf: '%v' expected: '%v'", len(tc.expectedReads[i]), n, string(mybuf[:n]), tc.expectedReads[i])
-				}
-				if string(mybuf[:n]) != tc.expectedReads[i] {
-					t.Errorf("Expected '%v' got '%v'", tc.expectedReads[i], string(mybuf[:n]))
-				}
+				/*
```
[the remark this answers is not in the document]
johnSchnake: Yup; was just while debugging the test. Thanks.

pkg/client/logs_test.go
```diff
@@ -62,7 +63,7 @@ func TestLateErrors(t *testing.T) {
 
 	// We are guaranteed to eventually get the error because we never close bytestream.
 	errcount := 0
-	for i := 0; i <= 3; i++ {
+	for i := 0; i <= 10; i++ {
```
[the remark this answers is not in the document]
johnSchnake: Was just while debugging; will see if it needs to increase at all. It may need to increase 1-2 times since there are the weird cases where 0 bytes are read because of a channel closing. I preferred that to using `goto` type statements to restart the method.

pkg/client/logs.go
```diff
@@ -297,6 +358,16 @@ func (l *logStreamer) stream() {
 	}
 }
 
+// tryError is just a non-blocking write to the channel. This is desireable since the Reader will
+// report the first error after it drains the log channels. As a result, we may have multiple senders
+// on this channel and we dont know how many and we dont want them all to block.
+func tryError(ch chan error, err error) {
```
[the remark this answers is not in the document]
johnSchnake: I think that was needed before I started really draining the channel. Thanks.

pkg/client/logs_test.go
```diff
@@ -338,22 +324,41 @@ func TestPodsForLogs(t *testing.T) {
...
+				if podCount != tc.expectedPodCount {
+					t.Errorf("Unexpected number of pods returned, expected %d, got %d", tc.expectedPodCount, podCount)
 				}
 			}
-
 		})
 	}
 }
 
+func getFakeKubernetesClientSet() kubernetes.Interface {
```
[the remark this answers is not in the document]
johnSchnake: No; I think that was from the initial PR I grabbed but it has completely changed since then as the issues with the design were found.

pkg/client/logs_test.go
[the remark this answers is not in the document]
johnSchnake: Slightly unsure about that approach now; I tried to double check the `io.Reader` documentation and it says:
```
Implementations of Read are discouraged from returning a zero byte count with a nil error, except when len(p) == 0. Callers should treat a return of 0 and nil as indicating that nothing happened; in particular, it does not indicate EOF.

Implementations must not retain p.
```

Currently, we have 2 problems:
 - if n==0 we currently retain `p` which is a `must not` directive.
 - we return `0, nil` which is discouraged by not completely inappropriate.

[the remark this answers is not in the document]
johnSchnake: By making the method a _tad_ recursive we solve this. If in the `len(data)==0` case we call into the method again, we will either get new data or have EOF eventually.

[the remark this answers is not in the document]
johnSchnake: If you can double-check just this section of logic really; I think it is fine but would appreciate your 👀 and a manual check or two.

pkg/client/logs.go
```diff
@@ -111,30 +140,65 @@ func (r *Reader) Read(p []byte) (int, error) {
 	return len(data), nil
```
zubron: Should we be returning `r.err` here? In the case where there was an error, we treat the error message as `data` which is what we copy to `p` and return the length of here, so we would be providing the contents of the error, but not the actual error itself.

johnSchnake: No, in case there is overflow data we want to not return the error until all the data is done being sent.

So what we're doing now is printing _all_ the data from the bystream now and mixing in the printed errors as they arrive. However we don't return an `error` object until all the data is printed so that someone can still do things like reading until encountering an error.

We don't want to return an error until after all data is printed or else 1 broken set of logs will prevent us from getting all the other logs.

zubron: Yep, but there are separate paths for when the data doesn't fit into the buffer (just above this return and also at the beginning of the function when we prioritise the overflow buffer) so that's already being taken into consideration. The reason this was confusing me is that in [`TestLateErrors`](https://github.com/vmware-tanzu/sonobuoy/pull/1007/files#diff-b3336300806d23af14262c2b62843558R63-R65) we are now making 5 calls to `Read`, so we print out all of the error, but don't get the actual error until the final call when there is no data from either stream. We want to entirely empty the bytestream before any returning any error. I've messed around with a test and I think this is making sense now :) Sorry I'm being so slow with this!

PR #1011, opened 2019-11-22
Default to enableing progress updates from E2E plugin

### LINKED ISSUE #992: Utilize progress updates with E2E plugin after v1.17+
https://github.com/kubernetes/kubernetes/pull/84524 merged meaning that after v1.17 is released we can routinely get progress updates as a run is in flight.

That ticket shows how to accomplish this (we just need to provide a default value for the E2E_EXTRA_ARGS) but we should also accompany this with a blog post to show off the new feature. It has been a big ask for a long time (I'll create another ticket for the blogpost)

I think this is the first time that the plugin definition will be based partially on the cluster version. There may be a bit of thinking required to find the most reasonable place to insert the tiny bit of code needed.

johnSchnake:
As of v1.17 the Kubernetes E2E test image will support progress
updates if the URL is specified.

Fixes #992 

Do the following to test:

```
$ go install && sonobuoy run --kube-conformance-image-version=v1.17.0-beta.2
```

It doesn't matter that your cluster may not be that version, since we are targeting that version of the image it will work. Likewise, you can do `sonobuoy gen` with that same option to see the YAML as well.

After starting the run, just poll against `sonobuoy status --json | jq` to confirm that you're getting progress updates.

Note: This will _not_ work if you are specifying the exact image via digest, e.g. `gcr.io/google-containers/conformance@sha256:8f55d97d42690532e44068b6db98051bc3782c916987cff8db3d43447b09a94e` . I think this is just a limitation that we will live with and is OK. Users that want to do that can always specify then env var themselves.

**Release note**:
```
If targeting v1.17 clusters and above, the e2e plugin will not provide the necessary logic in order to enable progress updates from the plugin. At the current time, you can see progress via `sonobuoy status --json`.
```

### DIFF
--- a/pkg/client/gen.go
+++ b/pkg/client/gen.go
@@ -25,6 +25,7 @@ import (
 	"sort"
 	"strings"
 
+	version "github.com/hashicorp/go-version"
 	"github.com/pkg/errors"
 
 	"github.com/vmware-tanzu/sonobuoy/pkg/config"
@@ -38,8 +39,11 @@ import (
 )
 
 const (
-	e2ePluginName   = "e2e"
-	systemdLogsName = "systemd-logs"
+	e2ePluginName                 = "e2e"
+	systemdLogsName               = "systemd-logs"
+	lastE2EVersionWithoutProgress = "1.16.99"
+
+	envVarKeyExtraArgs = "E2E_EXTRA_ARGS"
 )
 
 // templateValues are used for direct template substitution for manifest generation.
@@ -288,6 +292,10 @@ func systemdLogsManifest(cfg *GenConfig) *manifest.Manifest {
 }
 
 func e2eManifest(cfg *GenConfig) *manifest.Manifest {
+	if cfg.Config == nil {
+		cfg.Config = config.New()
+	}
+
 	m := &manifest.Manifest{
 		SonobuoyConfig: manifest.SonobuoyConfig{
 			PluginName:   "e2e",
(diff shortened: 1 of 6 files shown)


pkg/client/gen.go
```diff
@@ -375,5 +377,30 @@ func e2eManifest(cfg *GenConfig) *manifest.Manifest {
...
+}
+
+func updateExtraArgs(envs []corev1.EnvVar) []corev1.EnvVar {
+	for _, env := range envs {
+		// If set by user, just leave as-is.
+		if env.Name == "E2E_EXTRA_ARGS" {
+			return envs
+		}
+	}
+	envs = append(envs, corev1.EnvVar{Name: "E2E_EXTRA_ARGS", Value: "--progress-report-url=http://localhost:8099/progress"})
```
johnSchnake: Will have to double check where to grab the port from; I believe it is configurable.

pkg/client/gen.go
```diff
@@ -375,5 +377,30 @@ func e2eManifest(cfg *GenConfig) *manifest.Manifest {
...
+	if err != nil {
+		return false
+	}
+	return imageVersion.GreaterThanOrEqual(version.Must(version.NewVersion(firstE2EVersionWithProgress)))
+}
+
+func updateExtraArgs(envs []corev1.EnvVar) []corev1.EnvVar {
+	for _, env := range envs {
+		// If set by user, just leave as-is.
+		if env.Name == "E2E_EXTRA_ARGS" {
```
johnSchnake: Can make this a const.

pkg/client/gen.go
```diff
@@ -38,8 +39,9 @@ import (
 )
 
 const (
-	e2ePluginName   = "e2e"
-	systemdLogsName = "systemd-logs"
+	e2ePluginName               = "e2e"
+	systemdLogsName             = "systemd-logs"
+	firstE2EVersionWithProgress = "1.16.99"
```
johnSchnake: Used this instead of `v1.17.0` since we want prereleases to work as well.

johnSchnake: May need to rename to be more accurate; e.g. `lastE2EVersionWithoutProgress`

johnSchnake:
- The port was configurable in the `config` object which is inside of the `GenConfig` object already.
- fallback to default values if not set for the config (to avoid panics while reading it) and for the port string (to ensure we get a value there for the port)
- added a new test for ensuring the port is customizable

Will merge once green.

pkg/client/gen_internal_test.go
```diff
@@ -0,0 +1,66 @@
+/*
+Copyright 2018 Heptio Inc.
```
johnSchnake: Need to fix this; just a copy/paste error

PR #1019, opened 2019-11-27
Ruby gems CVE-2019-13117 and CVE-2019-16892

SDBrett:
Resolves: CVE-2019-13117 and CVE-2019-16892

Fixes: #1019

Updated gemfile.lock for security vulnerability.

Updated Gemfile to specify gem versions, providing more control over versions when using bundle update. Including the Jekyll version in the Gemfile tells Nelify which version to build with.

Signed-off-by: Brett Johnson <brett@sdbrett.com>


SDBrett:
> Everything seems good to me; seems like an unusually large diff for the intended change but everything still seems to work as expected and I confirmed bundle-audit showed no issues.
> 
> LGTM once @jonasrosland has a chance to peek at it. (also, fwiw even though this is "out of date" with master I should be able to merge w/o issue)

The changes to Gemfile.lock to extend beyond the scope of fixing the vulnerable packages. I felt that it was a good opportunity to update the rest of the gems and prevent dependency issues.

PR #1045, opened 2019-12-12
Fix bug which caused daemonset status to not be updated with results

### LINKED ISSUE #1001: Duplicated plugin name in json status
`sonobuoy status --json | jq`

Should have a list of the plugins and their aggregate data. However, you can see in the output below that it duplicated the plugin and only put the data really into one of them.

```
{
  "plugins": [
    {
      "plugin": "daemonset-worker",
      "node": "ip-172-31-30-196.us-west-2.compute.internal",
      "status": "complete",
      "result-status": "failed",
      "result-counts": {
        "failed": 12,
        "passed": 32,
        "skipped": 4
      }
    },
    {
      "plugin": "daemonset-worker",
      "node": "ip-172-31-47-230.us-west-2.compute.internal",
      "status": "complete",
      "result-status": "",
      "result-counts": null
    }
  ],
```

It also didnt list the `daemonset-master` info (which didnt really run since none of the nodes were schedulable but I would think should be in that list).

FWIW the plugin `daemonset-worker` was targeting nodes without the master label on them; `daemonset-master` the opposite.

Sonobuoy v0.16.4

johnSchnake:
The bug would cause only one of the plugin/node values in the
`sonobuoy status --json` output to be updated and it would include
all the results for all nodes.

Instead, we want each plugin/node value in that list to report its
own values for passed/failed.

Fixes #1001

**Release note**:
```
Fixed bug which caused daemonset plugins' status to not be updated with results properly.
```

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -222,9 +222,7 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 		}
 
 		// Update the plugin status with this post-processed information.
-		statusInfo := map[string]int{}
-		statusCounts(&item, statusInfo)
-		updatePluginStatus(kubeClient, cfg.Namespace, p.GetName(), item.Status, statusInfo)
+		updatePluginStatus(kubeClient, cfg.Namespace, p.GetName(), item)
 	}
 
 	// Saving plugin definitions in their respective folders for easy reference.
@@ -357,20 +355,38 @@ func updateStatus(client kubernetes.Interface, namespace string, status string, 
 	return setPodStatusAnnotation(client, namespace, podStatus)
 }
 
-func updatePluginStatus(client kubernetes.Interface, namespace string, pluginType string, pluginResultStatus string, pluginResultCounts map[string]int) error {
+func updatePluginStatus(client kubernetes.Interface, namespace string, pluginName string, item results.Item) error {
 	podStatus, err := pluginaggregation.GetStatus(client, namespace)
 	if err != nil {
 		return errors.Wrap(err, "failed to get the existing status")
 	}
 
-	for i := range podStatus.Plugins {
-		if podStatus.Plugins[i].Plugin == pluginType {
-			podStatus.Plugins[i].ResultStatus = pluginResultStatus
-			podStatus.Plugins[i].ResultStatusCounts = pluginResultCounts
-			break
-		}
-	}
+	integrateResultsIntoStatus(podStatus, pluginName, &item)
+
 	return setPodStatusAnnotation(client, namespace, podStatus)
+}
+
+func integrateResultsIntoStatus(podStatus *pluginaggregation.Status, pluginName string, item *results.Item) {
+	for i := range podStatus.Plugins {
+		if podStatus.Plugins[i].Plugin != pluginName {
+			continue
+		}
+		var itemForNode *results.Item
+		if podStatus.Plugins[i].Node == plugin.GlobalResult {
+			itemForNode = item
+		} else {
+			itemForNode = item.GetSubTreeByName(podStatus.Plugins[i].Node)
+		}
+
+		if itemForNode == nil {
+			return
+		}
+
+		statusInfo := map[string]int{}
+		statusCounts(itemForNode, statusInfo)
+		podStatus.Plugins[i].ResultStatus = itemForNode.Status
+		podStatus.Plugins[i].ResultStatusCounts = statusInfo
+	}
 }
 
 // setPodStatusAnnotation sets the status on the pod via an annotation. It will overwrite the
--- a/cmd/sonobuoy/app/results.go
+++ b/cmd/sonobuoy/app/results.go
@@ -167,7 +167,7 @@ func printSinglePlugin(input resultsInput, r *results.Reader) error {
 		return err
 	}
 
-	obj = getItemInTree(obj, input.node)
+	obj = obj.GetSubTreeByName(input.node)
 	if obj == nil {
 		return fmt.Errorf("node named %q not found", input.node)
 	}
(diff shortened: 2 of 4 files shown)


pkg/discovery/discovery.go
```diff
@@ -357,20 +355,38 @@ func updateStatus(client kubernetes.Interface, namespace string, status string,
...
+
+	return setPodStatusAnnotation(client, namespace, podStatus)
+}
+
+func integrateResultsIntoStatus(podStatus *pluginaggregation.Status, pluginName string, item *results.Item) {
 	for i := range podStatus.Plugins {
-		if podStatus.Plugins[i].Plugin == pluginType {
-			podStatus.Plugins[i].ResultStatus = pluginResultStatus
-			podStatus.Plugins[i].ResultStatusCounts = pluginResultCounts
-			break
```
johnSchnake: This was the main idea that was broken before. This was looping over the plugins and updating one value (the first that matched by name) and breaking the loop.

Instead you need to try and update every plugin in the list that matches on name. It is just a slight misnomer since `podStatus.Plugins` is really meaning `podStatus.PluginNodePairs`

This meant I had to add the code (which existed elsewhere) to just snip the `item` at the right spot to get the results for just a particular node.

pkg/discovery/discovery.go
```diff
@@ -397,3 +413,24 @@ func setPodStatusAnnotation(client kubernetes.Interface, namespace string, statu
 	_, err = client.CoreV1().Pods(namespace).Patch(podName, types.MergePatchType, patchBytes)
 	return err
 }
+
+func getItemInTree(i *results.Item, root string) *results.Item {
```
johnSchnake: This was copied from elsewhere; I considered putting it into `results` and exporting it; maybe even as method on the `Item` type. Let me know if you'd prefer that.

pkg/discovery/discovery.go
```diff
@@ -357,20 +355,38 @@ func updateStatus(client kubernetes.Interface, namespace string, status string,
...
-			podStatus.Plugins[i].ResultStatusCounts = pluginResultCounts
-			break
+		if podStatus.Plugins[i].Plugin != pluginName {
+			continue
 		}
+		var itemForNode *results.Item
+		if podStatus.Plugins[i].Node == plugin.GlobalResult {
+			itemForNode = item
+		} else {
+			itemForNode = getItemInTree(item, podStatus.Plugins[i].Node)
```
johnSchnake: Here I just traverse and choose the item/branch that matches the node name. It works and I think is sufficient, but I also considered adding some sort of metadata field so that we know that not only does `name=NodeName` but it has some `type=hostname` or something like that. Maybe not really necessary though.

pkg/discovery/discovery.go
```diff
@@ -357,20 +355,38 @@ func updateStatus(client kubernetes.Interface, namespace string, status string,
...
 		return errors.Wrap(err, "failed to get the existing status")
 	}
 
+	integrateResultsIntoStatus(podStatus, pluginType, &item)
+
+	return setPodStatusAnnotation(client, namespace, podStatus)
+}
+
+func integrateResultsIntoStatus(podStatus *pluginaggregation.Status, pluginName string, item *results.Item) {
 	for i := range podStatus.Plugins {
```
zubron: Unrelated to this change, but the second value from `range` could be used here rather than needing to index into the slice with `i`.

pkg/discovery/discovery.go
zubron: I think it could help for readability, as I find myself getting lost in what some of these objects are and what they represent. I think just a comment would be enough though to inform readers of the code for next time that the `Name` field for a DS result item will be the node name.

pkg/discovery/discovery.go
zubron: :+1: I think having just one copy would be good.

pkg/discovery/discovery.go
johnSchnake: Only for the value lookups; the range loop makes a local copy and so sometimes you think a value is editable but it isn't necessary depending on if its a pointer/value/whatever. Most times if I'm editing the value I just err on the side of caution and use the index.

Testing this out, it is one of the cases where the value doesnt get edited when you try and set the status/counts a few lines below.

pkg/discovery/discovery.go
johnSchnake: Created a new issue; adding it to just nodes doesn't make as much sense as adding it to a few different things (overall summary, file summaries, etc) and so the change is a little more than I want to smush into this PR.

johnSchnake:
Waiting to go green; expect it to do so though.

As of now this is the output for the various status views:
```
$ sonobuoy status --json | jq
{
  "plugins": [
    {
      "plugin": "e2e",
      "node": "global",
      "status": "complete",
      "result-status": "passed",
      "result-counts": {
        "passed": 1,
        "skipped": 4896
      }
    },
    {
      "plugin": "systemd-logs",
      "node": "kind-control-plane",
      "status": "complete",
      "result-status": "passed",
      "result-counts": {
        "passed": 1
      }
    },
    {
      "plugin": "systemd-logs",
      "node": "kind-worker",
      "status": "complete",
      "result-status": "passed",
      "result-counts": {
        "passed": 1
      }
    }
  ],
  "status": "complete",
  "tar-info": {
    "name": "201912122232_sonobuoy_93015c64-5e06-446a-81b0-9262b3c78ff0.tar.gz",
    "created": "2019-12-12T22:32:40.2666856Z",
    "sha256": "d9cc3080c6e85da6f76e70028d52e7541d21c330c37594e8230b57cd7bd2d4bd",
    "size": 233587
  }
}
```

```
$ sonobuoy status
         PLUGIN     STATUS   RESULT   COUNT
            e2e   complete   passed       1
   systemd-logs   complete   passed       2

Sonobuoy has completed. Use `sonobuoy retrieve` to get results.
```
```
$ sonobuoy status --show-all
         PLUGIN                 NODE     STATUS   RESULT
            e2e               global   complete   passed
   systemd-logs   kind-control-plane   complete   passed
   systemd-logs          kind-worker   complete   passed
```

pkg/discovery/discovery.go
zubron: Ah, I didn't know though that. TIL!

PR #1048, opened 2019-12-12
Tag Items with metadata about the type if item (e.g. node/file/etc)

### LINKED ISSUE #1047: Add some metadata to result.Item to describe the type of entry
Improve readability if looking at the sonobuoy_results.yaml file. 

The `Name` field can imply different things; sometimes it is the name of the plugin, the name of a node, the name of a file, or the name of a testcase.

I dont think we should add it on every test case, but for summaries, nodes, and files, it seems reasonable to add context if someone looks at the raw data.

_ref @zubron in https://github.com/vmware-tanzu/sonobuoy/pull/1045_

johnSchnake:
If you look at the sonobuoy_results.yaml it can be a bit confusing
understanding the tree structure and what each node in that tree means.

This adds some extra metadata to certain Items (plugin summary, nodes,
and files) in order to make the raw file more comprehensible and
potentially aid in other processing.

Fixes #1047

**Release note**:
```
Adds additional metadata into the sonobuoy_results.yaml file to help indicate what each node in the tree represents.
```

### DIFF
--- a/pkg/client/results/junit.go
+++ b/pkg/client/results/junit.go
@@ -174,9 +174,12 @@ func junitProcessFile(pluginDir, currentFile string) (Item, error) {
 	}
 
 	resultObj := Item{
-		Name:     filepath.Base(currentFile),
-		Status:   StatusUnknown,
-		Metadata: map[string]string{"file": relPath},
+		Name:   filepath.Base(currentFile),
+		Status: StatusUnknown,
+		Metadata: map[string]string{
+			metadataFileKey: relPath,
+			metadataTypeKey: metadataTypeFile,
+		},
 	}
 
 	infile, err := os.Open(currentFile)
--- a/pkg/client/results/processing.go
+++ b/pkg/client/results/processing.go
@@ -57,6 +57,18 @@ const (
 	// PostProcessedResultsFile is the name of the file we create when doing
 	// postprocessing on the plugin results.
 	PostProcessedResultsFile = "sonobuoy_results.yaml"
+
+	// metadataFileKey is the key used in an Item's metadata field when the Item is
+	// representing the a file summary (and its leaf nodes are individual tests or suites).
+	metadataFileKey = "file"
+
+	// metadataTypeKey is the key used in an Item's metadata field when describing what type
+	// of entry in the tree it is. Currently we just tag summaries, files, and nodes.
+	metadataTypeKey = "type"
+
+	metadataTypeNode    = "node"
+	metadataTypeFile    = "file"
+	metadataTypeSummary = "summary"
 )
 
 // ResultFormat constants are the supported values for the resultFormat field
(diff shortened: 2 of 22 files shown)


pkg/client/results/processing.go
```diff
@@ -57,6 +57,18 @@ const (
...
+
+	// metadataFileKey is the key used in an Item's metadata field when the Item is
+	// representing the a file summary (and its leaf nodes are individual tests or suites).
+	metadataFileKey = "file"
+
+	// metadataTypeKey is the key used in an Item's metadata field when describing what type
+	// of entry in the tree it is. Currently we just tag summaries, files, and nodes.
+	metadataTypeKey = "type"
+
+	metadataTypeNode    = "node"
```
johnSchnake: Haven't exported these for now; may eventually want to but keep them unexported if possible.

pkg/client/results/processing.go
```diff
@@ -57,6 +57,18 @@ const (
...
+	metadataTypeFile    = "file"
+	metadataTypeSummary = "plugin-summary"
```
johnSchnake: Is this the name we want here? The other values (node/file) seem right but I'm not sure about this one.

pkg/client/results/processing.go
```diff
@@ -57,6 +57,18 @@ const (
...
 	// postprocessing on the plugin results.
 	PostProcessedResultsFile = "sonobuoy_results.yaml"
+
+	// metadataFileKey is the key used in an Item's metadata field when the Item is
+	// representing the a file summary (and its leaf nodes are individual tests or suites).
+	metadataFileKey = "file"
+
+	// metadataTypeKey is the key used in an Item's metadata field when describing what type
+	// of entry in the tree it is. Currently we just tag summaries, files, and nodes.
+	metadataTypeKey = "type"
```
johnSchnake: Is `type` the best word here? Open to other ideas, happy with type if you are though.

zubron: The only thing I can think of is “source”, to indicate where a result came from, but I’m not particularly keen on that. I’m happy to stick with type.

pkg/client/results/junit.go
```diff
@@ -174,9 +174,12 @@ func junitProcessFile(pluginDir, currentFile string) (Item, error) {
 	}
 
 	resultObj := Item{
-		Name:     filepath.Base(currentFile),
-		Status:   StatusUnknown,
-		Metadata: map[string]string{"file": relPath},
+		Name:   filepath.Base(currentFile),
+		Status: StatusUnknown,
+		Metadata: map[string]string{
+			metadataFileKey: relPath,
```
zubron: Is it necessary to have the file key/entry here given that it’s already in the name? Or is it to just to make it really clear?

pkg/client/results/processing.go
zubron: I think just “summary” alone might be enough?

pkg/client/results/junit.go
johnSchnake: Yes, they are unique since the `Name` is just the basename and the actual `Metadata: {"file":<val>}` will be the full filename relative to where the sonobuoy_results.yaml file is located at.

For instance:
```
  - name: systemd_logs
    status: passed
    meta:
      file: results/kind-control-plane/systemd_logs
```

PR #1051, opened 2019-12-16
Adds commands to print default plugin definitions

### LINKED ISSUE #991: Facilitate printing the default plugin definitions
I routinely want to print the plugin definition for the e2e or systemd-logs plugin so that I can tweak a value or two. It is not difficult to do `sonobuoy gen` and copy/paste but it is a a slight frustration and requires worrying abouut the extra level of indentation we get.

It would be trivial to add support to print these out via something like this:
```
sonobuoy gen plugin e2e
```

We already have `sonobuoy gen config` to help print configs, `gen plugin` to come up with custom plugins, so it seems natural to support the built-in values too.

johnSchnake:
We have the `sonobuoy gen` command which prints a lot of
different types of resources and `sonobuoy gen plugin` which
prints custom plugin definitions. Sometimes it would be helpful
to be able to print the default e2e or sytemdlogs plugin definitions
so that they can be tweaked.

Added two commands `sonobuoy gen plugins e2e` and
`sonobuoy gen plugins systemd-logs` which does this.

Minor tweak to code organization to improve code reuse.

Fixes #991

These were just printing some data by hitting a code path that was already being used/tested. I didn't add any new tests for this code. Let me know if you want to add some basic one.

**Release note**:
```
Adds two subcommands (e2e and systemd-logs) to `sonobuoy gen plugin` which will print those plugin definitions respectively. This will hopefully make certain workflows easier by giving users a starting point for their plugin definition.
```

### DIFF
--- a/cmd/sonobuoy/app/gen.go
+++ b/cmd/sonobuoy/app/gen.go
@@ -67,7 +67,10 @@ type genFlags struct {
 	genflags *pflag.FlagSet
 }
 
+// TODO(jschnake): Avoid using these globals if possible.
 var genflags genFlags
+var genSystemdLogsflags genFlags
+var genE2Eflags genFlags
 
 func GenFlagSet(cfg *genFlags, rbac RBACMode) *pflag.FlagSet {
 	genset := pflag.NewFlagSet("generate", pflag.ExitOnError)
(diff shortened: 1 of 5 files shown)


johnSchnake:
Thanks to @zubron I found the tests were failing because of some issue with the sonobuoy gen logic; still to be fixed but `cd tests/integration && sonobuoy gen -p testImage/yaml/job-junit-passing-singlefile.yaml` will output the default plugins _and_ the specified plugin. This means that it is trying to run a full conformance run which causes the tests to time out.

PR #1053, opened 2019-12-17
Clarify that users shouldn't provide more options when running a file

### LINKED ISSUE #1027: Flags for `run` do not override any values when a manifest is passed with `-f`
There is an issue where custom values specified via command line flags do not override content that is in a generated manifest (the file you are passing with `-f`)

_Originally posted by @zubron in https://github.com/vmware-tanzu/sonobuoy/issues/984#issuecomment-561608412_

This may be the intended behaviour but it is confusing for users. If they specify both, we should either prevent them from doing that, or give a warning that the command line flags will have no effect.

johnSchnake:
If using the --file option for `sonobuoy run` they shouldn't also
try and set other options.

This is because we are not able to take the file output and turn
that into a genConfig object which can be merged with the other
command line options.

When in the expanded file format, we are unsure if which parts
should be merged/how.

These are meant to be exclusive options.

Fixes #1027

**Release note**:
```
Added an error message to clarify that users should not specify other customization flags when trying to use `sonobuoy run --file`. You are able to provide the kubeconfig and context flags as those are still used to connect to the cluster though.
```

### DIFF
--- a/cmd/sonobuoy/app/run.go
+++ b/cmd/sonobuoy/app/run.go
@@ -17,6 +17,7 @@ limitations under the License.
 package app
 
 import (
+	"fmt"
 	"os"
 	"time"
 
@@ -36,7 +37,11 @@ type runFlags struct {
 	genFile       string
 }
 
-var runflags runFlags
+var (
+	runflags runFlags
+
+	allowedGenFlagsWithRunFile = []string{kubeconfig, context}
+)
 
 func RunFlagSet(cfg *runFlags) *pflag.FlagSet {
 	runset := pflag.NewFlagSet("run", pflag.ExitOnError)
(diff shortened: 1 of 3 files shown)


cmd/sonobuoy/app/run.go
```diff
@@ -36,7 +37,11 @@ type runFlags struct {
 	genFile       string
 }
 
-var runflags runFlags
+var (
+	runflags runFlags
+
+	allowedGenFlagsWithRunFile = []string{"kubeconfig", "context"}
```
johnSchnake: This was the big ? for me. Is this the full list of things that users should be able to specify while also using `--file`?

I couldn't find others that weren't meant to customize the run.

johnSchnake: Also, its unfortunate but we do just have to hardcode the names unless we overengineer an alternative solution.

We do not typically change the flag names so I dont expect it will be a problem though.

zubron: I think these are the only flags that need to be excluded. Yeah, I guess you could have the flag names as constants so that they can be included here and reference the same strings. The other alternative I could think of was extracting these config values so that they are added explicitly to the `run` command rather than being included through the `GenFlagSet` so that you then error on _any_ changed flag within that set. But it's probably an overengineered solution :)

johnSchnake: I went with just making a constant; it gets a bit tricky to pull out the values from the genFlagSet since it also stores those flags within the data structure; I just didn't want to have anything cascade. Not against a change in the future if this annoys us though.

PR #1060, opened 2019-12-20
Adds `namespace` labels to Cluster resources when creating and deleting

danielfbm:

When running in parallel `delete` will remove all cluster resources. This PR will add a label using the same namespace and make sure that one `run` will not affect another

None reported

Will squash commits if PR is accepted.

**Release note**:
```
Adds namespace labels to Cluster resource when creating and deleting data allowing parallel runs
```


johnSchnake:
Thanks for the contribution; I'm certain this will help a bunch of folks. I know that it is common to run multiple copies in different namespaces. 👍

Issue #1058, opened 2019-12-19
Is the "images download" command needed?


zubron:
There is a `download` subcommand under `images`. The purpose of this command is to download the images for the E2E plugin into a tarball. However, this won't work for our recent releases as a number of the images included in the E2E images list are unreachable. They either don't exist or require authentication. These images have been kept in the list to maintain consistency with upstream however Sonobuoy can't do anything with them. They also cause the `download` command to always fail as we can't exclude them.

Given that we have had no feedback from users regarding this, should we remove the `download` command? Or should we update our images list to exclude images that can't be accessed by Sonobuoy?

johnSchnake:
@stevesloka Do you remember the details on the download command from when you implemented it? I assume it was for a flow of downloading them, transferring them physically to another network via a thumb drive or something, then uploading them with docker upload?

I don't recall what our thinking was for the unreachable/auth-required images. I don't really utilize those commands personally and had thought the behavior was that each image would pass/fail on its own and that the whole command wouldn't be aborted.

I think that is a reasonable way to approach it; just log everything appropriately and even change the exit code perhaps if we hit errors. In the end the user can still get the resulting tarball with all the images we could collect.

Thoughts?

stevesloka:
If I remember correctly, the download command used the local user's docker client to download images.  

1. Download images to your local docker client
2. Define a new registry to upload and push them to a private registry

The model around using the local docker client was that it would have the proper credentials to pull/push to any registry 

I'm not sure what images would require auth to pull, they should all be the upstream images and public but agree a log level would be a good start to see what's going on. 

Let me know if I can help further. =)

vladimirvivien:
@wilsonehusin take a 👀

wilsonehusin:
> Given that we have had no feedback from users regarding this, should we remove the download command?

The Marie Kondo in me agrees with this, since I suspect the more popular workflow is to run `pull` then `push`.

Since images are tagged to the repository, I suspect the downloaded images will need manual re-tagging to be pushed to another repository. One way I can think where `download` command can be useful is for the workflow as such:
- `sonobuoy images download`
- `kind load image-archive downloaded-images.tar.gz`
- `sonobuoy run`

I don't know if any other cluster provisioner provide the same `load image-archive` feature as KinD though.

On the other hand, `sonobuoy images push` handles this re-tagging workflow nicely.

@vladimirvivien thoughts?

PR #1061, opened 2019-12-20
Add support for pulling and pushing all images

### LINKED ISSUE #1028: Include all necessary images in images pull/push commands
If you want to use a custom registry, you can do it but it involves a lot of manual steps and the steps differ for the e2e and systmed-logs plugins. It's proving to be an error prone and confusing workflow.

I would like to modify the `images` command to take not only the e2e test images into account, but also the sonobuoy, conformance, and systemd-logs images. It would be ideal if users just needed to run one set of commands to pull and then push the images.

The commands would also have to accept the `sonobuoy-image` and `conformance-image` flags to allow custom images/versions to be used. If omitted we would use the defaults.

### LINKED ISSUE #1059: Make e2eRegistryConfig flag not required
As part of #1028 and adapting the `images` command to work with other plugins, we need to make sure that we remove `--e2e-repo-config` as a required flag for `images push` as this is only applicable to the `e2e` plugin.

> Lets make a follow up issue for this

_Originally posted by @johnSchnake in https://github.com/vmware-tanzu/sonobuoy/diffs_

zubron:
This changes adds support for pulling the Sonobuoy worker image, the
systemd-logs image and the Kubernetes conformance image in addition to
the images required for the e2e plugin.

The individual images to pull cannot be configured. Only the default
version for the cluster will be pulled. The images can only be pushed to
one registry which is specified with the new flags `--custom-registry`.

This currently only works for the default built in plugins. A follow up
change will be made to apply the same approach to any image in a plugin
definition.

- Fixes #1028 
- Fixes #1059 

I've tested the new commands for this manually. I ran the following locally:
```
sonobuoy images pull -p systemd-logs
sonobuoy images push -p systemd-logs --custom-registry zubron
```
And you can see that the images were pushed to my public registry: https://hub.docker.com/u/zubron

**Release note**:
```
Sonobuoy can now pull all images for Sonobuoy and the built-in plugins and push them to a custom registry. The E2E test images are still configured as before, however all other images will be pushed to a custom registry specified using the flag `--custom-registry`. The image operations can also be previewed by using the `--dry-run` flag.
```

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -40,6 +40,7 @@ const (
 	pluginFlag          = "plugin"
 	timeoutFlag         = "timeout"
 	waitOutputFlag      = "wait-output"
+	customRegistryFlag  = "custom-registry"
 	kubeconfig          = "kubeconfig"
 	context             = "context"
 )
@@ -124,17 +125,19 @@ func AddKubeconfigFlag(cfg *Kubeconfig, flags *pflag.FlagSet) {
 	flags.StringVar(&cfg.Context, "context", "", "Context in the kubeconfig to use.")
 }
 
-// AddPluginFlag describes which plugin's images to interact with
-func AddPluginFlag(cfg *string, flags *pflag.FlagSet) {
-	// The default is 'e2e' since it's the only plugin enabled at the moment
-	flags.StringVarP(cfg, pluginFlag, "p", "e2e", "Describe which plugin's images to interact (Valid plugins are 'e2e').")
-}
-
 // AddE2ERegistryConfigFlag adds a e2eRegistryConfigFlag flag to the provided command.
 func AddE2ERegistryConfigFlag(cfg *string, flags *pflag.FlagSet) {
 	flags.StringVar(
 		cfg, e2eRegistryConfigFlag, "",
-		"Specify a yaml file acting as KUBE_TEST_REPO_LIST, overriding registries for test images.",
+		"Specify a yaml file acting as KUBE_TEST_REPO_LIST, overriding registries for test images. Required when pushing images for the e2e plugin.",
+	)
+}
+
+// AddCustomRepoFlag adds a custom registry flag to the provided command.
+func AddCustomRegistryFlag(cfg *string, flags *pflag.FlagSet) {
+	flags.StringVar(
+		cfg, customRegistryFlag, "",
+		"Specify a registry to override the Sonobuoy and Plugin image registries.",
 	)
 }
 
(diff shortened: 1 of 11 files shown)


cmd/sonobuoy/app/images.go
```diff
@@ -165,124 +176,180 @@ func getClusterVersion(kubeconfig Kubeconfig) (string, error) {
...
-		}
-
-		images, err := image.GetE2EImages(e2eRegistryConfig, version)
-		if err != nil {
-			return errors.Wrap(err, "couldn't get images")
-		}
+func downloadImages(plugins []string, kubeconfig Kubeconfig, e2eRegistryConfig string) error {
+	for _, plugin := range plugins {
+		switch plugin {
+		case e2ePlugin:
```
zubron: I haven't updated this function as I don't know if we should maintain it: #1058

cmd/sonobuoy/app/images.go
```diff
@@ -165,124 +176,180 @@ func getClusterVersion(kubeconfig Kubeconfig) (string, error) {
...
 
-		imageClient := image.NewImageClient()
+	imageClient := image.NewImageClient()
+	return imageClient.DeleteImages(images, numDockerRetries)
+}
 
-		return imageClient.DeleteImages(images, numDockerRetries)
+func substituteRegistry(image string, customRegistry string) string {
+	trimmedRegistry := strings.TrimRight(customRegistry, "/")
+	components := strings.SplitAfter(image, "/")
```
zubron: I thought I left a comment on this but it's gotten lost :confused: 

This is a best attempt at replacing the registry in a full image name. It's possible for the "name" component of an image name to also contain a `/` and we see that in [some of the E2E test images](https://github.com/vmware-tanzu/sonobuoy/blob/master/pkg/image/v1.17.go#L71-L74). In cases like these, this function would substitute everything up to the last `/` which could include the registry *and* part of the image name. I don't think there's a way to identify what is the registry URL component of a FQIN. The images will still tag and push correctly, but it might be confusing for users when replacing images in their plugin definitions.

cmd/sonobuoy/app/args.go
```diff
@@ -348,6 +351,11 @@ func AddPluginEnvFlag(p *PluginEnvVars, flags *pflag.FlagSet) {
 	flags.Var(p, "plugin-env", "Set env vars on plugins. Values can be given multiple times and are in the form plugin.env=value")
 }
 
+// AddPluginListFlag adds the flag to keep track of which built-in plugins to use.
+func AddPluginListFlag(p *[]string, flags *pflag.FlagSet) {
+	flags.StringSliceVarP(p, "plugin", "p", []string{"e2e"}, "Describe which plugin's images to interact (Valid plugins are 'e2e', 'systemd-logs').")
```
zubron: Should this default to both plugins now? Seems like that would be simpler UX here.

cmd/sonobuoy/app/images.go
johnSchnake: Its a short function but I think this is a good function to have in a function so that we can visually ensure it handles certain cases.

cmd/sonobuoy/app/images.go
```diff
@@ -165,124 +176,180 @@ func getClusterVersion(kubeconfig Kubeconfig) (string, error) {
...
 }
 
-func listImages(plugin string, kubeconfig Kubeconfig) error {
-	switch plugin {
-	case "e2e":
-		version, err := getClusterVersion(kubeconfig)
-		if err != nil {
-			return errors.Wrap(err, "failed to get cluster version")
-		}
+func listImages(plugins []string, kubeconfig Kubeconfig) error {
```
johnSchnake: This is another good candidate for a test function; we can add it when we expand to other plugins though. I know the systemd-logs test wouldnt be very significant and the e2e one is a bit awkward to test.

cmd/sonobuoy/app/args.go
johnSchnake: Yes it looks like this is just used for the images command. If we are going to, by default, try to run both of those plugins and the `images` commands can support both of them, I agree it should.

johnSchnake:
My only other idea for this is if it would be useful to have a `dry-run` option when pushing so that you can confirm the repos/tags are as expected.

When trying to manually confirm this all works, I can't actually check much manually without _actually_ pushing some images.

Otherwise, my only comments were (already commented but adding for clarity):
 - ideally add another test to ensure we are splitting registries/tags appropriately
 - default to doing both systemd-logs and e2e

zubron:
Thanks for the feedback! I've addressed your two comments. I also added a `dry-run` option which just performs the logging operations that the existing docker image client does. I've left it as a separate commit for now. Let me know what you think!

PR #1065, opened 2020-01-07
Update goreleaser to remove deprecated field

johnSchnake:
https://github.com/goreleaser/goreleaser/commit/8defb77b0eea3d775b016b32f8d5f91890e4e1a7#diff-6bc8131dff9242f298c3811fada0c174L180-L196 removed the archive field and it needs updated for the release.


johnSchnake:
We should consider adding a step in our build which, even when not on master, will try and use goreleaser with `--skip-publish` so we can check things like this more frequently.

johnSchnake:
Locally I can run `goreleaser release --skip-publish --skip-validate` locally where I was otherwise able to repro this issue. Going to go ahead and merge this.

PR #1068, opened 2020-01-09
pkg/client/logs.go: retry watching pods to stream logs from

### LINKED ISSUE #1064: 100% CPU usage after running "sonobuoy logs -f"
After running even single test, e.g.: 
```
sonobuoy run --e2e-focus 'Aggregator Should be able to support the 1.10 Sample API Server using the current Aggregator'
```

And then leaving `sonobuoy logs -f` running for (in my case 2h), the process starts consuming 100% CPU of a single core.

`sonobuoy` to not trigger fans in my laptop :)

`strace` doesn't say much, didn't try `gdb` or anything:
```
$ sudo strace -p 2554549
strace: Process 2554549 attached
futex(0x555c4acd9468, FUTEX_WAIT_PRIVATE, 0, NULL) = 0
futex(0x555c4acd9468, FUTEX_WAIT_PRIVATE, 0, NULL) = 0
futex(0x555c4acd9468, FUTEX_WAIT_PRIVATE, 0, NULL
```

If you have some suggestions for debugging, please let me know.

**Environment:**

- Sonobuoy version: 
```
$ sonobuoy version --kubeconfig local-testing/kubeconfig 
Sonobuoy Version: v0.17.0
MinimumKubeVersion: 1.15.0
MaximumKubeVersion: 1.17.99
GitSHA: 
API Version:  v1.17.0
```
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:20:10Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:12:17Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```
- Kubernetes installer & version: https://github.com/flexkube/libflexkube
- Cloud provider or hardware configuration: VirtualBox with Vagrant
Kubernetes nodes:
```
core@controller01 ~ $ cat /etc/os-release 
NAME="Flatcar Container Linux by Kinvolk"
ID=flatcar
ID_LIKE=coreos
VERSION=2345.99.1
VERSION_ID=2345.99.1
BUILD_ID=2019-12-19-0008
PRETTY_NAME="Flatcar Container Linux by Kinvolk 2345.99.1 (Rhyolite)"
ANSI_COLOR="38;5;75"
HOME_URL="https://flatcar-linux.org/"
BUG_REPORT_URL="https://issues.flatcar-linux.org"
FLATCAR_BOARD="amd64-usr"
```
Host (where `sonobuoy` runs):
```
$ cat /etc/os-release 
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="0;36"
HOME_URL="https://www.archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux
```

invidian:

TODO

- Fixes #1064

Draft PR, as I want this to go out for initial review, as I have little idea how to fix that "properly". Will improve the description etc tomorrow. Feel free to push additional fixes if needed.

**Release note**:
```
Prevent CPU drain when streaming logs from pods, when connection to API server breaks
```

### DIFF
--- a/pkg/client/logs.go
+++ b/pkg/client/logs.go
@@ -171,23 +171,40 @@ func getPodsToStreamLogs(client kubernetes.Interface, cfg *LogConfig, podCh chan
 //  If a plugin name has been provided, retrieve the pods with only the plugin label matching that plugin name. If no pods are found,
 // or no plugin has been specified, retrieve all pods within the namespace. It will return an error if unable to create the watcher
 // but will continue to add pods to the channel in a separate go routine.
-func watchPodsToStreamLogs(client kubernetes.Interface, cfg *LogConfig, podCh chan *v1.Pod) error {
+func watchPodsToStreamLogs(client kubernetes.Interface, cfg *LogConfig, podCh chan *v1.Pod, errc chan error) {
 	listOptions := metav1.ListOptions{}
 	if cfg.Plugin != "" {
 		selector := metav1.AddLabelToSelector(&metav1.LabelSelector{}, "sonobuoy-plugin", cfg.Plugin)
 		listOptions = metav1.ListOptions{LabelSelector: metav1.FormatLabelSelector(selector)}
 	}
 
-	watcher, err := client.CoreV1().Pods(cfg.Namespace).Watch(listOptions)
-	if err != nil {
-		return errors.Wrap(err, "failed to watch pods")
+	var ch <-chan watch.Event
+
+	f := func() (bool, error) {
+		watcher, err := client.CoreV1().Pods(cfg.Namespace).Watch(listOptions)
+		if err != nil {
+			errc <- errors.Wrap(err, "failed to watch pods")
+
+			return false, nil
+		}
+
+		ch = watcher.ResultChan()
+
+		return true, nil
 	}
-	ch := watcher.ResultChan()
+
+	waitWithBackoff(f)
 
 	go func() {
 		for {
 			select {
-			case v := <-ch:
+			case v, ok := <-ch:
+				if !ok {
+					waitWithBackoff(f)
+
+					continue
+				}
+
 				if v.Type == watch.Added && v.Object != nil {
 					switch t := v.Object.(type) {
 					case *v1.Pod:
@@ -198,7 +215,6 @@ func watchPodsToStreamLogs(client kubernetes.Interface, cfg *LogConfig, podCh ch
 			}
 		}
 	}()
-	return nil
 }
 
 // LogReader configures a Reader that provides an io.Reader interface to a merged stream of logs from various containers.
@@ -218,19 +234,18 @@ func (s *SonobuoyClient) LogReader(cfg *LogConfig) (*Reader, error) {
 	podCh := make(chan *v1.Pod)
 	var wg sync.WaitGroup
 
+	errc := make(chan error)
+
 	if cfg.Follow {
 		// Extra waitGroup item ensures we just keep processing forever.
 		wg.Add(1)
-		if err := watchPodsToStreamLogs(client, cfg, podCh); err != nil {
-			return nil, errors.Wrap(err, "failed to watch for new pods")
-		}
+		watchPodsToStreamLogs(client, cfg, podCh, errc)
 	} else {
 		if err := getPodsToStreamLogs(client, cfg, podCh); err != nil {
 			return nil, errors.Wrap(err, "failed to list pods")
 		}
 	}
 
-	errc := make(chan error)
 	agg := make(chan *message)
 
 	drainPodChannelAndStartStreaming := func() {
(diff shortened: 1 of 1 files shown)


invidian:
Now I think, list of pods should be wiped on reconnect, otherwise logs will be showing twice.

zubron:
Thanks for taking the time to submit this PR! I'm sorry it's taken me so long to get back to you.

I think this is along the right track. I came across something in client-go which seems to wrap up the approach used here: [`RetryWatcher`](https://github.com/kubernetes/client-go/blob/master/tools/watch/retrywatcher.go#L44-L57). This takes care of re-establishing the `Watch` in the case where it is closed. Out of curiosity I tried it out on a branch here: https://github.com/vmware-tanzu/sonobuoy/compare/master...zubron:watch-retry. In that branch I also set a timeout on the `ListOptions` to trigger the behaviour original behaviour. I think we'd still need some handling for the watcher being `Done` or the channel being closed, but in those cases, I think we would just return from the function. In my tests, the watcher gets re-established, and in cases where the client can no longer connect, an error message indicating that is printed.

The only thing is that I'm not sure what the `initialResourceVersion` should be when creating the RetryWatcher :confused: I think you can call `List` to get the `PodList` and use that list's `ResourceVersion` but I'm not sure.

Also, we still have the problem you mentioned where the logs would be duplicated upon reconnection. Do you think it would be enough to keep track of the pods we have seen and only send the pods which we haven't seen before on `podCh`?

invidian:
I was aware of `RetryWatcher`, but I haven't figured out how to pass `ListOptions` to it, I didn't spent much time working on that, so nice work!

>The only thing is that I'm not sure what the initialResourceVersion should be when creating the RetryWatcher confused I think you can call List to get the PodList and use that list's ResourceVersion but I'm not sure.

As we explicitly ignore this parameter in closure, I think it doesn't matter what we pass, so long it passes this check: https://github.com/kubernetes/client-go/blob/master/tools/watch/retrywatcher.go#L67-L74.

It is then being passed here: https://github.com/kubernetes/client-go/blob/master/tools/watch/retrywatcher.go#L103-L105, but we ignore it in the branch you posted: https://github.com/vmware-tanzu/sonobuoy/compare/master...zubron:watch-retry#diff-7a369ae51c001d42bbaee0e2b0efd386R186.

>Also, we still have the problem you mentioned where the logs would be duplicated upon reconnection. Do you think it would be enough to keep track of the pods we have seen and only send the pods which we haven't seen before on podCh?

As far as I understand, if watching breaks, streaming logs may also break (e.g. network outage on client side). Then once watching is established, it would push no more pods, meaning that logs following will be broken (until new pods shows up etc).

Does streaming logs has its own retry logic? If I understand https://github.com/vmware-tanzu/sonobuoy/blob/master/pkg/client/logs.go#L353 correctly, it does not.

I see 2 paths here then:
- Store list of pods, which has already been sent to streaming and don't send duplicates. Perhaps `map[string]struct{}` with `pod.metadata.uid` as a key could be used here. This option assumes, that streaming request itself will re-establish the connection after network outage.
- When watching is re-established, cancel all currently opened streams, as next all pods will be added to streaming. We currently don't store the references to `logStreamer` objects, so we would have to add that + cancelling logic.

I am not very experienced in either go routines or working with `client-go`, but I'm happy to try resolving this and learn something new. But if this somehow becomes urgent, feel free to take over the PR :)

invidian:
>Store list of pods, which has already been sent to streaming and don't send duplicates. Perhaps map[string]struct{} with pod.metadata.uid as a key could be used here. This option assumes, that streaming request itself will re-establish the connection after network outage.

After re-reading this, it seems that retry logic in streaming itself + keeping a map of logged pods seems like a good way to go. This way, when log following breaks for a given pod, logger may check if pod disappeared and if not, continue following it. And watching pod list will work independently. 

I wonder how logger should restart following the logs. Should we set some buffer there to read last N logs? If so, some log messages may be duplicated. Also reasonable N would depend on the individual pod log rate and how fast we can re-establish following logs. Or should we start from 0? But then some log messages might get lost. Or we could track last log line and start from there then, that shouldn't be too computational intensive I think.

Selection_276

johnSchnake:
Thanks @zubron for the branch; used that as the base for https://github.com/vmware-tanzu/sonobuoy/pull/1222 . Closing this.

PR #1071, opened 2020-01-16
POC/WIP Windows image

### LINKED ISSUE #732: Support for Windows nodes
We need to make sure that Sonobuoy is built for and works on Windows clusters.

There may be some build changes to make but other than that I think this will involve mainly running/rerunning on Windows machines and finding each situation where we do something like use `path` instead of `filepath` because we assumed the node would be on linux.

johnSchnake:
First draft + tinkering of building Sonobuoy to run on Windows nodes.

- Fixes #732 

I'll be just using this as a reference as I put up smaller PRs that fix misc issues I ran into.


examples/plugins/windows-plugin/run.ps1
```diff
@@ -0,0 +1,6 @@
+$strString = "Hello World"
+$outputFile = 'c:\tmp\results\file.txt'
+$doneFile = 'c:\tmp\results\done'
+
+$strString | Out-File $outputFile -Encoding ascii
+$outputFile | Out-File $doneFile -Encoding ascii
```
johnSchnake: I'd like something more impressive than "hello world". Open to ideas, would prefer something clearly Windows-ish

pkg/client/retrieve.go
```diff
@@ -65,6 +64,13 @@ func (c *SonobuoyClient) RetrieveResults(cfg *RetrieveConfig) (io.Reader, <-chan
 		return nil, nil, errors.Wrap(err, "failed to get the name of the aggregator pod to fetch results from")
 	}
 
+	// DEBUG; dont assume Linux os
```
johnSchnake: Need to still have some logic to choose the right tar command.

I think I just need to GET the aggregator pod, see which node it is on, GET the node and check its labels for the OS.

Better idea?

pkg/client/retrieve.go
```diff
@@ -125,8 +131,9 @@ func UntarAll(reader io.Reader, destFile, prefix string) (filenames []string, re
 
 		for i := range b {
 			if b[i] != 0 {
-				returnErr = fmt.Errorf("non-zero data %v read after tar EOF", b[i])
-				return
+				//returnErr = fmt.Errorf("non-zero data %v read after tar EOF", b[i])
+				//return
+				fmt.Fprintf(os.Stderr, "non-zero data %v read after tar EOF", b[i])
```
johnSchnake: Need to revert this change. Originally I thought there was a tar difference on Windows so I wanted to see this data that was left around. In the end it was just an error on my part.

pkg/client/retrieve.go
```diff
@@ -135,14 +142,16 @@ func UntarAll(reader io.Reader, destFile, prefix string) (filenames []string, re
 		header, err := tarReader.Next()
 		if err != nil {
 			if err != io.EOF {
+				//fmt.Println("schnake next err")
```
johnSchnake: Revert.

johnSchnake:
@zubron Still a few fixups but ready for early review if you want to take a look.

manifest_spec.yaml.tmpl
```diff
@@ -0,0 +1,17 @@
+image: REGISTRY/sonobuoy:TAG
+manifests:
+  -
+    image: REGISTRY/sonobuoy-amd64:TAG
+    platform:
+      architecture: amd64
+      os: linux
+  -
+    image: REGISTRY/sonobuoy-amd64:TAG
```
johnSchnake: s/amd64/arm64

examples/plugins/windows-plugin/run.ps1
zubron: I came across this powershell command [`Get-WinEvent`](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.diagnostics/get-winevent?view=powershell-7) which gets events from the event logs. I tried it out on my Windows partition with the following `Get-WinEvent -LogName System` but that seems to include historical events that were stored, not just since the machine was booted. There are some other examples on that page though which you could probably modify and use for this.

examples/plugins/windows-plugin/plugin.yaml
```diff
@@ -0,0 +1,28 @@
+podSpec:
+  nodeSelector:
```
zubron: Will you be including a README or anything along with this to explain what is required to create a Windows plugin or will that be included as separate docs?

pkg/worker/worker.go
```diff
@@ -87,7 +89,7 @@ func GatherResults(waitfile string, url string, client *http.Client, stopc <-cha
 		select {
 		case <-ticker:
 			if resultFile, err := ioutil.ReadFile(waitfile); err == nil {
-				logrus.WithField("resultFile", string(resultFile)).Info("Detected done file, transmitting result file")
+				resultFile = bytes.TrimSpace(resultFile)
```
zubron: Is removing the logging line intentional?

DockerfileWindows
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM BASEIMAGE
+MAINTAINER John Schnake "jschnake@vmware.com"
+
+ADD build/windows/amd64/sonobuoy.exe /sonobuoy.exe
```
zubron: Should `build/windows/amd64/sonobuoy.exe` be `BINARY` as that's what the Makefile is expecting to replace.

pkg/client/retrieve.go
zubron: The only other thing I can think of right now is to have a utility script or something with the same name that exists inside both the Linux and Windows images that would take care of forming the correct tar command -  similarly to how we're now relying on the Dockerfile to know how to run the aggregator. I don't think that would work though due to the differences in client/server versions. It seems much more likely for the `retrieve` command that you could be using a newer client version against an older server version.

I think checking the node labels is probably the way to go.

DockerfileWindows
johnSchnake: 👍 thanks; started out with hardcoding and didnt revert that part on the next iteration I guess.

examples/plugins/windows-plugin/plugin.yaml
johnSchnake: Yes I'll add a readme to this plugin folder and talk about windows specific points of interest. Seems like a reasonable place to put it; may add a link elsewhere in docs to point to this as well.

scripts/ci/publish.sh
```diff
@@ -10,7 +10,7 @@ fi
 
 function image_push() {
     echo ${DOCKERHUB_TOKEN} | docker login --username sonobuoybot --password-stdin
-    make containers push
```
johnSchnake: In CI the images will be created in another step (linux+windows) and the publishing step will just need to push.

johnSchnake:
Since I am blocked currently on the actual circleCI work but not the windows stuff itself I'm going to open a separate PR and split those up.

This would mean one PR makes Windows nodes supported through a manual build process (which I'll document).

The other PR will tie it into CircleCI.

johnSchnake:
I believe this is outdated and https://github.com/vmware-tanzu/sonobuoy/pull/1081 takes over adding the windows stuff. Some of the other code was moved into another PR that already merged. If mistaken we can just reopen this.

johnSchnake:
So I think the other PR has most of this code but this one also adds stuff to the circleci config.

Intent is to first get the other PR fixed up and merged, then consider this one in some capacity. The other will enable the manual building of the windows image which may be enough for now as a POC effectively. Then this can be revisited if the windows stuff gets utilized

PR #1081, opened 2020-02-06
Increase compatibility for  building of Windows images

johnSchnake:
Currently in a beta state, we are experimenting with building
Windows images so that Sonobuoy can be run on Windows nodes.

This change stops short of adding the Windows build to CI
while we figure out some of the last technical problems there.

Changes include:
 - Utilizing circleci workspaces to load/store docker images across stages
 - Adds a Windows sample plugin
 - Updates the makefile to support building windows images and cleans up some of the makefile structure to build the binaries as appropriate
 - Updates the makefile to build/push images for windows only if PUSH_WINDOWS=true
 - Updates the way the manifest-tool is invoked to use a template instead of the command line options (which provides more flexibility)
 - Adds a flag `aggregator-node-selector` to gen/run commands in order to help direct where to place the aggregator in mixed-node clusters
 - Modify which command is used to grab the tarball of results from the aggregator depending on whether or not it is a Windows system or not (therefore if bash or powershell should be used)

- Related #732

The hardest parts of this are:
 - gettings a Windows cluster
 - getting a Windows docker context

Once those are handled the rest isn't so bad.

**Release note**:
```
Several changes related to supporting running on a Windows machine including building Windows compatible images.
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,5 @@
-version: 2
+version: 2.1
+
 jobs:
   check_go_mod:
     docker:
@@ -30,6 +31,10 @@ jobs:
       enabled: true
     steps:
       - checkout
+      - run:
+          name: Load images into docker from workspace
+          command: |
+            docker load -i /home/circleci/project/sonobuoyImages.tar.gz
       - run: ./scripts/ci/publish.sh
 
   build_clients:
@@ -45,7 +50,6 @@ jobs:
           paths: 
             - build/linux/amd64/*
             - build/linux/arm64/*
-            - build/linux/windows/amd64/*
 
   build_containers:
     machine:
@@ -57,7 +61,7 @@ jobs:
       - run:
           name: "Build Sonobuoy container images"
           command: |
-            make containers
+            make linux_containers
       - run:
           name: Save as tar files to persist to next step
           command: |
@@ -119,10 +123,10 @@ workflows:
       - check_go_mod
       - check_readme_sync
       - build_clients
+      - unit_tests
       - build_containers:
           requires:
             - build_clients
-      - unit_tests
       - integration_tests:
           requires:
             - build_clients
--- a/DockerfileWindows
+++ b/DockerfileWindows
@@ -1,0 +1,20 @@
+# Copyright 2019 Sonobuoy contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM BASEIMAGE
+MAINTAINER John Schnake "jschnake@vmware.com"
+
+ADD BINARY /sonobuoy.exe
+WORKDIR /
+CMD /sonobuoy.exe master --no-exit -v 3 --logtostderr
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,8 @@ IMAGE = $(REGISTRY)/$(TARGET)
 DIR := ${CURDIR}
 DOCKER ?= docker
 LINUX_ARCH := amd64 arm64
+WIN_ARCH := amd64
 DOCKERFILE :=
-PLATFORMS := $(subst $(SPACE),$(COMMA),$(foreach arch,$(LINUX_ARCH),linux/$(arch)))
 KIND_CLUSTER = kind
 
 # Not used for pushing images, just for local building on other GOOS. Defaults to
@@ -50,6 +50,7 @@ BUILDMNT = /go/src/$(GOTARGET)
 BUILD_IMAGE ?= golang:1.13.0-stretch
 AMD_IMAGE ?= debian:stretch-slim
 ARM_IMAGE ?= arm64v8/ubuntu:16.04
+WIN_IMAGE ?= mcr.microsoft.com/windows/servercore:1809
 
 TESTARGS ?= $(VERBOSE_FLAG) -timeout 60s
 COVERARGS ?= -coverprofile=coverage.txt -covermode=atomic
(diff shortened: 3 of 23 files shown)


.circleci/config.yml
```diff
@@ -1,4 +1,5 @@
-version: 2
+version: 2.1
```
johnSchnake: Some of these changes (like this one) are slightly preemptive. I started from a version where I was making Windows builds (unsuccessfully) and just backtracked on the explicitly windows+CI portions.

Version 2.1 is required here if we want to use the windows orb on circleCI which enables use of the windows executor.

examples/plugins/windows-plugin/run.ps1
```diff
@@ -0,0 +1,6 @@
+$strString = "Hello World"
```
johnSchnake: no need for this line

go.sum
```diff
@@ -95,6 +95,7 @@ github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMyw
 github.com/google/gofuzz v0.0.0-20161122191042-44d81051d367/go.mod h1:HP5RmnzzSNb993RKQDq4+1A4ia9nllfqcQFTQJedwGI=
 github.com/google/gofuzz v1.0.0 h1:A8PeW59pxE9IoFRqBp37U+mSNaQoZ46F1f0f863XSXw=
 github.com/google/gofuzz v1.0.0/go.mod h1:dBl0BpW6vV/+mYPU4Po3pmUjxk6FQPldtuIdl/M65Eg=
+github.com/google/martian v2.1.0+incompatible h1:/CP5g8u/VJHijgedC/Legn3BAbAaWPgecwXBIDzw5no=
```
johnSchnake: Not sure why this is updated

manifest_spec.yaml.tmpl
```diff
@@ -0,0 +1,17 @@
...
+    platform:
+      architecture: amd64
+      os: linux
+  -
+    image: REGISTRY/sonobuoy-arm64:TAG
+    platform:
+      architecture: arm64
+      os: linux
+  -
+    image: REGISTRY/sonobuoy-win-amd64:TAG
```
johnSchnake: Should have removed this section.

Moving to this manifest version instead of the `--from-args` for manifests enables easy addition of Windows in the future though.

pkg/worker/worker.go
```diff
@@ -87,7 +88,7 @@ func GatherResults(waitfile string, url string, client *http.Client, stopc <-cha
 		select {
 		case <-ticker:
 			if resultFile, err := ioutil.ReadFile(waitfile); err == nil {
-				logrus.WithField("resultFile", string(resultFile)).Info("Detected done file, transmitting result file")
```
johnSchnake: Add back log line.

scripts/ci/publish.sh
```diff
@@ -10,7 +10,7 @@ fi
 
 function image_push() {
     echo ${DOCKERHUB_TOKEN} | docker login --username sonobuoybot --password-stdin
-    make containers push
+    make push
```
johnSchnake: Eventually (when windows+linux nodes supported) we can't build both on the same machine. So presumably we've `make container` and `make win-container` in different machines and loaded the images for this step.

DockerfileWindows
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM BASEIMAGE
+MAINTAINER John Schnake "jschnake@vmware.com"
+
+ADD build/windows/amd64/sonobuoy.exe /sonobuoy.exe
```
johnSchnake: BINARY

Makefile
```diff
@@ -50,6 +50,7 @@ BUILDMNT = /go/src/$(GOTARGET)
 BUILD_IMAGE ?= golang:1.13.0-stretch
```
johnSchnake: Remove stretch just use golang so windows can build binary too

Makefile
```diff
@@ -165,13 +185,39 @@ push_images:
...
+	-e 's|REGISTRY|$(REGISTRY)|g' \
+	-e '/^WIN_ONLY/d' \
+	manifest_spec.yaml.tmpl > ./build/manifest_spec.yaml;
+endif
+
+push_manifest: gen_manifest
+	./manifest-tool push from-spec ./build/manifest_spec.yaml
 
-push: pre containers
+push: #pre
```
johnSchnake: Debug; add back pre

examples/plugins/windows-plugin/plugin.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - effect: NoSchedule
+    key: node-role.kubernetes.io/master
+    operator: Exists
+  - key: CriticalAddonsOnly
+    operator: Exists
+  - key: kubernetes.io/e2e-evict-taint-key
+    operator: Exists
+sonobuoy-config:
+  driver: Job
+  plugin-name: windows-hello-world
```
johnSchnake: rename? not really hello world now

examples/plugins/windows-plugin/plugin.yaml
```diff
@@ -0,0 +1,28 @@
...
+  result-format: raw
+  skip-cleanup: true
+spec:
+  image: schnake/windows-plugin:v1
```
johnSchnake: Will need to push a sonobuoy version of this

pkg/client/retrieve.go
```diff
@@ -141,8 +189,9 @@ func UntarAll(reader io.Reader, destFile, prefix string) (filenames []string, re
 		}
 		entrySeq++
 		mode := header.FileInfo().Mode()
-		outFileName := path.Join(destFile, header.Name[len(prefix):])
+		outFileName := filepath.Join(destFile, header.Name)
```
johnSchnake: Double check prefix

examples/plugins/windows-plugin/README.md
```diff
@@ -0,0 +1,42 @@
...
+That will schedule the Sonobuoy aggregator on a Linux node and the plugin on the Windows node. If you want the Sonobuoy aggregator to also run on a Windows node, you'll need to add the `--aggregator-node-selector=kubernetes.io/os:windows` flag:
+
+```
+sonobuoy run --plugin plugin.yaml --aggregator-node-selector=kubernetes.io/os: windows --wait
+```
+
+Optionally, if you have multiple types of Windows nodes, you may need to add the node selector for the architecture as well:
+
+```
+sonobuoy run --plugin plugin.yaml --aggregator-node-selector=kubernetes.io/os:windows --aggregator-node-selector=kubernetes.io/os:amd64 --wait
```
johnSchnake: s/os/arch the 2nd time

Makefile
johnSchnake: Not important right now actually; I dont think that is required to accomplish our goal here.

manifest_spec.yaml.tmpl
johnSchnake: Actually added a WIN_ONLY prefix so we can optionally remove the windows stuff or add it, depending on the options given to MAKE.

Makefile
johnSchnake: This should have a better solution; but `pre` does things that are CI specific (getting linux manifest-tool and logging in as sonobuoybot). When working locally we dont need that. Maybe we can just gate it instead, but still seems ugly.

examples/plugins/windows-plugin/plugin.yaml
johnSchnake: Using windows-events

pkg/client/retrieve.go
johnSchnake: So this _kind of_ doesnt really matter. The reason being that we take the base name right below this so it is a bit redundant to also strip the prefix.

I'll undo this change for now, but not sure how this may change as both OSs are supported.

johnSchnake:
Need to manually check this again on windows and probably update the manual steps in the release.md; however I think otherwise the code is good. Had manually run it all but that was a few weeks ago. Will post back results when done.

pkg/client/gen_test.go
```diff
@@ -393,6 +393,30 @@ func TestGenerateManifestGolden(t *testing.T) {
...
+		}, {
+			name: "Multiple node selectors can be added",
+			inputcm: &client.GenConfig{
+				E2EConfig:     &client.E2EConfig{},
+				Config:        newConfigWithoutUUID(),
+				NodeSelectors: map[string]string{"foo": "bar", "fizz": "buzz"},
+			},
+			goldenFile: filepath.Join("testdata", "multiple-node-selector.golden"),
+		}, {
+			name: "Node selector with empty string is OK",
```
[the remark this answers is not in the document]
johnSchnake: So the `Set` method of the flag should handle that; I added a node_selector_test.go that has a unit test for this. As a result, I removed this test case entirely since it (1) shouldn't occur and (2) produces invalid yaml anyways.

examples/plugins/windows-plugin/plugin.yaml
johnSchnake: Updated the name; still need to push the image though.

johnSchnake:
Based on @zubron 's last comment I'm going to merge once CI goes green. I fixed up some problems I found in the makefile, added 2 more tests (and removed one), and made some doc tweaks.

PR #1085, opened 2020-03-03
Add kubernetes-version flag to images subcommands

### LINKED ISSUE #1082: Add --kubernetes-version to images command and sub-commands
Hello :wave:
Thanks for building Sonobuoy

I'm generating Node base-images that can be used for multi-node conformance testing.
I'd like to use the `sonobuoy images pull` command to pre-pull all of the necessary images before baking the base image. However, when baking the base image (which includes things like installing kubernetes and docker), there is no kubernetes cluster yet.

The `sonobuoy images pull` command currently requires a kubernetes API client which means I can't use it before running a kube-apiserver.

The kubeadm U/X for this same workflow is:
```
kubeadm config images pull --kubernetes-version v1.17.0
```
It does not require an API server.

I think the auto-version inspection using the APIserver is a smart default behavior, but could we add an optional `--kubernetes-version` flag to the following sonobuoy commands that would short-circuit the kubeAPI client dependency?
```
sonobuoy images --kubernetes-version v1.17.0
sonobuoy images pull --kubernetes-version v1.17.0
sonobuoy images push --kubernetes-version v1.17.0
sonobuoy images download --kubernetes-version v1.17.0
sonobuoy images delete --kubernetes-version v1.17.0
```

I think this will be especially helpful for testing multi-node clusters and shaking down flaky test-suites.

Cheers!

**Environment:**

- Sonobuoy version: v0.17.2
- Kubernetes version: (use `kubectl version`): v0.17.0
- Kubernetes installer & version: v0.17.0
- Cloud provider or hardware configuration: Vagrant virtualbox
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

johnSchnake:
If the cluster the user wants to get images for is not available
at the time (preparing images ahead of time, for a coworker, etc)
then the user needs a way to specify the version rather than relying
on the autodetection.

This change also causes the images from `sonobuoy images` to be printed
in a sorted order for consistency.

Fixes #1082

**Release note**:
```
Adds a kubernetes-version flag to the images subcommands to allow manual specification instead of auto-detection of cluster version.
```

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -368,6 +368,15 @@ func AddPluginListFlag(p *[]string, flags *pflag.FlagSet) {
 	flags.StringSliceVarP(p, "plugin", "p", []string{"e2e", "systemd-logs"}, "Describe which plugin's images to interact with (Valid plugins are 'e2e', 'systemd-logs').")
 }
 
+// AddKubernetesVersionFlag adds a string flag for the user to specify the cluster version to assume instead of reaching
+// out to the cluster to auto-detect it.
+func AddKubernetesVersionFlag(version *string, flags *pflag.FlagSet) {
+	flags.StringVar(
+		version, "kubernetes-version", "",
+		fmt.Sprintf("Version to assume for Kubernetes. If empty, the cluster will be queried for its version"),
+	)
+}
+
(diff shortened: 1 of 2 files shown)


stealthybox:
Thanks for taking this on :+1:

PR #1092, opened 2020-03-20
Minor tweak after last commit

johnSchnake:
Uncommented something that was just for local testing and added a file to the gitignore.


johnSchnake:
Just waiting for green and then merging.

PR #1130, opened 2020-06-09
Allow arbitrary data in Details map in results

zubron:
The previous `Details` structure in the Sonobuoy results format was a
`map[string]string`. As we are creating more plugins that are providing
reports in the Sonobuoy results format, this structure is limiting. To
allow more flexibility, the type is being changed to
`map[string]interface{}` to allow plugins to provide arbitrary data in
the `Details` field.

The version of the YAML library used when parsing the results in a
results tarball has been upgraded to v3. This is due to the fact that it
unmarshals data into a `map[string]interface{}` type when all the keys
are strings. In v2, the behaviour was to always marshal to
`map[interface{}]interface{}`. This is important as if the keys are not
strings then marshalling to JSON will fail when using the `--detailed`
mode of the `results` command.

See: go-yaml/yaml#384 and go-yaml/yaml#591.

**Release note**:
```
The Sonobuoy Results format now uses map[string]interface{} for the Details field. This allows arbitrary data to be included in this field. For compatibility with other results processing operations, only string keys should be used in any maps contained within this field.
```


pkg/client/results/manual.go
```diff
@@ -58,6 +58,7 @@ func manualProcessFile(pluginDir, currentFile string) (Item, error) {
 
 	rootObj.Status = resultObj.Status
 	rootObj.Items = resultObj.Items
+	rootObj.Details = resultObj.Details
```
zubron, line 61: Including this as this field wasn't being set on the root item. It's covered in the new test in `manual_test.go`.

PR #1138, opened 2020-07-06
Upgrade Kubernetes client libraries to 0.18.5

zubron:
This changes upgrades the version the Kubernetes libraries we are using
to the latest version, 0.18.5. With this, a number of the APIs had
changed to take a `context.Context` parameter. As most of the API uses
do not have a surrounding context, these updated calls take a `TODO`
context as recommended in the Golang context documentation:
https://golang.org/pkg/context/#TODO


cmd/sonobuoy/app/args.go
```diff
@@ -42,7 +42,7 @@ const (
 	waitOutputFlag      = "wait-output"
 	customRegistryFlag  = "custom-registry"
 	kubeconfig          = "kubeconfig"
-	context             = "context"
+	kubecontext         = "context"
```
zubron, line 45: Modified to avoid clash with imported `context` package.

PR #1139, opened 2020-07-07
Deprecate use of "master" in resource labels and names

zubron:
We renamed the `master` command to `aggregator` in #847 (released in
v0.15.3). However, `master` was still being used in some of our
Kubernetes resource labels and names. We want to remove use of
insensitive language and as such use different labels for
identifying resources.

Now, any aggregator resources will have the label
`sonobuoy-component: aggregator` and plugin resources will have
`sonobuoy-component: plugin`. The existing label is still in place as it
was used to identify the aggregator pod (which is required most sonobuoy
commands once a run has started). Removing it now would prevent users
with older clients from using newer images, or querying the state of
runs started by newer clients. Instead, both labels will be applied
howeve the `sonobuoy-component: aggregator` label is the preferred
label. If the aggregator pod cannot be found using the new label, the
client will retry using the now deprecated label. When retrying, the
client will display a warning that it is attempting to find the pod
using the deprecated label.

Newer clients will be able to query the state of runs with old labels
applied, and older clients will be able to query the state of runs
created by the newer clients.

The name of the service created by Sonobuoy has been renamed to
`sonobuoy-aggregator`. This has no user facing impact.

**Release note**:
```
The label for identifying the aggregator pod is now `sonobuoy-component: aggregator` (replacing `run: sonobuoy-master`). The label `run: sonobuoy-master` is still applied but is now deprecated and its use will be removed in a later release. Once removed, older clients will not be able to query runs started by newer versions of Sonobuoy which do not use this label. A warning about the deprecated label will now be printed if using a newer client to query a run by an older version.
```

### DIFF
--- a/pkg/client/testdata/custom-systemd-logs-image.golden
+++ b/pkg/client/testdata/custom-systemd-logs-image.golden
@@ -69,6 +69,7 @@ metadata:
   labels:
     component: sonobuoy
     run: sonobuoy-master
+    sonobuoy-component: aggregator
     tier: analysis
   name: sonobuoy
   namespace: 
@@ -109,8 +110,8 @@ kind: Service
 metadata:
   labels:
     component: sonobuoy
-    run: sonobuoy-master
-  name: sonobuoy-master
+    sonobuoy-component: aggregator
+  name: sonobuoy-aggregator
   namespace: 
 spec:
   ports:
@@ -118,5 +119,5 @@ spec:
     protocol: TCP
     targetPort: 8080
   selector:
-    run: sonobuoy-master
+    sonobuoy-component: aggregator
   type: ClusterIP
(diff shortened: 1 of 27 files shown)


pkg/client/testdata/custom-systemd-logs-image.golden
```diff
@@ -109,14 +110,14 @@ kind: Service
 metadata:
   labels:
     component: sonobuoy
-    run: sonobuoy-master
-  name: sonobuoy-master
+    sonobuoy-component: aggregator
```
zubron, line 113: The aggregator service is not queried within the Sonobuoy codebase so this label could be updated without impact.

pkg/templates/manifest.go
```diff
@@ -177,15 +178,15 @@ kind: Service
 metadata:
   labels:
     component: sonobuoy
-    run: sonobuoy-master
-  name: sonobuoy-master
+    sonobuoy-component: aggregator
```
zubron, line 181: The aggregator service is not queried within the Sonobuoy codebase so this label could be updated without impact.

PR #1160, opened 2020-08-26
Prepare for Kubernetes release 1.19

vladimirvivien:
This PR syncs with the release of Kubernetes 1.19


site/docs/master/release.md
```diff
@@ -1,61 +1,77 @@
...
+2. Update the `GetImageConfigs` function within `pkg/image/manifest.go` and add a new `case` statement to the minor version check which will be the minor version of the new Kubernetes release.  In this new case, call the newly created function (e.g. `r.v1_20()`).
+3. Add the new default image registry configuration by updating the function `GetDefaultImageRegistries` within `pkg/image/manifest.go` to return the default image registries for the new version.
+    * Add a new case to the minor version check which will correspond the minor version of the new Kubernetes release.
+    * Follow previous case statements (i.e. `case 19:`), return a new `RegistryList` object which includes only the registry keys used within the registry config for that version.
+    Some registries are not applicable to include in this object as they are there to test specific image pull behavior such as pulling from a private or non-existent registry. This object should only include registries that can be successfully pulled from. The other registries are not used within the end-to-end tests.
+
+### Updating the versioned docs
```
zubron, line 42: This should be moved up a level in headings (h2) as it should be done for every release, not just new K8s releases.

site/docs/master/release.md
```diff
@@ -1,61 +1,77 @@
 # Release
 
-## Preparing a new release
+## Preparing release version
```
zubron, line 3: Did you want to include these changes in the v0.19.0 docs as well?

site/docs/master/release.md
```diff
@@ -1,61 +1,77 @@
...
+
+```
+cd pkg/image
+cp v1.{19,20}.go
+```
+         
+This file will contain a function to return the list of test images via a `map[int]Config` value.
+1. Update the name of the function accordingly. For example, if the file is for the v1.20 release, 
+         ensure the function name is `v1_20`.
+1. Copy all `map` entries stored in variable `configs` from upstream Kubernetes file [`kubernetes/test/utils/image/mafnifest`](https://github.com/kubernetes/kubernetes/blob/e19964183377d0ec2052d1f1fa930c4d7575bd50/test/utils/image/manifest.go#L202-L252). Remove the previous map entries and replace them wi the The copied entries.
```
zubron, line 35:
```suggestion
1. Copy all `map` entries stored in variable `configs` from upstream Kubernetes file [`kubernetes/test/utils/image/mafnifest`](https://github.com/kubernetes/kubernetes/blob/e19964183377d0ec2052d1f1fa930c4d7575bd50/test/utils/image/manifest.go#L202-L252). Remove the previous map entries and replace them with the copied entries.
```

pkg/image/manifest.go
```diff
@@ -114,25 +213,21 @@ func NewRegistryList(repoConfig, k8sVersion string) (*RegistryList, error) {
...
-		case 13:
-			return r.v1_13(), nil
-		case 14:
-			return r.v1_14(), nil
-		case 15:
-			return r.v1_15(), nil
-		case 16:
-			return r.v1_16(), nil
+		case 13, 14, 15, 16:
+			return nil, fmt.Errorf("version not supported for this build: %v", r.K8sVersion)
```
zubron, line 221: 🎉 I like this. Also, given that we are no longer supporting these earlier versions, can the files which define the `v1_13` to `v1_16` functions be removed?

pkg/image/manifest.go
```diff
@@ -84,8 +182,9 @@ func NewRegistryList(repoConfig, k8sVersion string) (*RegistryList, error) {
 		PrivateRegistry:         privateRegistry,
 		QuayIncubator:           quayIncubator,
 		QuayK8sCSI:              quayK8sCSI,
-		SampleRegistry:          sampleRegistry,
 		PromoterE2eRegistry:     promoterE2eRegistry,
+		SampleRegistry:          sampleRegistry,
```
zubron, line 186: I haven't checked all the registries, but this one looks like it's no longer being used so you could remove it.

pkg/image/manifest_test.go
```diff
@@ -40,21 +40,36 @@ func TestGetDefaultImageRegistryVersionValidation(t *testing.T) {
 		{
 			name:    "v1.13 is valid",
```
zubron, line 41: I would update the names of these tests to `X is not valid`.

pkg/image/v1.19.go
```diff
@@ -0,0 +1,75 @@
...
+	gcAuthenticatedRegistry := r.GcAuthenticatedRegistry
+	googleContainerRegistry := r.GoogleContainerRegistry
+	invalidRegistry := r.InvalidRegistry
+	privateRegistry := r.PrivateRegistry
+	dockerGluster := r.DockerGluster
+	quayIncubator := r.QuayIncubator
+	promoterE2eRegistry := r.PromoterE2eRegistry
+
+	configs := map[int]Config{}
+	configs[Agnhost] = Config{promoterE2eRegistry, "agnhost", "2.12"}
```
zubron, line 32: I think this might still be the 1.18 list of images? It's missing a couple that I can see here: https://github.com/kubernetes/kubernetes/blob/v1.19.0/test/utils/image/manifest.go#L202-L242 such as an iptables image.

Yet another sign that it would be good to automate this 😞

zubron:
I've left some suggestions around removing unused code (such as the functions for returning the image config for now unsupported versions), but I think it would be fine to wait and address that in a follow up PR 👍

site/docs/master/release.md
vladimirvivien: 👍

site/docs/master/release.md
vladimirvivien: Good catch, I need to regenerate the docs! Thanks.

pkg/image/manifest.go
vladimirvivien: Yes they can!!!!!!!!! 🎉

pkg/image/manifest.go
vladimirvivien: 👍

pkg/image/manifest_test.go
vladimirvivien: Oh ok got it.

pkg/image/v1.19.go
vladimirvivien: Man awesome catch.  Yeah It's possible I pulled from 1.18. 

*sigh* yes, I will push for this to be better soon.

pkg/image/manifest.go
```diff
@@ -199,62 +296,50 @@ func GetDefaultImageRegistries(version string) (*RegistryList, error) {
...
 				// cannot be pulled as they are used as part of tests for checking image pull
 				// behavior. They are omitted from the resulting config.
 				// InvalidRegistry:         invalidRegistry,
 				// GcAuthenticatedRegistry: gcAuthenticatedRegistry,
 				// PrivateRegistry:         privateRegistry,
 			}, nil
-		case 18:
+		case 19:
 			return &RegistryList{
 				E2eRegistry:             e2eRegistry,
```
zubron, line 336: This list also needs to include `BuildImageRegistry`, `E2eVolumeRegistry` and `SigStorageRegistry`.

pkg/image/v1.19.go
```diff
@@ -0,0 +1,73 @@
...
+	configs[Etcd] = Config{gcRegistry, "etcd", "3.4.9"}
+	configs[GlusterDynamicProvisioner] = Config{dockerGluster, "glusterdynamic-provisioner", "v1.0"}
+	configs[Httpd] = Config{dockerLibraryRegistry, "httpd", "2.4.38-alpine"}
+	configs[HttpdNew] = Config{dockerLibraryRegistry, "httpd", "2.4.39-alpine"}
+	configs[InvalidRegistryImage] = Config{invalidRegistry, "alpine", "3.1"}
+	configs[IpcUtils] = Config{e2eRegistry, "ipc-utils", "1.0"}
+	configs[JessieDnsutils] = Config{e2eRegistry, "jessie-dnsutils", "1.0"}
+	configs[Kitten] = Config{e2eRegistry, "kitten", "1.0"}
+	configs[Nautilus] = Config{e2eRegistry, "nautilus", "1.0"}
+	configs[NFSProvisioner] = Config{sigStorageRegistry, "nfs-provisioner", "v2.2.2"}
```
zubron, line 53: Both `sigStorageRegistry` and `e2eVolumeRegistry` (used below) need to be defined in this scope with the corresponding value from the `RegistryList` to ensure that the registry configuration is applied. Otherwise I think this is falling back to the default `const` values in `manifest.go`.

site/_config.yml
```diff
@@ -52,6 +52,18 @@ defaults:
       version: master
       gh: https://github.com/vmware-tanzu/sonobuoy/tree/master
       layout: "docs"
+  - scope:
```
zubron, line 55: Ah, this is an issue with the docs generation script. If you run it twice, it will just add the same entry. You will need to remove the duplicate values here and in the `versions` list below.

site/_data/toc-mapping.yml
```diff
@@ -5,6 +5,8 @@
 # Below is a commented out example of what this may look like:
 
 master: master-toc
+v0.19.0: v0-19-0toc
+v0.19.0: v0-19-0toc
```
zubron, line 9: This is another duplicate value.

pkg/image/manifest.go
vladimirvivien: So, that is what I wasnt sure, how is that info determined?

zubron: This function is used to create the YAML file when running `sonobuoy gen default-image-config` so it should include all the registries that form part of the [`v1_19` image list:](https://github.com/vmware-tanzu/sonobuoy/pull/1160/files/6f097931031ad12dc68dfa85445c4f040075d3bb#diff-4af7a4d28032f00415359ceaf5d98de4R18). This is just a helper so that if someone wants to provide their own custom repo config, they don't need to look in the source to figure out all the keys they need to override.

When you do `sonobuoy run --e2e-repo-config <config>` (using the config file generated from the command above), the values in that config file are then used to form the full registry path of where to push the image to. You can try this out using the `--dry-run` mode of the `images push` command.

```
sonobuoy gen default-image-config > config.yaml
# modify config.yaml to use your custom registry for all keys
sonobuoy images push --custom-registry my-registry --e2e-repo-config config.yaml --dry-run
```

The output from this should show your custom registry being used for all images.

In the `v1_19` function, every registry value used in a `Config` should be defined as the value from the `RegistryList`. This is _super_ awkward and confusing so please tell me if I've done a terrible job explaining it and I will try again!

Issue #1178, opened 2020-10-16
Move CI/CD to GitHub Action


vladimirvivien:
Investigate and move to GitHub Action

- [x] lint and sync
- [x] unit tests
- [x] integration tests
- [ ] release pipeline / workflow

wilsonehusin:
We should also integrate Windows to this to track how far / close we are on having full compatibility with Windows clusters.

vladimirvivien:
@wilsonehusin I havent look into the Windows support situation other than that one comment about how to build it on Windows machine.  So, let's discuss.  When we reboot windows support, i'd like to find a way around the requirement of using a Windows box for build.

wilsonehusin:
uhm I just realized that I still haven't ported integrations tests to GitHub Actions -- reopening this.

vladimirvivien:
Not stale

johnSchnake:
> i'd like to find a way around the requirement of using a Windows box for build.

There is a circleCI windows image that I had hoped would be able to do the build for us. I don't think there is a way around _some_ windows box _somewhere_ doing the build though. I could be wrong though but I recall that being a hard requirement on how the Windows containers work/are built.

I'm not sure how github actions impact the idea of windows builds though. I haven't had time to look though.

I 100% agree the Windows build process is a pain as-is and more/less it was a POC to show the capability.

Also, just curious what the motivation for going to Github Actions was. Is it consistency with some of the other Tanzu OSS flows? I remember when we went to circleCI it was just a tad before github actions had fully opened up/had support for what we needed and we originally wanted to be on Github actions to mainly just have all build/code etc all in the same spot.

vladimirvivien:
@johnSchnake 
I think the motivation to GitHub was simplicity and one less service to worry about.

johnSchnake:
Completed by #1243

Issue #1176, opened 2020-10-16
Housecleaning - Tidy up source code


vladimirvivien:
The code has accumulated some debt over the years and this is a small step in repaying.

 * [x] Apply linter to standardize code during CI/CD - https://github.com/golangci/golangci-lint
 * [x] Apply proper Copyright header
 * [x] Remove reference of Heptio in Header and packages
 * [x] Remove dead/unused code path

vladimirvivien:
This was done in v0.20

PR #1179, opened 2020-10-20
Integrating golangci-lint

wilsonehusin:
Repository housecleaning.

Partially addresses #1176, baby steps towards #1178 

Seems like GitHub is not running the actions on this PR as `vmware-tanzu`, but it's [available on my fork](https://github.com/wilsonehusin/sonobuoy/actions)

**Release note**:
```
NONE
```

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -112,7 +112,7 @@ func AddSystemdLogsImage(image *string, flags *pflag.FlagSet) {
 // AddKubeConformanceImageVersion initialises an image version flag.
 func AddKubeConformanceImageVersion(imageVersion *image.ConformanceImageVersion, flags *pflag.FlagSet) {
 	help := "Use default Conformance image, but override the version. "
-	help += fmt.Sprintf("Default is 'auto', which will be set to your cluster's version if detected, erroring otherwise.")
+	help += "Default is 'auto', which will be set to your cluster's version if detected, erroring otherwise."
 
 	*imageVersion = image.ConformanceImageVersionAuto
 	flags.Var(imageVersion, "kube-conformance-image-version", help)
@@ -309,7 +309,7 @@ func AddShowDefaultPodSpecFlag(flag *bool, flags *pflag.FlagSet) {
 func AddWaitOutputFlag(mode *WaitOutputMode, flags *pflag.FlagSet, defaultMode WaitOutputMode) {
 	*mode = defaultMode
 	flags.Var(
-		mode, "wait-output",
+		mode, waitOutputFlag,
 		"Whether to enable spinner on Sonobuoy. Valid modes are Silent and Spinner")
 }
 
@@ -328,7 +328,7 @@ func AddImagePullPolicyFlag(policy *ImagePullPolicy, flags *pflag.FlagSet) {
 func AddSSHKeyPathFlag(path *string, flags *pflag.FlagSet) {
 	flags.StringVar(
 		path, "ssh-key", "",
-		fmt.Sprintf("Path to the private key enabling SSH to cluster nodes."),
+		"Path to the private key enabling SSH to cluster nodes.",
 	)
 }
 
@@ -337,7 +337,7 @@ func AddSSHKeyPathFlag(path *string, flags *pflag.FlagSet) {
 func AddSSHUserFlag(user *string, flags *pflag.FlagSet) {
 	flags.StringVar(
 		user, "ssh-user", "",
-		fmt.Sprintf("SSH user for ssh-key."),
+		"SSH user for ssh-key.",
 	)
 }
 
@@ -363,7 +363,7 @@ func AddPluginListFlag(p *[]string, flags *pflag.FlagSet) {
 func AddKubernetesVersionFlag(version *string, flags *pflag.FlagSet) {
 	flags.StringVar(
 		version, "kubernetes-version", "",
-		fmt.Sprintf("Version to assume for Kubernetes. If empty, the cluster will be queried for its version"),
+		"Version to assume for Kubernetes. If empty, the cluster will be queried for its version",
 	)
 }
 
--- a/cmd/sonobuoy/app/common.go
+++ b/cmd/sonobuoy/app/common.go
@@ -1,18 +1,11 @@
 package app
 
 import (
-	"time"
 	"github.com/vmware-tanzu/sonobuoy/pkg/client"
 	sonodynamic "github.com/vmware-tanzu/sonobuoy/pkg/dynamic"
 
 	"github.com/pkg/errors"
 	"k8s.io/client-go/rest"
-)
-
-var (
-	spinnerType     int           = 14
-	spinnerDuration time.Duration = 2000 * time.Millisecond
-	spinnerColor                  = "red"
 )
 
 func getSonobuoyClient(cfg *rest.Config) (*client.SonobuoyClient, error) {
--- a/cmd/sonobuoy/app/gen.go
+++ b/cmd/sonobuoy/app/gen.go
@@ -48,7 +48,6 @@ type genFlags struct {
 	sshUser                     string
 	kubeConformanceImageVersion imagepkg.ConformanceImageVersion
 	imagePullPolicy             ImagePullPolicy
-	e2eRepoList                 string
 	timeoutSeconds              int
 	showDefaultPodSpec          bool
 
--- a/cmd/sonobuoy/app/results.go
+++ b/cmd/sonobuoy/app/results.go
@@ -136,6 +136,9 @@ func result(input resultsInput) error {
 		// capabilities.
 		r, cleanup, err := getReader(input.archive)
 		defer cleanup()
+		if err != nil {
+			lastErr = err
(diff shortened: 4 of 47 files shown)


pkg/client/logs.go
```diff
@@ -187,14 +187,12 @@ func watchPodsToStreamLogs(client kubernetes.Interface, cfg *LogConfig, podCh ch
...
-					case *v1.Pod:
-						podCh <- t
-					default:
-					}
+			v := <-ch
+			if v.Type == watch.Added && v.Object != nil {
+				switch t := v.Object.(type) {
+				case *v1.Pod:
+					podCh <- t
+				default:
```
wilsonehusin, line 195: @vladimirvivien `gosimple` flags `for { select {} }` as redundant here. I _think_ this should preserve the original behavior, since `v := <-ch` is a blocking call on either case, correct?

vladimirvivien: I am guessing there is only one channel in that select block and that's why it's flagged. Sure.

pkg/client/preflight.go
```diff
@@ -33,22 +33,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-const (
```
vladimirvivien, line 36: Similar question here, what happened to the `const` block, is this unused code or is it moved somewhere else ?

pkg/client/run.go
```diff
@@ -121,6 +121,12 @@ func (c *SonobuoyClient) RunManifest(cfg *RunConfig, manifest []byte) error {
 				seenStatus = true
 			}
 
+			// if nil below was added for coverage on staticcheck
+			// TODO: ensure this is the desired behavior
+			if status == nil {
```
vladimirvivien, line 126: Let's discuss this further.  Agreed we need to understand this is not adding new side effects.

pkg/config/loader.go
```diff
@@ -55,7 +55,7 @@ func LoadConfig() (*Config, error) {
 	}
 	defer jsonFile.Close()
 
-	b, err := ioutil.ReadAll(jsonFile)
+	b, _ := ioutil.ReadAll(jsonFile)
```
vladimirvivien, line 58: Error handling: the error should not be silenced here.

pkg/discovery/pods.go
```diff
@@ -51,9 +51,11 @@ func getPodLogOptions(cfg *config.Config) *v1.PodLogOptions {
 		LimitBytes:   podLogLimits.LimitBytes,
 	}
 
+	// TODO: investigate using LimitBytes (SizeLimitBytes is deprecated)
+	//       and SinceSeconds (TimeLimitDuration is deprecated)
 	// Only set values if they have values greater than 0 (as in they user specified).
-	limitBytes := podLogLimits.SizeLimitBytes(0)
-	sinceSeconds := int64(podLogLimits.TimeLimitDuration(0) / time.Second)
+	limitBytes := podLogLimits.SizeLimitBytes(0)                           //nolint:staticcheck
```
vladimirvivien, line 57: This is good for now.  Lets investigate what would satisfy this check and if it requires package upgrade then we can handle that in a different PR.

vladimirvivien (commented):
This is a great first start. Let's discuss further.

cmd/sonobuoy/app/args.go
```diff
@@ -309,7 +309,7 @@ func AddShowDefaultPodSpecFlag(flag *bool, flags *pflag.FlagSet) {
 func AddWaitOutputFlag(mode *WaitOutputMode, flags *pflag.FlagSet, defaultMode WaitOutputMode) {
 	*mode = defaultMode
 	flags.Var(
-		mode, "wait-output",
+		mode, waitOutputFlag,
```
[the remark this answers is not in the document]
wilsonehusin: yeah, there is
```go
var (
...
	waitOutputFlag = "wait-output"
...
)
```
which was marked as _unused_, so looks like this is the appropriate way to use it.

cmd/sonobuoy/app/common.go
```diff
@@ -1,20 +1,13 @@
...
 import (
-	"time"
 	"github.com/vmware-tanzu/sonobuoy/pkg/client"
 	sonodynamic "github.com/vmware-tanzu/sonobuoy/pkg/dynamic"
 
 	"github.com/pkg/errors"
 	"k8s.io/client-go/rest"
 )
 
-var (
```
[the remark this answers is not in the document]
wilsonehusin: yep this block is dead code. I see plenty of dead references to `spinner` throughout the codebase, so seems like it was _a thing_ back then, but we stopped at one point. we can investigate further if this doesn't sit right with you.

cmd/sonobuoy/app/gen.go
```diff
@@ -279,7 +278,7 @@ func (g *genFlags) resolveConfig() *config.Config {
 		// Set these values as if the user had requested the defaults.
 		if g.genflags != nil {
 			for _, v := range conf.PluginSelections {
-				g.genflags.Lookup("plugin").Value.Set(v.Name)
+				g.genflags.Lookup("plugin").Value.Set(v.Name) //nolint:errcheck
```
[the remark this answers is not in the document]
wilsonehusin: `//nolint:errcheck` will tell `golangci-lint` to not mark this as `errcheck` lint failure. the only resolutions I can think of are:
```go
// practically silencing error
_ = g.genflags.Lookup("plugin").Value.Set(v.Name)
```
or 
```go
// potentially changing behavior, because previously we ignored this error if they exist
err := g.genflags.Lookup("plugin").Value.Set(v.Name)
if err != nil {
	...
}
```

is there other alternatives?

cmd/sonobuoy/app/results.go
```diff
@@ -134,7 +134,7 @@ func result(input resultsInput) error {
 
 		// Load file with a new reader since we can't assume this reader has rewind
 		// capabilities.
-		r, cleanup, err := getReader(input.archive)
+		r, cleanup, _ := getReader(input.archive) // TODO: handle error checking
 		defer cleanup()
```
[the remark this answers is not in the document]
wilsonehusin: yeah this was marked as "unused variable `err`"

pkg/client/preflight.go
wilsonehusin: iirc, this one was never really used for anything, so maybe it was part of scaffolding which didn't end up being used.

wilsonehusin (commented):
some clarifications on changes and responses. I've been defaulting to silencing previously unhandled errors to preserve behavior, but I agree with the general sentiment that we should ensure they are handled appropriately -- whether to ignore (mark as insignificant) or surface it to user.

cmd/sonobuoy/app/gen.go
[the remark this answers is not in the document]
vladimirvivien: I think there may be one more option.  Looking at the doc for `errcheck` linter, it supports an external file where you can put package, functions, methods to skip.

Take a look here https://github.com/kisielk/errcheck#excluding-functions

vladimirvivien (commented):
Looks good. Left a link for you to check out for the `errcheck` linter

pkg/config/loader.go
vladimirvivien: Same thing here, let's add the error handling logic needed.  Probably should add something like this

```go
b, err := ioutil.ReadAll(jsonFile)
if err != nil {
   return nil, errors.Wrap(err, "read config")
}

File: pkg/discovery/discovery.go [hunk 10]
```diff
@@ -152,7 +152,7 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 
 	// 5. Run the queries
 	recorder := NewQueryRecorder()
-	clusterResources, nsResources, err := getAllFilteredResources(apiHelper, cfg.Resources)
+	clusterResources, nsResources, _ := getAllFilteredResources(apiHelper, cfg.Resources) // TODO: handle error checking
```
vladimirvivien, line 155: Same deal here, let's add an error handler 

```go
clusterResources, nsResources, err := getAllFilteredResources(apiHelper, cfg.Resources)
if err != nil {
	errlog.LogError(errors.Wrap(err, "getting all filtered resources"))
	return errCount + 1
}
```

vladimirvivien (commented):
Ok, added some comments about error handling in several places.

cmd/sonobuoy/app/results.go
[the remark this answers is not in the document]
wilsonehusin: I see, yeah I can agree with that. I'm a little worried that this is potentially masking errors because "only the last error is printed", but that's perhaps a separate discussion at a different time.

pkg/client/run.go
wilsonehusin: so after taking another look, it seems that in scenario where `status` is `nil`, previous behavior would have runtime de-referencing error on `switch{}`. so _prooooobably_ ok to have a more channeled error delivery?

vladimirvivien (approved):
LGTM

vladimirvivien:
@wilsonehusin Please squash commits and include signoff

Issue #1191, opened 2020-11-16
Cleanup pkg/images and remove hardcoded image configurations


wilsonehusin:
**Describe the solution you'd like**
I'd like for Sonobuoy to determine dynamically which images are needed by any plugin.

Long term work is to define a proper protocol in configuration / manifest.

Some milestones (will be continuously updated):
- [x] stop hard-coding images required by `k8s.gcr.io/conformance` image, [found in Sonobuoy's source code](https://github.com/vmware-tanzu/sonobuoy/tree/be8c386368f6c2dbefc000b2faf91b759f4db87c/pkg/image) under the files `v1.*.go` today.
...
- [ ] cleanup `pkg/images`

This is an umbrella issue for this effort.

**Anything else you would like to add:**
One way to maintain feature parity without bundling `v1.*.go` files mentioned above is to leverage [`-list-images` flag in `e2e.test` binary](https://github.com/kubernetes/kubernetes/blob/d20e3246bade17564860655e7fa57ead922d7307/test/e2e/framework/test_context.go#L308). I have a running proof-of-concept [on my fork under the branch `sync-dynamic-images`](https://github.com/wilsonehusin/sonobuoy/tree/sync-dynamic-images). **This creates dependency on Docker and `conformance` image to successfully run** `sonobuoy images {,SUBCOMMAND}`.

vladimirvivien:
@wilsonehusin i think this could be closed.

PR #1192, opened 2020-11-16
Dynamically determine image of plugins

wilsonehusin:
- Determine images required by plugins dynamically using `e2e.test -list-images`
- Removing hard-coded list of images for e2e

- Partially #1191

We may not need to do `docker run --rm` every time by caching the results. However, I think that's best addressed in a separate PR as we have discussed about having externalized persistent config.

**TODO** :
- [x] add tests to `cmd/sonobuoy/app/images_dynamic_test.go`

**Release note**:
```
Sonobuoy no longer bundles list of images required by e2e / conformance. The required images will now be determined by `k8s.gcr.io/conformance` image at runtime. This also means that all operations in `sonobuoy images` and its subcommands now attempts to pull `k8s.gcr.io/conformance` and executes `docker run --rm` internally.
```


kind-config.yaml
```diff
@@ -1,9 +1,9 @@
 kind: Cluster
-apiVersion: kind.sigs.k8s.io/v1alpha3
+apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
-  image: sonobuoy/kind-node:v1.19.0
+  image: &kind_node_image kindest/node:v1.19.3
```
vladimirvivien, line 5: Why this change to `&kind_node_image kindest/node...`

pkg/image/dryrun_client.go
```diff
@@ -18,12 +18,57 @@ package image
...
 import (
 	"github.com/sirupsen/logrus"
+	"strings"
 )
 
 // DryRunClient is an implementation of Client that logs the image operations that would
 // be performed rather than performing them.
 type DryRunClient struct{}
 
+const v1_19_images = `docker.io/library/httpd:2.4.38-alpine
```
vladimirvivien, line 28: I did not expect to see this hard coded list of images... seeing this PR is to solve that.  Is this const for a test ?

pkg/image/docker_client.go
```diff
@@ -109,6 +109,14 @@ func (i DockerClient) DeleteImages(images []string, retries int) []error {
 	return errs
 }
 
+func (i DockerClient) RunImage(image string, args ...string) ([]string, error) {
```
vladimirvivien, line 112: I don't see where this is used in this PR.

vladimirvivien (commented):
Let's clarify my comments.

kind-config.yaml
wilsonehusin: I thought it'd be more helpful to make sure that the `control-plane` and `worker` nodes have consistent versions, instead of manually ensuring they are the same.

pkg/image/docker_client.go
wilsonehusin: `cmd/sonobuoy/app/images.go` under `collectPluginImages`

https://github.com/vmware-tanzu/sonobuoy/pull/1192/files#diff-7adc43f24646df465c7cf42f8867fd929edc34efe0a346af720bc909766e2114R418

pkg/image/dryrun_client.go
wilsonehusin: ah yeah it's just for testing sonobuoy, this is a dryrun client which doesn't actually execute Docker and doesn't test the cluster.

vladimirvivien (approved):
LGMT

PR #1196, opened 2020-11-30
Use mirror.gcr.io by default

wilsonehusin:
While still rate-limited, GCR's mirror of Docker Hub has a significantly higher threshold (assuming the same rate limit for any GCR repository)

Reference: https://cloud.google.com/container-registry/quotas

None

None

**Release note**:
```
dockerLibraryRegistry default entry is now mirror.gcr.io
```

### DIFF
--- a/pkg/image/manifest.go
+++ b/pkg/image/manifest.go
@@ -118,7 +118,7 @@ const (
 const (
 	buildImageRegistry      = "k8s.gcr.io/build-image"
 	dockerGluster           = "docker.io/gluster"
-	dockerLibraryRegistry   = "docker.io/library"
+	dockerLibraryRegistry   = "mirror.gcr.io/library"
 	e2eRegistry             = "gcr.io/kubernetes-e2e-test-images"
 	e2eVolumeRegistry       = "gcr.io/kubernetes-e2e-test-images/volume"
 	etcdRegistry            = "quay.io/coreos"


johnSchnake:
Was this being debated or does this need to merge?

wilsonehusin:
This is not enough actually -- changing the manifest would not result in Sonobuoy passing the manifest to the Conformance container.

Do you want to pick this up?

johnSchnake:
Will do.

johnSchnake:
Are we trying to, by default, pass that to the k8s container in the same way they do custom registries in general? E.g. the same way we do airgap?

wilsonehusin:
> E.g. the same way we do airgap?

In my mind, yes. I don't know if that's still the best approach though. Last time I worked on this, I had these edge cases in mind:

- we use GCR by default, meaning always passing the manifest to conformance image
- users specify some other registry, so we pass that on to conformance image
- users specify that they actually want to use DockerHub, then we should detect and prevent our GCR override from being put in place

johnSchnake:
I'm not certain this is the right approach tbh at this point. It sounds like this is something that upstream should be fixing and actually, I've not seen the rate limit issues on those core images. Does anyone have logs showing it rate limiting those? Certain OSS repos can get exempted and other projects may pay for it.

As for our dockerhub images, we are starting to mirror them in our public harbor registry and so we may consider using that by default which would actually be our responsibility and a much more simple, in-code fix.

johnSchnake:
Closing for now, can be reopened if someone feels differently and we talk about it.

PR #1205, opened 2021-01-07
Use distroless images

wilsonehusin:
Use distroless container image for smaller size of deployment and less maintenance on potentially vulnerable programs which we do not use but included in Debian base container.

**Blockers**:
 #1206

N/A

- [x] amd64
- [ ] arm64

**Release note**:
```
Sonobuoy aggregator pod is now built based on Distroless container
```

### DIFF
--- a/cmd/sonobuoy/app/splat.go
+++ b/cmd/sonobuoy/app/splat.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2021 VMware Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package app
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/vmware-tanzu/sonobuoy/pkg/errlog"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmdSplat() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "splat AGGREGATOR_RESULTS_PATH",
+		Short: "Reads all tarballs in the specified directory and prints the content to STDOUT (for internal use)",
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := runSplat(args[0]); err != nil {
+				errlog.LogError(err)
+				os.Exit(1)
+			}
+		},
+		Hidden: true,
+		Args:   cobra.ExactArgs(1),
+	}
+	return cmd
+}
+
+func runSplat(dirPath string) error {
+	sonobuoyResults, err := filepath.Glob(filepath.Join(dirPath, "*.tar.gz"))
+	if err != nil {
+		return err
+	}
+
+	if err = loadResults(os.Stdout, sonobuoyResults); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func loadResults(w *os.File, filenames []string) error {
+	gzipWriter := gzip.NewWriter(w)
+	defer gzipWriter.Close()
+
+	tarWriter := tar.NewWriter(gzipWriter)
+	defer tarWriter.Close()
+
+	for _, filename := range filenames {
+		if err := addFileToTarball(tarWriter, filename); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func addFileToTarball(tarWriter *tar.Writer, filename string) error {
+	file, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	infoHeader, err := file.Stat()
+	if err != nil {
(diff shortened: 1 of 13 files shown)


wilsonehusin:
I'm putting this off for a bit, but as a reminder when I pick it back up, [`ko`](https://github.com/google/ko) might be useful to solving this.

wilsonehusin:
okay here's my investigation so far -- there are 2 types of testing failures that I have observed:
- `error retrieving results: error: /tmp/results no such file or directory`: this is when `sonobuoy status` has returned `passed` but aggregator has not finished processing. theoretically `sonobuoy run --wait` should still be blocking, yet somehow this shows up on CI
- `starting container process caused: exec: "/usr/bin/env": stat /usr/bin/env: no such file or directory: unknown`: this is how #1206 manifests in CI, as the CLI is telling the pod to execute `tar` through bash sub-shell -- this needs to be reimplemented in pure Go as `aggregate` subcommand (or a new subcommand) https://github.com/vmware-tanzu/sonobuoy/blob/cccc0b0fa92a794688e06d08a3a63ec7224bda38/pkg/client/retrieve.go#L46-L51

wilsonehusin:
in the case of arm64 support, I _think_ it's achievable? tag seems to suggest that there exists a base image for arm64 but architecture section seems to imply otherwise.

```json
{
    "Name": "gcr.io/distroless/static",
    "Digest": "sha256:b579e0c8d9ef85e3d2bb184af02cd0ea9d90b6d3b02adaa8bacea1cb346fab51",
    "RepoTags": [
        ...
        "latest",
        "latest-amd64",
        "latest-arm",
        "latest-arm64",
        "latest-ppc64le",
        "latest-s390x",
        "nonroot",
        "nonroot-amd64",
        "nonroot-arm",
        "nonroot-arm64",
        "nonroot-ppc64le",
        "nonroot-s390x"
    ],
    "Created": "1970-01-01T00:00:00Z",
    "DockerVersion": "",
    "Labels": null,
    "Architecture": "amd64",
    "Os": "linux",
    "Layers": [
        "sha256:5749e56bea7178768b00aac0bf087558fccd5b8e0c807610618be7568459f359"
    ],
    "Env": [
        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
        "SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt"
    ]
}
```

cmd/sonobuoy/app/splat.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/vmware-tanzu/sonobuoy/pkg/errlog"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmdSplat() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "splat AGGREGATOR_RESULTS_PATH",
+		Short: "Reads all tarballs in the specified directory and splat it to STDOUT (for internal use)",
```
vladimirvivien, line 36: I would what "splat" means.

cmd/sonobuoy/app/splat.go
```diff
@@ -0,0 +1,117 @@
...
+		Long:  "Sonobuoy is an introspective kubernetes component that generates reports on cluster conformance, configuration, and more",
```
vladimirvivien, line 37: `Long:` seems to be a copy/pasted. Update with proper description.

cmd/sonobuoy/app/splat.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, filename := range filenames {
+		if err := addFileToTarball(tarWriter, filename); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func addFileToTarball(tarWriter *tar.Writer, filename string) error {
```
vladimirvivien, line 84: Ah!, Your tweet makes sense now. Yes I wrote a tarrer and blogged about it cause it's one of those you should do once. After that it's copy and paste.

cmd/sonobuoy/app/splat.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// archive all the `files` found above into a single (in-memory) tarball
+	// meaning untar-ing `buffer` below would yield `files`, which are also tarballs
+	buf := bytes.NewBuffer([]byte{})
+
+	if err = loadResults(buf, sonobuoyResults); err != nil {
+		return err
+	}
+
+	fmt.Print(buf.String())
```
vladimirvivien, line 64: So if I understand this, it tars the files into an in memory buffer then write that buffer to standard out  as string?

If so, we can make this a bit better.  Since we're using an streaming io, you really don't have to save it in buffer first. You should be able to stream the tar binary directly to stdout and voiding using a buffer.  This is important for large files specially.

Here is how this could work:
* change loadResult as follows 
  ```go
   func loadResult(stdout os.File, filenames []string){
       gzipWriter := gzip.NewWriter(stdout)
   }
   ```
* Call it as follow
  ```go
   if err = loadResults(os.Stdout, sonobuoyResults); err != nil {
		return err
   }
  ```

vladimirvivien: Or something like that ^^^

vladimirvivien (commented):
OK first pass looks good.
Left some comment regarding streaming result.

cmd/sonobuoy/app/splat.go
wilsonehusin: ah okay, my worry with that approach that it might get messy with open pipes to stream vs buffer. but maybe I'm overthinking it :smile:

vladimirvivien (approved):

PR #1210, opened 2021-02-03
Build containers with ko

wilsonehusin:

To minimize the maintenance cost, we are integrating [`ko`](https://github.com/google/ko) to our process of building Sonobuoy container.

- [x] use ko-built container for running integration test
- [x] build / release pipelines to use / rely on `ko` ~(WIP: I want to get this in this PR as well -- it involves migrating Circle CI job to GitHub Actions)~
- [ ] add dockerhub token

**Release note**:
```
Building Sonobuoy containers in integration tests with `ko`
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,9 @@ WIN_ARCH := amd64
 DOCKERFILE :=
 KIND_CLUSTER = kind
 
+KO ?= ko
+KO_FLAGS = --platform=all --base-import-paths
+
 # Not used for pushing images, just for local building on other GOOS. Defaults to
 # grabbing from the local go env but can be set manually to avoid that requirement.
 HOST_GOOS ?= $(shell go env GOOS)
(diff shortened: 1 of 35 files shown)


pkg/plugin/driver/base_test.go
```diff
@@ -170,12 +169,6 @@ func TestCreateWorkerContainerDefinition(t *testing.T) {
 			return fmt.Errorf("expected worker container pull policy to be %q, but got %q", v1.PullNever, container.ImagePullPolicy)
 		}
 
-		for i, c := range container.Command {
-			if c != command[i] {
-				return fmt.Errorf("expected command item %v to be %q, got %q", i, command[i], c)
-			}
-		}
-
```
wilsonehusin, line 178: 🤔 I don't remember changing this, perhaps part of golden tests being updated?

Makefile
```diff
@@ -94,176 +93,32 @@ test:
...
-# check-kind-env will show you what will be built/tagged before doing so with kind_images
-check-kind-env:
-ifndef K8S_PATH
-	$(error K8S_PATH is undefined)
-endif
-ifndef KIND_K8S_TAG
-	$(error KIND_K8S_TAG is undefined)
-endif
-	echo --kube-root=$(K8S_PATH) tagging as --image $(REGISTRY)/kind-node:$(KIND_K8S_TAG)
+# sonobuoy run --sonobuoy-image azd2k.azurecr.io/sonobuoy:latest
```
wilsonehusin, line 124: derp. this needs to be removed.

Makefile
```diff
@@ -94,176 +93,32 @@ test:
...
-	$(MAKE) build_sonobuoy GO_SYSTEM_FLAGS="GOOS=linux GOARCH=amd64" BINARY=$@
+# Integration tests
+int: ARTIFACTS_DIR=/tmp/artifacts KUBECONFIG=$(KUBECONFIG) SONOBUOY_CLI=$(SONOBUOY_CLI)
+int: TESTARGS=$(VERBOSE_FLAG) -timeout 3m
+int: native
+	KO_DOCKER_REPO=ko.local \
+	KIND_CLUSTER_NAME=$(KIND_CLUSTER) \
+	$(KO) publish --platform=linux/amd64 --tags $(GIT_VERSION) --base-import-paths github.com/vmware-tanzu/sonobuoy
+	docker tag ko.local/sonobuoy:$(GIT_VERSION) sonobuoy/sonobuoy:$(GIT_VERSION)
+	kind load docker-image sonobuoy/sonobuoy:$(GIT_VERSION)
```
wilsonehusin, line 107: so we kinda need to do this hacky thing instead of ko directly deploying to kind because it's harder to modify the integration test such that they will be pointing to `ko.local/sonobuoy` instead of `sonobuoy/sonobuoy` since I don't think we necessarily want users to run ko?

johnSchnake:
What is left on this? Last we spoke there were still a few issues blocking it weren't there?

wilsonehusin:
so I think we're going to drop this effort. one thing is that we can't pass many `ldflags` in `ko` build context (at least I have failed to discover how). I can't really recall what else was the problem (maybe none?) but at this point it feels like marginal improvement.

I know this is supposed to pay down technical debt of our Makefile magic, but maybe not this time 😬

johnSchnake:
Closing; not really on our radar and I'm not sure what improvements this causes. Already have gotten rid of the makefile for now for simpler bash logic and have support for windows now which probably complicates this as well.

PR #1216, opened 2021-02-25
Datetime regression fix and go1.16

wilsonehusin:

Preparing for next release:
- Use `nonroot` Distroless
- Ensuring ARM64 images are built with Distroless as well
- Use go1.16
- Fix regression of missing datetime prefix on `sonobuoy retrieve`

None

**Release note**:
```
Use go1.16 and nonroot Distroless for building images
```

### DIFF
--- a/cmd/sonobuoy/app/retrieve.go
+++ b/cmd/sonobuoy/app/retrieve.go
@@ -19,7 +19,6 @@ package app
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/sonobuoy/app/retrieve.go
```diff
@@ -77,7 +75,7 @@ func retrieveResults(cmd *cobra.Command, args []string) {
 	eg := &errgroup.Group{}
 	eg.Go(func() error { return <-ec })
 	eg.Go(func() error {
-		filesCreated, err := client.UntarAll(reader, outDir, prefix)
+		filesCreated, err := client.UntarAll(reader, outDir, "")
```
vladimirvivien, line 78: Will this change work as before? If not, will that change any docs ?

wilsonehusin: This will make the result the same as before as if we still relied on `tar` binary provided by OS / distro

vladimirvivien (approved):

PR #1222, opened 2021-03-18
Avoid infinite loop when a following logs

### LINKED ISSUE #1064: 100% CPU usage after running "sonobuoy logs -f"

johnSchnake:
When streaming logs, we previously watched a list of pods so that
we could identify when new ones were created and we would stream
their logs as well.

The problem is when the watcher breaks we do not automatically
retry the connection and a channel gets closed causing an infinite
loop which eats CPU.

This uses a RetryWatcher to solve the problem.

Fixes #1064

Thanks to @zubron for her excellent notes and POC branch which I barely had to modify :)

**Release note**:
```
Sonobuoy no longers consumes 100% CPU when certain errors are encountered while streaming logs.
```

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -84,6 +84,7 @@ github.com/gogo/protobuf v1.3.1 h1:DqDEcV5aeaTmdFBePNpYsp3FlcVH/2ISVVM9Qf8PSls=
 github.com/gogo/protobuf v1.3.1/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b/go.mod h1:SBH7ygxi8pfUlaOkMMuAQtPIUF8ecWP5IEl/CR7VP2Q=
 github.com/golang/groupcache v0.0.0-20160516000752-02826c3e7903/go.mod h1:cIg4eruTrX1D+g88fzRXU5OdNfaM+9IcxsU14FzY7Hc=
+github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef h1:veQD95Isof8w9/WXiA+pa3tz3fJXkt5B7QaRBrM62gk=
 github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef/go.mod h1:cIg4eruTrX1D+g88fzRXU5OdNfaM+9IcxsU14FzY7Hc=
 github.com/golang/mock v1.1.1/go.mod h1:oTYuIxOrZwtPieC+H1uAHpcLFnEyAGVDL/k47Jfbm0A=
 github.com/golang/mock v1.2.0/go.mod h1:oTYuIxOrZwtPieC+H1uAHpcLFnEyAGVDL/k47Jfbm0A=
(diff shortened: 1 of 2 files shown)


go.sum
```diff
@@ -84,6 +84,7 @@ github.com/gogo/protobuf v1.3.1 h1:DqDEcV5aeaTmdFBePNpYsp3FlcVH/2ISVVM9Qf8PSls=
 github.com/gogo/protobuf v1.3.1/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
 github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b/go.mod h1:SBH7ygxi8pfUlaOkMMuAQtPIUF8ecWP5IEl/CR7VP2Q=
 github.com/golang/groupcache v0.0.0-20160516000752-02826c3e7903/go.mod h1:cIg4eruTrX1D+g88fzRXU5OdNfaM+9IcxsU14FzY7Hc=
+github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef h1:veQD95Isof8w9/WXiA+pa3tz3fJXkt5B7QaRBrM62gk=
```
wilsonehusin, line 87: 🤔 which version of go are you running?

pkg/client/logs.go
```diff
@@ -173,15 +175,30 @@ func getPodsToStreamLogs(client kubernetes.Interface, cfg *LogConfig, podCh chan
...
+	listOptions := metav1.ListOptions{TimeoutSeconds: &timeoutSeconds}
 	if cfg.Plugin != "" {
 		selector := metav1.AddLabelToSelector(&metav1.LabelSelector{}, "sonobuoy-plugin", cfg.Plugin)
 		listOptions = metav1.ListOptions{LabelSelector: metav1.FormatLabelSelector(selector)}
 	}
 
-	watcher, err := client.CoreV1().Pods(cfg.Namespace).Watch(context.TODO(), listOptions)
+	lw := &cache.ListWatch{
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			return client.CoreV1().Pods(cfg.Namespace).Watch(context.TODO(), listOptions)
```
wilsonehusin, line 187: probably not to be addressed in this PR, but I'm wondering if we need to handle graceful cancellation here if user hits Ctrl+C / sends SIGINT, then we should cancel the context?

wilsonehusin (commented):
overall LGTM!

go.sum
johnSchnake: 1.16.1; you get a different `go mod tidy` result? The CI test passed for that, I presumed that meant this file was as expected.

wilsonehusin: heh, [I think our CI is still on go1.15](https://github.com/vmware-tanzu/sonobuoy/blob/master/.github/workflows/ci-lint.yaml#L18). I thought go1.16 mod is always less noisy because I've only seen go1.16 doing `go mod tidy` to delete entries, and still backwards compatible with 1.15 🤔

PR #1225, opened 2021-03-22
Bump sonobuoy version

johnSchnake:
The executable on github is correctly reporting v0.50.0 but not
master. So developer builds are incorrectly showing v0.20.0.


wilsonehusin (approved):
derp I missed this, thanks for catching it John!

PR #1235, opened 2021-04-20
Add support for testing against the most recent dev E2E image

### LINKED ISSUE #1231: Test easily against k8s tip
It would be useful to have an easy way to test against k8s master. I'm assuming there are nightly releases somewhere that we can grab. It is a pain. to have to specify the image manually though if the name is changing. Something like a `tip` option would be great.

cc @jayunit100

johnSchnake:
Special handling to identify the E2E image most recently pushed
for dev builds. Very helpful if you are developing recent features
or demoing them. Specify the --kube-conformance-image-version=latest
to get this behavior.

Fixes #1231

**Release note**:
```
Test against the master branch of k8s by specifying the --kube-conformance-image-version=latest.
```

### DIFF
--- a/cmd/sonobuoy/app/images.go
+++ b/cmd/sonobuoy/app/images.go
@@ -407,7 +407,7 @@ func collectPluginsImages(plugins []string, k8sVersion string, client image.Clie
 		case systemdLogsPlugin:
 			images = append(images, config.DefaultSystemdLogsImage)
 		case e2ePlugin:
-			conformanceImage := resolveConformanceImage(k8sVersion)
+			conformanceImage := fmt.Sprintf("%v:%v", config.UpstreamKubeConformanceImageURL, k8sVersion)
 			images = append(images, conformanceImage)
 			logrus.Info("conformance image to be used: ", conformanceImage)
 
--- a/pkg/image/imageversion.go
+++ b/pkg/image/imageversion.go
@@ -18,10 +18,13 @@ package image
 
 import (
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"strings"
 
 	version "github.com/hashicorp/go-version"
 	"github.com/pkg/errors"
+	"github.com/vmware-tanzu/sonobuoy/pkg/config"
 	"k8s.io/client-go/discovery"
 )
 
@@ -36,8 +39,12 @@ var (
 const (
 	// ConformanceImageVersionAuto represents detecting the server's kubernetes version.
 	ConformanceImageVersionAuto = "auto"
-	// ConformanceImageVersionLatest represents always using the server's latest version.
+	// ConformanceImageVersionLatest represents always using the server's latest dev version.
 	ConformanceImageVersionLatest = "latest"
+
+	// DevVersionURL is the URL which should respond with a simple text of the latest version for devs.
+	DevVersionURL      = "https://storage.googleapis.com/k8s-release-dev/ci/latest.txt"
+	DevVersionImageURL = "gcr.io/k8s-staging-ci-images/conformance"
 )
 
 // String needed for pflag.Value.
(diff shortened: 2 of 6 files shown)


johnSchnake:
Thanks @dims for pointing me to the right registries. I had found the ci tarballs but not the images so this 100% solved our issue. 🥇

pkg/image/imageversion.go
```diff
@@ -69,22 +76,53 @@ func (c *ConformanceImageVersion) Set(str string) error {
...
+	}
+	defer r.Body.Close()
+	b, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return "", errors.Wrap(err, "error reading body of latest dev version")
+	}
+
+	// Metadata is added as part of the tag prefixed with an underscore instead of a plus
+	// e.g. X.Y.z-alpha0.123+ef839g is tagged as X.Y.z-alpha0.123_ef839g
+	version := strings.Replace(string(b), "+", "_", 1)
```
dims, line 124: i forgot to mention this :) sorry!

johnSchnake: No worries; it was easy to spot on a quick test. 👍

cmd/sonobuoy/app/images.go
```diff
@@ -407,7 +407,7 @@ func collectPluginsImages(plugins []string, k8sVersion string, client image.Clie
 		case systemdLogsPlugin:
 			images = append(images, config.DefaultSystemdLogsImage)
 		case e2ePlugin:
-			conformanceImage := resolveConformanceImage(k8sVersion)
+			conformanceImage := fmt.Sprintf("%v:%v", config.UpstreamKubeConformanceImageURL, k8sVersion)
```
wilsonehusin, line 410: :100:

pkg/image/imageversion.go
```diff
@@ -69,22 +76,53 @@ func (c *ConformanceImageVersion) Set(str string) error {
...
 }
 
-// Get retrieves the preset version if there is one, or queries client if the ConformanceImageVersion is set to `auto`.
+// Get retrieves the preset version if there is one, queries client if the ConformanceImageVersion is set to `auto`,
+// or finds the latest dev image published.
 // kubernetes.Interface.Discovery() provides ServerVersionInterface.
 // Don't require the entire kubernetes.Interface to simplify the required test mocks
-func (c *ConformanceImageVersion) Get(client discovery.ServerVersionInterface) (string, error) {
-	if *c == ConformanceImageVersionAuto {
+func (c *ConformanceImageVersion) Get(client discovery.ServerVersionInterface, latestURL string) (registry, version string, returnErr error) {
```
wilsonehusin, line 83: nit: probably the named return values is insignificant here?

wilsonehusin (approved):
neat! 🎉

pkg/image/imageversion.go
johnSchnake: The reason to add naming is to disambiguate the two string values; otherwise its not clear what a `string string error` values will indicate. Since those are named the error needs named too to compile.

wilsonehusin: ah ok makes sense

pkg/image/imageversion.go
```diff
@@ -36,8 +39,12 @@ var (
 const (
 	// ConformanceImageVersionAuto represents detecting the server's kubernetes version.
 	ConformanceImageVersionAuto = "auto"
-	// ConformanceImageVersionLatest represents always using the server's latest version.
+	// ConformanceImageVersionLatest represents always using the server's latest dev version.
 	ConformanceImageVersionLatest = "latest"
+
+	// DevVersionURL is the URL which should respond with a simple text of the latest version for devs.
+	DevVersionURL      = "https://storage.googleapis.com/k8s-release-dev/ci/latest.txt"
+	DevVersionImageURL = "gcr.io/k8s-staging-ci-images/conformance"
```
wilsonehusin, line 47: oh wait isn't this supposed to be a `k8s.gcr.io` URL?

dims: nope. this is the image minted by CI jobs against tip of master. we don't push those to prod registry

pkg/image/imageversion.go
```diff
@@ -36,8 +39,12 @@ var (
 const (
 	// ConformanceImageVersionAuto represents detecting the server's kubernetes version.
 	ConformanceImageVersionAuto = "auto"
-	// ConformanceImageVersionLatest represents always using the server's latest version.
+	// ConformanceImageVersionLatest represents always using the server's latest dev version.
 	ConformanceImageVersionLatest = "latest"
+
+	// DevVersionURL is the URL which should respond with a simple text of the latest version for devs.
+	DevVersionURL      = "https://storage.googleapis.com/k8s-release-dev/ci/latest.txt"
```
vladimirvivien, line 46: @johnSchnake what happens if those urls change or go away? Can user overwrite with a flag (i know we already got so much flags).

vladimirvivien (commented):
Looks good. Left a question/comment.

pkg/image/imageversion.go
johnSchnake: You can always specify an image directly with an existing flag. I dont think we need to provide a flag to separately change this URL. If it all changes then we patch sonobuoy is all, which is the same thing we'd have to do even if we had this set-but-customizable.

PR #1237, opened 2021-04-22
WIP Emits CloudEvents with status changes

johnSchnake:
Any status change would trigger an event; this includes
changes in plugin status meaning that as E2E tests finish
then sonobuoy would emit a cloudEvent.

- Fixes #

**Release note**:
```
release-note
```


johnSchnake:
Would like just a simple cloudevents echo-er as a sidecar with sonobuoy to test this. Thought that would already exist but may need to just build/push one?

wilsonehusin:
@n3wscott CloudEvents :tada: `:n3wscott:`

johnSchnake:
If this is all it takes then 👍 We just need a DD about what events to send, their types, etc. I think we'd need to have good docs  on a feature like this from the start or its unusable mostly.

So I think open questions would be:
 - List of events names/types
 - structure of those values (status change here was straight forward since its json that was already exposed)
 - when to send events
 - presumably, we'd need a flag to specify where to send events
 - need to allow multiple addresses for recipients? Or will we only send them to one?
 - Do we need to allow users to specify which events to get or do we just send them all and let the receiver filter?

Anything else?

wilsonehusin:
so... I think most points above can be iterated upon, but if we can shift the JSON statuses to CloudEvents-compliant format, then the rest can take shape as we figure out "how are we going to serve this" while getting some input from folks upstream (like the iicoop who uses Sonobuoy quite extensively iirc).

wilsonehusin:
alternatively, what an MVP could be is to have `sonobuoy status --progress` which shows how many tests have run and time elapsed with regular progress bar.

we can achieve this without CloudEvents -- but I think if we start promoting progress bar as a feature, folks might want to get creative on event consumers which makes me think it might be beneficial for us to have the structure in first 😉

johnSchnake:
@wilsonehusin The progress bar is very low hanging fruit; that info is all available from the data in `sonobuoy status --json`. If you want you can experiment with different output formats.

There are the various `--wait` options where it can wait silently or with periodic output. Perhaps you can just add another one that is optional for now (just to see how it works) that has `--wait=progress` and will output/update progress bars and/or failures for the various plugins.

johnSchnake:
Could even make a separate issue for that. I think it is a nobrainer that we need to utilize that; we just hadn't gotten around to it yet

PR #1240, opened 2021-04-30
Move away from Makefile

johnSchnake:
Every dev hates the Makefile with its unique syntax and rules
and we sometimes run into issues as a result of them causing
unintended consequences. Most everyone can write/use bash and
we were really just using Make to organize/call bash functions
anyways.

Now we have a script with functions we can source/call. If we want
we can later add argument handling so that we don't have to source
it and can instead actually call it like an executable.

Note: Will undoubtedly iterate on this after merging but since the makefile is still in place and
publishing will use that, then I think its fine to just move forward. Once everything uses the
script for a bit then we can remove the makefile.

2nd Note: Also have a PR after this one where I break up/reorg the github actions to be a bit more
efficient. It will cache builds and also help facilitate the publishing after that.


.github/workflows/ci-test.yaml
```diff
@@ -8,24 +8,24 @@ jobs:
...
-          make stress
+          source ./scripts/build.sh; stress
   integration-test-on-kind:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: run tests
         run: |
-          SONOBUOY_CLI=../../build/linux/amd64/sonobuoy ./scripts/run_integration_tests.sh
+          source ./scripts/build.sh; prepare_and_run_integration
```
johnSchnake, line 31: Using source for now; will probably add a handler for arguments so that we can call it directly. I ran into a silly problem where I was testing this all locally and my terminal kept closing. But it was because of the source & set -e, then on failure my own terminal would exit :) Took me longer than necessary to figure that one out.

I'll go ahead and start working on the handler code, it can be in this PR or the next depending on when you review/when I'm done.

johnSchnake, line 31: Using source for now; will probably add a handler for arguments so that we can call it directly. I ran into a silly problem where I was testing this all locally and my terminal kept closing. But it was because of the source & set -e, then on failure my own terminal would exit :) Took me longer than necessary to figure that one out.

I'll go ahead and start working on the handler code, it can be in this PR or the next depending on when you review/when I'm done.

scripts/build.sh
```diff
@@ -0,0 +1,293 @@
+#!/bin/bash
+
+set -e
+
+BINARY=sonobuoy
+TARGET=sonobuoy
+GOTARGET=github.com/vmware-tanzu/$TARGET
+GOPATH=$(go env GOPATH)
+REGISTRY=sonobuoy
+#REGISTRY=schnake
```
johnSchnake, line 10: delete

scripts/build.sh
```diff
@@ -0,0 +1,293 @@
...
+}
+
+pre() {
+	wget https://github.com/estesp/manifest-tool/releases/download/v1.0.1/manifest-tool-linux-amd64 \
+        -O manifest-tool
+	chmod +x ./manifest-tool
+	echo $DOCKERHUB_TOKEN | docker login --username sonobuoybot --password-stdin
+}
+
+# TODO: Make it easy to build single container for a specific arch
```
johnSchnake, line 82: Do we care about this? Would be relatively trivial to do.

scripts/build.sh
```diff
@@ -0,0 +1,293 @@
+#!/bin/bash
```
wilsonehusin, line 1: I'd recommend using `!/usr/bin/env bash` here -- it's not always desirable to use plain `/bin/bash` over the current environment.

scripts/build.sh
```diff
@@ -0,0 +1,293 @@
...
+
+linux_containers() {
+	for arch in $LINUX_ARCH; do
+		if [ "$arch"="amd64" ]; then
+			sed -e "s|BASEIMAGE|$AMD_IMAGE|g" \
+			    -e 's|CMD1||g' \
+			    -e 's|BINARY|build/linux/amd64/sonobuoy|g' Dockerfile > Dockerfile-$arch
+			build_container_dockerfile_image Dockerfile-$arch sonobuoy
+			build_container_dockerfile_image Dockerfile-$arch sonobuoy-$arch
+	    elif [ "$arch"="arm64" ]; then
```
wilsonehusin, line 101: indentation seems off on this line. should we add some bash linters or would it be overkill? 🤔

scripts/build.sh
```diff
@@ -0,0 +1,293 @@
...
+			build_container_dockerfile_image DockerfileWindows-$arch sonobuoy
+			build_container_dockerfile_image DockerfileWindows-$arch sonobuoy-win-$arch
+		else 
+			echo "Windows ARCH unknown"
+        fi
+	done
+}
+
+build_binary_GOOS_GOARCH() {
+	LDFLAGS="-s -w -X $GOTARGET/pkg/buildinfo.Version=$GIT_VERSION -X $GOTARGET/pkg/buildinfo.GitSHA=$GIT_REF_LONG"
```
johnSchnake, line 126: Should add a comment here. The LDFLAGS is the only real issue with the script. It was a bit of a pain to work out because our env vars are making up the string which is being saved into a variable which is being passed as an argument into the container and then (previously) was being interpreted by that shell.

As a comment below says; I removed one layer of shell and then worked it out.

Very good/helpful article here: http://mywiki.wooledge.org/BashFAQ/050

scripts/build.sh
```diff
@@ -0,0 +1,293 @@
...
+}
+
+linux_containers() {
+	for arch in $LINUX_ARCH; do
+		if [ "$arch"="amd64" ]; then
+			sed -e "s|BASEIMAGE|$AMD_IMAGE|g" \
+			    -e 's|CMD1||g' \
+			    -e 's|BINARY|build/linux/amd64/sonobuoy|g' Dockerfile > Dockerfile-$arch
+			build_container_dockerfile_image Dockerfile-$arch sonobuoy
+			build_container_dockerfile_image Dockerfile-$arch sonobuoy-$arch
```
wilsonehusin, line 100: do you think it's worth extracting this to a separate command? allowing people to just recreate the Dockerfile without building the container?

scripts/build.sh
```diff
@@ -0,0 +1,293 @@
...
+setup_kind_cluster(){
+    if ! kind get clusters | grep -q "^$KIND_CLUSTER$"; then
+        kind create cluster --name "$KIND_CLUSTER" --config kind-config.yaml
+        # Although the cluster has been created, not all the pods in kube-system are created/available
+        sleep 20
+    fi
+}
+
+# High level function; composes functions above and additional functionality. Typically just used by CI.
+prepare_and_run_integration() {
```
johnSchnake, line 276: This function takes over what the run_integration_tests.sh used to do. Seems like it belongs here now. The only reason it was split off was because make was a pain I think :)

test/integration/testImage/build.sh
```diff
@@ -0,0 +1,14 @@
...
+IMAGE_VERSION=v0.1
+
+build(){
+	docker build \
+    	-t $REGISTRY/$TARGET:$IMAGE_VERSION \
+    	-f Dockerfile \
+		.
+}
+
+build
```
johnSchnake, line 14: Converted the tiny makefile here to a tiny script as well. No reason not to really. Just being consistent.

scripts/build.sh
```diff
@@ -0,0 +1,293 @@
+#!/bin/bash
+
```
wilsonehusin, line 2: I started using `PS4` to show transparency to users on which command we're running. I'd like to know if you think this is useful.

```bash
PS4='\033[1m[$0:${LINENO}] $\033[0m '
set -ex
set -o pipefail
```

wilsonehusin (commented):
I didn't realize we're nuking the whole Makefile, I was under the impression that we're doing the Kubernetes style Makefile with single line call to shell script. I'm okay with this too!

scripts/build.sh
wilsonehusin: I think I added this comment. I found the delta between building 2 containers and 1 container to be quite significant when I was trying to test locally. Would love to see it happen but also it seems to be a pretty rare case.

scripts/build.sh
johnSchnake: Cool idea 👍

scripts/build.sh
johnSchnake: Yea, that works with the other comment above about wanting to build just a single os/arch too. I'll split it into finer parts and compose as needed.

.github/workflows/ci-test.yaml
johnSchnake: I've gone back and forth on this. Sourcing may end up being the best way since the alternatives are:
 - a separate "cli" in bash to ensure we call into the functions properly. This ends up being nearly as long as the file itself since we have so many short functions
 - a go wrapper with the same pitfalls as the above

At least with the script as is, it is just a single, ubiquitous language and we don't have to bounce around to figure out whats going on or google (as much) esoteric language/edge cases.

Just shouldn't source into your own shell usually :) Happy to discuss and bounce ideas around if there's something better. It really is just for devs+ci so I think it'll be fine regardless of what we do.

scripts/build.sh
johnSchnake: Maybe in another PR to separate out changes. But not a bad idea, github actions has one so even if it isn't in our typical bash script to run locally we can put it on ci if we want. And devs can just run shellcheck manually if needbe

scripts/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+# Splitting build_funcs apart from this is a small quality of life improvement
+# for when testing these locally. You don't want to source the file to get the
+# functions but also necessarily set these values which will exit the terminal.
```
wilsonehusin, line 5: 🤔 I never thought of this but it's an interesting workflow. Does this mean that if I have `some_custom_func()` in the `build_funcs.sh` that I `source`d (without `build.sh`), and the 4th out of 6 lines in that function fails, will it still continue? Or will it abort?

scripts/build_funcs.sh
```diff
@@ -0,0 +1,310 @@
...
+    )
+    build_binaries
+    linux_containers
+
+    # Load images into kind cluster. Assume sonobuoy images already built.
+    kind load docker-image --name $KIND_CLUSTER $REGISTRY/$TARGET:$IMAGE_VERSION || true
+    kind load docker-image --name $KIND_CLUSTER $TEST_IMAGE
+
+    VERBOSE=true SONOBUOY_CLI=../../build/linux/amd64/sonobuoy integration
+}
```
wilsonehusin, line 310: nit: newline EOF

test/integration/testImage/build.sh
wilsonehusin, line 14: nit: newline EOF

scripts/build_funcs.sh
```diff
@@ -0,0 +1,310 @@
...
+IMAGE_VERSION=$(git describe --always --dirty --tags)
+IMAGE_TAG=$(echo $IMAGE_VERSION | cut -d. -f1,2)
+IMAGE_BRANCH=$(git rev-parse --abbrev-ref HEAD | sed 's/\///g')
+GIT_REF_SHORT=$(git rev-parse --short=8 --verify HEAD)
+GIT_REF_LONG=$(git rev-parse --verify HEAD)
+
+BUILDMNT=/go/src/$GOTARGET
+BUILD_IMAGE=golang:1.16
+AMD_IMAGE=gcr.io/distroless/static:nonroot
+ARM_IMAGE=gcr.io/distroless/static:nonroot-arm64
```
wilsonehusin, line 33: for the record, I don't know if / how this works (I don't have `arm64` Linux machine to test if the builds are _really_ `arm64`), pretty much trusting Docker for what it says.

wilsonehusin (approved):
small notes -- lgtm otherwise!

scripts/build.sh
johnSchnake: Yes, if you want `set -e` then you need to set it in your terminal first. This is the best way I can think to do it, otherwise you have no choice about it (since even if you set +e, the script will reset -e and close your terminal on failure). And like I found, exiting the terminal youre working in probably isn't what you wanted. Especially for tests usually when testing locally you just want to run and see pass/fail.

scripts/build_funcs.sh
johnSchnake: Yea, for now, that's kind of the status quo; we hit that a year or two ago as well and even talked about cutting off arm support since we weren't testing it. However, a few people in the community pushed back and said they do use it and so we just count on them for signal/feedback

test/integration/testImage/build.sh
johnSchnake: Thanks; used to have that setting but time away/updates seems to have removed it. Fixed.

scripts/build.sh
wilsonehusin: heh I wish there is a middle ground of "don't continue the rest of the lines, but also don't quit the current shell"

¯\\_(ツ)\_/¯

johnSchnake:
CircleCI is the only thing red so I'm merging and moving forward.

Lots of things hit shellcheck (mostly quoting) so I'm doing that next. Should be fast.

PR #1241, opened 2021-05-03
Add shellcheck and fix issues

johnSchnake:
- Adds a step in github actions to run shellcheck
- Fixes lots of quoting issues in bash scripts

### DIFF
--- a/.github/workflows/ci-lint.yaml
+++ b/.github/workflows/ci-lint.yaml
@@ -9,6 +9,14 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           version: v1.31
+  bash-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run Shellcheck
+        uses: azohra/shell-linter@v0.4.0
+        with:
+          severity: "error"
   go-mod-tidy:
     runs-on: ubuntu-latest
     steps:
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -9,5 +9,7 @@ set -o pipefail
 # Getting the scripts directory can be hard when dealing with sourcing bash files.
 # Github actions has this env var set already and locally you can just source the
 # build_func.sh yourself. This is just a best effort for local dev.
-GITHUB_WORKSPACE=${GITHUB_WORKSPACE:-$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)}
-source $GITHUB_WORKSPACE/scripts/build_funcs.sh
+GITHUB_WORKSPACE=${GITHUB_WORKSPACE:-$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)}
+
+# shellcheck source=/dev/null
+source "$GITHUB_WORKSPACE"/scripts/build_funcs.sh
--- a/scripts/build_funcs.sh
+++ b/scripts/build_funcs.sh
@@ -6,10 +6,9 @@ PS4='\033[1m[$0:${LINENO}] $\033[0m '
 
 BINARY=sonobuoy
 TARGET=sonobuoy
-GOTARGET=github.com/vmware-tanzu/$TARGET
+GOTARGET=github.com/vmware-tanzu/"$TARGET"
 GOPATH=$(go env GOPATH)
 REGISTRY=sonobuoy
-IMAGE=$REGISTRY/$TARGET
 LINUX_ARCH=(amd64 arm64)
 WIN_ARCH=amd64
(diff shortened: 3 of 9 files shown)


.github/workflows/ci-lint.yaml
```diff
@@ -9,6 +9,13 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           version: v1.31
+  bash-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run Shellcheck
+        uses: azohra/shell-linter@latest
+        run: |
+          shellcheck -x ./scripts/*.sh
```
johnSchnake, line 18: Trying this out; the instructions online show me how to do a few things, but not this exactly. Will watch CI and see how it goes.

.github/workflows/ci-lint.yaml
```diff
@@ -9,6 +9,14 @@ jobs:
...
         with:
           version: v1.31
+  bash-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run Shellcheck
+        uses: azohra/shell-linter@v0.4.0
+        with:
+          severity: "error"
```
johnSchnake, line 19: Using error rather than a more sensitive value. It gives warnings for odd things like certain files that aren't bash ("Dockerfile is not scanned, if its bash start with a shebang..." is a warning) and other things.

Error has caught all the real issues that I've seen (quoting, looping, spacing, etc)

scripts/build.sh
```diff
@@ -9,5 +9,7 @@ set -o pipefail
 # Getting the scripts directory can be hard when dealing with sourcing bash files.
 # Github actions has this env var set already and locally you can just source the
 # build_func.sh yourself. This is just a best effort for local dev.
-GITHUB_WORKSPACE=${GITHUB_WORKSPACE:-$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)}
-source $GITHUB_WORKSPACE/scripts/build_funcs.sh
+GITHUB_WORKSPACE=${GITHUB_WORKSPACE:-$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)}
+
+# shellcheck source=/dev/null
```
johnSchnake, line 14: Directive to avoid an issue here; it gets confused because of the variable reference.

scripts/build_funcs.sh
```diff
@@ -39,77 +38,77 @@ unit_local() {
...
 	docker build \
-        -t $REGISTRY/$2:$IMAGE_VERSION \
-        -t $REGISTRY/$2:$IMAGE_TAG \
-        -t $REGISTRY/$2:$IMAGE_BRANCH \
-        -t $REGISTRY/$2:$GIT_REF_SHORT \
-        -f $1 \
+        -t $REGISTRY/"$2":"$IMAGE_VERSION" \
+        -t $REGISTRY/"$2":"$IMAGE_TAG" \
+        -t $REGISTRY/"$2":"$IMAGE_BRANCH" \
+        -t $REGISTRY/"$2":"$GIT_REF_SHORT" \
```
johnSchnake, line 88: fixup, messed up some mass edits it looks like

johnSchnake:
So the integrationt tests "broke" in ci but I think it is a weird github re-run issue. It is pulling the wrong SHA for some reason. You can verify the short SHA in the env var (from git describe) is wrong in CI; different than what we get locally or what CI shows as the actual short sha. That is causing the image to not load into kind and fail.

I have a followup branch to this where I do more in CI and it uses the same code w/o problem: https://github.com/johnSchnake/sonobuoy/tree/gh-release

Either we can merge this as-is (IMO that is fine short term) or I can patch this to first check the current env for those values from CI.

scripts/build_funcs.sh
```diff
@@ -6,10 +6,9 @@ PS4='\033[1m[$0:${LINENO}] $\033[0m '
 
 BINARY=sonobuoy
 TARGET=sonobuoy
-GOTARGET=github.com/vmware-tanzu/$TARGET
+GOTARGET=github.com/vmware-tanzu/"$TARGET"
```
wilsonehusin, line 9: should this be quotes or `${TARGET}`?

wilsonehusin (approved):
one nit: i don't remember which one is the convention, but otherwise LGTM

scripts/build_funcs.sh
johnSchnake: Even with the brackets shellcheck will want quotes. The brackets can, in some cases, make it more readable because it maks it stand out a bit more. The quotes are more important because of word breaking and such.

PR #1243, opened 2021-05-05
Publish images/release in github actions

johnSchnake:
Lots of updates related to build/release:

 - Get rid of manifest tool and use docker manifest
 - Changes build/tagging so we always tag images with thier
arch and only the manifest is "sonobuoy/sonobuoy". This resolves
some build/testing issues where arm64 and amd64 images were both
being tagged the same and messing up CI/testing.
 - Slightly modifies which tags we push on release; no longer tagging
as HEAD
 - Logs in to dockerhub and runs goreleaser as github actions


vladimirvivien (commented):
/LGTM

johnSchnake:
I'll probably have a few fixups as I see how it works in the CI of Sonobuoy; I had made some temp changes on my fork so I could test it with my Dockerhub info.

Lots changed here, but its all interconnected and just on the dev side so I felt a bigger PR here was ok.

The main preamble to this would be to say I have no idea why the image tagging/pushing didn't have more problems before. We were building multiple architectures with the same name/tag but only once I started moving to github actions did I see issues caused by that; kind was frustrated by seeing the arm images.

Will double check the push logic for images is the same as it was in make/circleci

wilsonehusin (approved):
wooohoooo thank you for finishing this, you're the best! ✨

johnSchnake:
:+1: Thanks for the positive feedback. It was a bit of a slog at times but we should be just on one CI and not have t deal with makefile syntax after this :) Then I can focus on getting those windows images built.

Changes in my last push:
 - Ensure images are pushed when merging onto master, whether or not it is a tag.
 - If it is a tag, just tag slightly differently (latest, version, etc)

Maybe a tweak or two left in that bit of syntax I just modified.

Will merge once I feel confident, which luckily will trigger an image push I can verify without a full release.

johnSchnake:
- Fixed up syntax issue
- Found and fixed an issue where the windows binary wasn't in the artifacts (it was in the release, but not just the built artifacts)
- Tweaked the artifact naming so that it uses the run_id. Ensures when you download it you can track it back to the run that created it.

Merging and ensuring the image pushing works (already did confirm on my fork it did). Added dockerhub token/username as secrets on the repo to accomplish this.

Issue #1227, opened 2021-04-01
go1.17 deprecates GODEBUG=x509ignoreCN=0


wilsonehusin:
The way we generate CA for integration test isn't fully compliant with CA spec (this is all I remembered, don't quite remember _how_, will add details later) and go1.15 started enforcing that case which failed our tests. [I added `GODEBUG=x509ignoreCN=0`](https://github.com/vmware-tanzu/sonobuoy/pull/1186/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R59) because it was only for test at crunch time 😅 

To be clear: this isn't a bug in Sonobuoy itself, it's just a test data which needs to be fixed

[Deprecation ref here](https://github.com/golang/go/issues/24151#issuecomment-719068842).

PR #1246, opened 2021-05-06
Don't use x509 common names

johnSchnake:
Use Subject Alternate Names fields instead.

Fixes: #1227


johnSchnake:
FYI I just tried to remove the circleci integeation since I moved us to github actions; thats the only test failing.

johnSchnake:
I don't have a great understanding of the spec but just tried to swap out which field we used. Seems to be sufficient? I am only surprised since it worked on my first try :)

wilsonehusin:
lol i recently did this and didn't even realize this is the fix we need in Sonobuoy 😆 
https://github.com/wilsonehusin/confiar/blob/b91d2a4d8296afb7f23074bf59a3c05f88ff745c/internal/cryptographer/gostd.go#L85

thanks John!

wilsonehusin (approved):

PR #1255, opened 2021-05-10
Windows builds in CI

### LINKED ISSUE #732: Support for Windows nodes

johnSchnake:
Adds windows builds for a number of OS versions to our CI pipeline.

Utilizes buildx and skopeo to build the windows images on the linux
runner and move them to dockerhub, then uses docker manifest to
create the manifest as usual.

In a quick reversal of a previous commit, we actually HAVE to put
all the images as "sonobuoy" and have the different arch/os use
different tags instead of different image names. This has to do with
how the manifest command looks up images and prevents issues wsuch as
"manifest not found" even though it does (as a differently named image
from the others).

Fixes #732

**Release note**:
```
Windows images are now built and published alongside the Linux/amd64 and Linux/am64 versions.
```

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -17,11 +17,30 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      -
+        name: Inspect builder
+        run: |
+          echo "Name:      ${{ steps.buildx.outputs.name }}"
+          echo "Endpoint:  ${{ steps.buildx.outputs.endpoint }}"
+          echo "Status:    ${{ steps.buildx.outputs.status }}"
+          echo "Flags:     ${{ steps.buildx.outputs.flags }}"
+          echo "Platforms: ${{ steps.buildx.outputs.platforms }}"
       - name: Build binaries and images
         run: |
           source ./scripts/build.sh
           build_binaries
           linux_containers
+          windows_containers
           build_test_image
       - name: Save images to tar
         run: |
@@ -30,12 +49,9 @@ jobs:
       - name: Save artifacts
         uses: actions/upload-artifact@v2
         with:
-          name: sonobuoy-binaries-images-${{ github.run_id }}
+          name: sonobuoy-build-${{ github.run_id }}
           path: |
-            build/linux/amd64/sonobuoy
-            build/linux/arm64/sonobuoy
-            build/windows/amd64/sonobuoy.exe
-            sonobuoyImages.tar
+            build
   stress-test-linux:
     runs-on: ubuntu-latest
     steps:
@@ -58,7 +74,8 @@ jobs:
       - name: Download binaries and prebuilt images
         uses: actions/download-artifact@v2
         with:
-          name: sonobuoy-binaries-images-${{ github.run_id }}
+          name: sonobuoy-build-${{ github.run_id }}
+          path: build
       - name: Ensure binaries are executable
         run: |
           chmod +x build/linux/amd64/sonobuoy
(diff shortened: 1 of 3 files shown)


wilsonehusin:
Thinking out loud here -- before we branch out to multiple OS and Arch to maintain, should we reorganize the Dockerfiles into directories instead of being at top-level? I'm thinking something like the following, although we can further clean up the details

```
github.com/vmware-tanzu/sonobuoy
└── build
    ├── linux
    │   ├── amd64
    │   │   └── sonobuoy
    │   ├── arm64
    │   │   └── sonobuoy
    │   └── Dockerfile
    └── windows
        ├── amd64
        │   └── sonobuoy
        └── Dockerfile
```

.github/workflows/ci-test.yaml
```diff
@@ -70,7 +87,9 @@ jobs:
       - name: Build test image, load images, and run tests
         run: |
           source ./scripts/build.sh
```
wilsonehusin, line 89: nit: I'd put `source` next to `load_test_images_into_cluster` -- initially I thought this `source` was unused but turns out it is down there 😄

.github/workflows/ci-test.yaml
```diff
@@ -82,17 +101,25 @@ jobs:
...
         with:
-          name: sonobuoy-binaries-images-${{ github.run_id }}
+          name: sonobuoy-build-${{ github.run_id }}
+          path: build
+          # TODO: Load only what is needed
       - name: Load images and verify
         run: |
-          docker load -i sonobuoyImages.tar
-          docker images
+          source ./scripts/build.sh
```
wilsonehusin, line 109: is this line of `source` used?

scripts/build_funcs.sh
```diff
@@ -73,36 +75,46 @@ vet() {
...
-			build_container_dockerfile_image "DockerfileWindows-$arch" sonobuoy
-			build_container_dockerfile_image "DockerfileWindows-$arch" "sonobuoy-win-$arch"
+    elif [ "$1" = "windows" ]; then
+        if [ "$2" = "amd64" ]; then
+            # Onlhy doing one arch so this could be hardcoded, likewise we could handle the
+            # base image differently. Wanted something here for parity with linux in case we expand it though.
+			sed -e 's|BINARY|build/windows/amd64/sonobuoy.exe|g' DockerfileWindows > "DockerfileWindows-$2"
 		else 
 			echo "Windows ARCH unknown"
         fi
```
wilsonehusin, line 120: indentation seems off?

johnSchnake: So locally it looks good but I found that there is a mix of tabs/spaces in the file. Made the file consistent so the UI here should look better on the next push.

scripts/build_funcs.sh
```diff
@@ -23,32 +25,32 @@ GIT_VERSION=$(git describe --always --dirty --tags)
...
 IMAGE_BRANCH=$(git rev-parse --abbrev-ref HEAD | sed 's/\///g')
-GIT_REF_SHORT=$(git rev-parse --short=8 --verify HEAD)
 GIT_REF_LONG=$(git rev-parse --verify HEAD)
 
 BUILDMNT=/go/src/$GOTARGET
 BUILD_IMAGE=golang:1.16
 AMD_IMAGE=gcr.io/distroless/static:nonroot
 ARM_IMAGE=gcr.io/distroless/static:nonroot-arm64
-WIN_IMAGE=mcr.microsoft.com/windows/servercore:1809
+WIN_AMD64_BASEIMAGE=mcr.microsoft.com/windows/nanoserver
```
johnSchnake, line 34: Not sure if this matters. @perithompson do you have an opinion? In your blog post I saw both and I just went with nanoserver for size alone.

scripts/build_funcs.sh
```diff
@@ -174,83 +190,78 @@ native() {
...
 # in a pristine state.
 clean() {
     # Best effort for clean; don't exit if failure.
     set +e
-	rm -f "$TARGET"
-	rm Dockerfile-*
-	rm DockerfileWindows-*
-	rm -rf build
-
-	for arch in "${LINUX_ARCH[@]}"; do
```
johnSchnake, line 248: These should probably still be in place; the images are persisted as tarballs but they could be loaded locally and cause issues I presume. Going to add them back.

johnSchnake: Dur. Nm. Removing the image $TARGET is removing them all now. Thats why I removed them.

PR #1257, opened 2021-05-10
Refactor for explicit entrypoint

wilsonehusin:
With Kubernetes moving to use entrypoint when invoking conformance test,
we should follow with using entrypoint mechanism to retrieve list of
images.

**Which issue(s) this PR fixes**:
None, preemptive refactor before needing a fix 😄  

This is due to change 

**Release note**:
```
None
```


johnSchnake:
The one concern about this is backwards compatibility. If someone uses the new sonobuoy for a slightly older cluster (we explicitly support 3 versions at a time) then there would be a problem.

Just making sure; in the linked PR you said the image would have the existing script/gorunner removed in 1.25? So that gives us plenty of time and we shouldnt have a problem right?

We just wouldn't want to merge this until 2-3 versions from now or else we need to split our logic based on the version, right?

wilsonehusin:
> If someone uses the new sonobuoy for a slightly older cluster (we explicitly support 3 versions at a time) then there would be a problem.

You're right. I think this PR needs to be split into two.

First is the change to explicitly call `e2e.test` for `--list-images`. This should be done now and should not have any side effects between past versions and future versions. In fact, if we don't do this now, Sonobuoy won't work with Kubernetes 1.22 because entrypoint has been defined.

```diff
- e2eImages, err := client.RunImage(conformanceImage, "e2e.test", "--list-images")
+ e2eImages, err := client.RunImage("e2e.test", conformanceImage, "--list-images")
```

The other is changes to Kubernetes manifest, which should happen later, like you suggested, once we find ourselves with 1.22 being the oldest release we support.

```diff
- Command:         []string{"/run_e2e.sh"},
...
-     {Name: "E2E_USE_GO_RUNNER", Value: "true"},
```

pkg/image/docker_client_test.go
```diff
@@ -34,8 +34,8 @@ type FakeDockerClient struct {
 	deleteFails bool
 }
 
-func (l FakeDockerClient) Run(image string, args ...string) ([]string, error) {
-	return l.Run(image, args...)
+func (l FakeDockerClient) Run(entrypoint string, image string, args ...string) ([]string, error) {
+	return l.Run(entrypoint, image, args...)
```
vladimirvivien, line 38: Is this a recursive call into l.Run() ?

vladimirvivien (commented):
Looks good, but left couple comments.

pkg/image/docker_client_test.go
wilsonehusin: 🤔 I think it is? I was doing global find and replace but now maybe this code path isn't being traversed at all that it doesn't error out at all.

PR #1259, opened 2021-05-11
Run integration tests on Windows

johnSchnake:
Runs the same integration suite on Windows.
This is one way we can ensure Windows clients can process
results in the same way.


johnSchnake:
Closing; will consider what needs to be done on Windows but since automation of windows/mixed clusters in CI like github is a big hurdle we will probably just make a script to be run locally to run the tests on a periodic basis.

PR #1260, opened 2021-05-12
Removing unnecessary and incorrect load of testimage

johnSchnake:

Just didn't catch this since it was in the push job, run only against master/tags.

Just confirmed this fix worked on my own fork though after merging: https://github.com/johnSchnake/sonobuoy/actions/runs/835797345

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -106,7 +106,6 @@ jobs:
           # TODO: Load only what is needed
       - name: Load images and verify
         run: |
-          docker load -i build/testimage.tar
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-test.yaml
```diff
@@ -106,7 +106,6 @@ jobs:
           # TODO: Load only what is needed
       - name: Load images and verify
         run: |
-          docker load -i build/testimage.tar
```
johnSchnake, line 109: Added run-id to all the test image names and just forgot this one, hence the error in the last push from master.

johnSchnake:
Based on the error being in the push job, the CI here won't hit it anyways. I also tested the fix/push on my fork.

Going to just merge w/o review so we can get the windows images pushed.

PR #1267, opened 2021-05-18
Fix multiple Windows tar issues 

johnSchnake:
- When on the server side handling files, use filepath
- When loading files into the tarball, use path
- When on the clientside use filepath
- Use our own tar logic to avoid tarx and bugs with path/filepath
- Use splat on Windows as we do for linux. TODO (not on this PR): avoid reading entire tar into memory, for large runs that will cause an issue I'm pretty sure. Tarballs can get very big when you have 100 nodes and run tests for hours.

### DIFF
--- a/pkg/client/results/reader.go
+++ b/pkg/client/results/reader.go
@@ -23,12 +23,13 @@ import (
 	"encoding/json"
 	"encoding/xml"
 	"fmt"
-	"gopkg.in/yaml.v3"
 	"io"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
+
+	"gopkg.in/yaml.v3"
 
 	"github.com/vmware-tanzu/sonobuoy/pkg/config"
 
@@ -210,7 +211,7 @@ func (r *Reader) WalkFiles(walkfn filepath.WalkFunc) error {
 			header.FileInfo(),
 			tr,
 		}
-		err = walkfn(filepath.Clean(header.Name), info, err)
+		err = walkfn(path.Clean(header.Name), info, err)
 	}
 
 	if err == errStopWalk || err == io.EOF {
@@ -290,7 +291,7 @@ func (r *Reader) Metadata() string {
 func (r *Reader) ServerVersionFile() string {
 	switch r.Version {
 	case VersionEight:
-		return "serverversion/serverversion.json"
+		return path.Join("serverversion", "serverversion.json")
 	default:
 		return defaultServerVersionFile
 	}
@@ -306,7 +307,7 @@ func (r *Reader) NamespacedResources() string {
 func (r *Reader) NonNamespacedResources() string {
 	switch r.Version {
 	case VersionEight:
-		return "resources/non-ns/"
+		return path.Join("resources", "non-ns")
 	default:
 		return nonNamespacedResourcesDir
 	}
@@ -315,7 +316,7 @@ func (r *Reader) NonNamespacedResources() string {
 // NodesFile returns the path to the file that lists the nodes of the Kubernetes
 // cluster.
 func (r *Reader) NodesFile() string {
-	return filepath.Join(r.NonNamespacedResources(), defaultNodesFile)
+	return path.Join(r.NonNamespacedResources(), defaultNodesFile)
 }
 
 // ServerGroupsFile returns the path to the groups the Kubernetes API supported at the time of the run.
@@ -330,7 +331,7 @@ func ConfigFile(version string) string {
 	case VersionEight:
 		return "config.json"
 	default:
-		return "meta/config.json"
+		return path.Join("meta", "config.json")
 	}
 }
 
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -24,7 +24,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"time"
 
@@ -34,11 +33,11 @@ import (
 	"github.com/vmware-tanzu/sonobuoy/pkg/errlog"
 	"github.com/vmware-tanzu/sonobuoy/pkg/plugin"
 	pluginaggregation "github.com/vmware-tanzu/sonobuoy/pkg/plugin/aggregation"
+	"github.com/vmware-tanzu/sonobuoy/pkg/tarball"
 
 	"github.com/pkg/errors"
 	"github.com/rifflock/lfshook"
 	"github.com/sirupsen/logrus"
-	"github.com/viniciuschiele/tarx"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -81,8 +80,8 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 	// 1. Create the directory which will store the results, including the
 	// `meta` directory inside it (which we always need regardless of
 	// config)
-	outpath := path.Join(cfg.ResultsDir, cfg.UUID)
-	metapath := path.Join(outpath, MetaLocation)
+	outpath := filepath.Join(cfg.ResultsDir, cfg.UUID)
+	metapath := filepath.Join(outpath, MetaLocation)
 	err = os.MkdirAll(metapath, 0755)
 	if err != nil {
 		errlog.LogError(errors.Wrap(err, "could not create directory to store results"))
@@ -92,7 +91,7 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 	// Write logs to the configured results location. All log levels
 	// should write to the same log file
 	pathmap := make(lfshook.PathMap)
-	logfile := path.Join(metapath, "run.log")
+	logfile := filepath.Join(metapath, "run.log")
 	for _, level := range logrus.AllLevels {
 		pathmap[level] = logfile
 	}
@@ -135,7 +134,7 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 
 	// 3. Dump the config.json we used to run our test
 	if blob, err := json.Marshal(cfg); err == nil {
-		if err = ioutil.WriteFile(path.Join(metapath, "config.json"), blob, 0644); err != nil {
+		if err = ioutil.WriteFile(filepath.Join(metapath, "config.json"), blob, 0644); err != nil {
 			errlog.LogError(errors.Wrap(err, "could not write config.json file"))
 			return errCount + 1
 		}
@@ -209,7 +208,7 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
 
 	// 6. Dump the query times
 	trackErrorsFor("recording query times")(
-		recorder.DumpQueryData(path.Join(metapath, "query-time.json")),
+		recorder.DumpQueryData(filepath.Join(metapath, "query-time.json")),
 	)
 
 	// 7. Clean up after the plugins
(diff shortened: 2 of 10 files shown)


pkg/tarball/tarball.go
```diff
@@ -83,3 +85,76 @@ func DecodeTarball(reader io.Reader, baseDir string) error {
 
 	return nil
 }
+
+// TarDir tars up an entire directory and outputs the tarball to the specified output path.
+// If useGzip is true, it also gzips the resulting tarball.
+func TarDir(dir, outpath string, useGzip bool) error {
```
wilsonehusin, line 91: I probably should've done this instead of implementing whatever I wrote in https://github.com/vmware-tanzu/sonobuoy/blob/master/cmd/sonobuoy/app/splat.go 😅

wilsonehusin: nit: I prefer the name `DirToTarball` because `TarDir` feels like it can go either way of tar _to_ dir or tar _from_ dir (or is it second-language thing?)

pkg/tarball/tarball.go
```diff
@@ -83,3 +85,76 @@ func DecodeTarball(reader io.Reader, baseDir string) error {
...
+		header, err := tar.FileInfoHeader(fi, fi.Name())
+		if err != nil {
+			return errors.Wrapf(err, "creating file info header %v", fi.Name())
+		}
+
+		// Update the name to correctly reflect the desired destination when untaring.
+		// 1. Remove directory prefix and leading /
+		// 2: Store with Unix path seperators
+		// 3: Clean
+		header.Name = strings.TrimPrefix(path.Clean(filepath.ToSlash(strings.Replace(file, dir, "", -1))), "/")
```
wilsonehusin, line 140: 🤔 isn't this covered by [`path.Base`](https://golang.org/pkg/path/#Base)?

pkg/client/results/reader.go
```diff
@@ -330,7 +331,7 @@ func ConfigFile(version string) string {
 	case VersionEight:
 		return "config.json"
 	default:
-		return "meta/config.json"
+		return path.Join("meta", "config.json")
```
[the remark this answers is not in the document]
johnSchnake: So this is one of the things that wasn't well flushed out because it didn't have to be in the past since the server was always on linux nodes.

The git commit mentions the logic here:
 - on the client and server, when dealing with their own files you need to use filepath
 - when doing anything with files in the tarball, use path.

This way all local file handling works just fine (filepath) but the files in the tarball are in a predictable format (path) for traversing. Otherwise it gets messy to (1) record/detect what OS was responsible for making the tar (2) modifying which path separator to use when traversing

pkg/discovery/discovery.go
```diff
@@ -245,13 +244,13 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
...
-		err = ioutil.WriteFile(path.Join(metapath, results.InfoFile), blob, 0644)
+		err = ioutil.WriteFile(filepath.Join(metapath, results.InfoFile), blob, 0644)
 		trackErrorsFor("saving" + results.InfoFile)(err)
 	}
 
 	// 8. tarball up results YYYYMMDDHHMM_sonobuoy_UID.tar.gz
-	tb := cfg.ResultsDir + "/" + t.Format("200601021504") + "_sonobuoy_" + cfg.UUID + ".tar.gz"
-	err = tarx.Compress(tb, outpath, &tarx.CompressOptions{Compression: tarx.Gzip})
+	tb := filepath.Join(cfg.ResultsDir, t.Format("200601021504")+"_sonobuoy_"+cfg.UUID+".tar.gz")
+	err = tarball.TarDir(outpath, tb, true)
```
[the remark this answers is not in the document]
johnSchnake: Which seemed odd; I always assumed it was a very common library but it has ~0 stars, 1 follower, 1 contributor, 27 commits, and hasnt been updated in 5y :)

pkg/tarball/tarball.go
johnSchnake: /shrug, its fine either way but the 'DirToTarball' is even more explicit. I can change that.

wilsonehusin:
> Use splat on Windows as we do for linux. TODO (not on this PR): avoid reading entire tar into memory, for large runs that will cause an issue I'm pretty sure. Tarballs can get very big when you have 100 nodes and run tests for hours.

Good catch, I did that without too much thought because I wanted to get distroless out before another CVE report comes in 😅

pkg/tarball/tarball.go
johnSchnake: No, we don't want just the base name but we want to make them just relative to the root of the directory.

So the filepath is effectively \tmp\sonobuoy\dir\foo\bar
Trimming the dir we get \foo\bar
Changing to slashes we get /foo/bar
Remove leading prefix we get foo/bar

pkg/discovery/discovery.go
```diff
@@ -245,13 +244,13 @@ func Run(restConf *rest.Config, cfg *config.Config) (errCount int) {
...
 	if err == nil {
-		err = ioutil.WriteFile(path.Join(metapath, results.InfoFile), blob, 0644)
+		err = ioutil.WriteFile(filepath.Join(metapath, results.InfoFile), blob, 0644)
 		trackErrorsFor("saving" + results.InfoFile)(err)
 	}
 
 	// 8. tarball up results YYYYMMDDHHMM_sonobuoy_UID.tar.gz
-	tb := cfg.ResultsDir + "/" + t.Format("200601021504") + "_sonobuoy_" + cfg.UUID + ".tar.gz"
-	err = tarx.Compress(tb, outpath, &tarx.CompressOptions{Compression: tarx.Gzip})
+	tb := filepath.Join(cfg.ResultsDir, t.Format("200601021504")+"_sonobuoy_"+cfg.UUID+".tar.gz")
```
vladimirvivien, line 252: suggestion: use fmt.Sprintf() instead of string concatenation

pkg/tarball/tarball.go
```diff
@@ -83,3 +85,76 @@ func DecodeTarball(reader io.Reader, baseDir string) error {
...
+// TarDir tars up an entire directory and outputs the tarball to the specified output path.
+// If useGzip is true, it also gzips the resulting tarball.
+func TarDir(dir, outpath string, useGzip bool) error {
+	// ensure the src actually exists before trying to tar it
+	if _, err := os.Stat(dir); err != nil {
+		return errors.Wrapf(err, "tar unable to stat directory %v", dir)
+	}
+
+	outfile, err := os.Create(outpath)
+	if err != nil {
```
vladimirvivien, line 98: Should we check for the type of err (if path already exists, for instance,  maybe just overwrite).

pkg/tarball/tarball.go
```diff
@@ -83,3 +85,76 @@ func DecodeTarball(reader io.Reader, baseDir string) error {
...
+		// 1. Remove directory prefix and leading /
+		// 2: Store with Unix path seperators
+		// 3: Clean
+		header.Name = strings.TrimPrefix(path.Clean(filepath.ToSlash(strings.Replace(file, dir, "", -1))), "/")
+		if err := tw.WriteHeader(header); err != nil {
+			return errors.Wrapf(err, "writing header for tarball %v", header.Name)
+		}
+
+		// Only copy contents of regular files
+		if !fi.Mode().IsRegular() {
```
vladimirvivien, line 146: Seems like line 125 already does that check.

pkg/tarball/tarball.go
vladimirvivien: I think the following may also do the trick (figuring out relative path) automatically for you
```go
relFilePath := file
if filepath.IsAbs(path) {
    relFilePath, err = filepath.Rel("/", file)
        if err != nil {
	    return err
        }
}
```

test/integration/testImage/src/tarResults.go
```diff
@@ -64,3 +68,78 @@ func reportTarFile(cmd *cobra.Command, args []string) error {
 	err = ioutil.WriteFile(doneFile, []byte(tb), os.FileMode(0666))
 	return errors.Wrap(err, "failed to write to done file")
 }
+
+// tarDir tars up an entire directory and outputs the tarball to the specified output path.
+// If useGzip is true, it also gzips the resulting tarball.
+// NOTE: Copied from sonobuoy itself; duplicated since this is a separate binary/project
+// so it messed with the build when introducing this method the first time.
+func tarDir(dir, outpath string, useGzip bool) error {
```
vladimirvivien, line 76: Would it be better to make `tarDir` exportable in its own package as `TarDir` then reuse it in both places?

vladimirvivien (commented):
Left some comments. But overall looks good to me.

pkg/tarball/tarball.go
johnSchnake: I would rather avoid introducing an error path if possible, I know its not an elegant line but 0 errors and a reasonable comment I think is better in this particular case. Let me know if you disagree.

pkg/client/results/reader.go
```diff
@@ -290,7 +291,7 @@ func (r *Reader) Metadata() string {
 func (r *Reader) ServerVersionFile() string {
 	switch r.Version {
 	case VersionEight:
-		return "serverversion/serverversion.json"
+		return path.Join("serverversion", "serverversion.json")
```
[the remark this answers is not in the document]
johnSchnake: Per your other comment, resolving this.

pkg/client/results/reader.go
```diff
@@ -306,7 +307,7 @@ func (r *Reader) NamespacedResources() string {
 func (r *Reader) NonNamespacedResources() string {
 	switch r.Version {
 	case VersionEight:
-		return "resources/non-ns/"
+		return path.Join("resources", "non-ns")
```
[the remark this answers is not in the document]
johnSchnake: Per other comment; resolving this.

pkg/client/results/reader.go
```diff
@@ -315,7 +316,7 @@ func (r *Reader) NonNamespacedResources() string {
 // NodesFile returns the path to the file that lists the nodes of the Kubernetes
 // cluster.
 func (r *Reader) NodesFile() string {
-	return filepath.Join(r.NonNamespacedResources(), defaultNodesFile)
+	return path.Join(r.NonNamespacedResources(), defaultNodesFile)
```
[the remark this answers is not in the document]
johnSchnake: Per other comment; resolving this.

pkg/tarball/tarball.go
johnSchnake: Its on the server and is based on a UUID of the run; there really isn't a reasonable way that this would error now besides a crazy os issue. I don't think extra work on that path is necessary.

pkg/tarball/tarball.go
johnSchnake: So the first gate allows directories or regular files so that we write headers. This check prevents trying to `os.Copy` on directories which causes an error.

test/integration/testImage/src/tarResults.go
johnSchnake: So it would have to be outside of sonobuoy entirely and outside of testdir.

I was tempted to make another repo entirely in my own github name but just thought this was easier for now. It does introduce an issue with keeping them in sync though.

Would you prefer if I made a tiny repo for tar/untar? Effectively a super simple version of tarx?

vladimirvivien: @johnSchnake maybe I misunderstood the issue. But it seems that if sonobuoy repo is viewed as a monorepo, but generates multiple binaries, you should be able to reuse the `TarDir` (from it's own package).   

Unless `test` is treated as a completely separate Go project, in that case it should have its own go.mod, then it can import a package with the `TarDir` (I think you can have go.mod in subpackages, could be wrong). 

But please set me straight if I misunderstand the issue.

vladimirvivien (commented):
Left a short comment. But other than than we should be good.

test/integration/testImage/src/tarResults.go
johnSchnake: No, thats pretty much the issue but there is just a problem when you simultaneously try and change the sonobuoy repo and the test repo at the same time. So the test binary, when getting built says "oh I need to go get sonobuoy for this tar function..." but then since that isn't anywhere other than my branch/fork it can't find it. So we can depend on sonobuoy just fine, but it would always need to be looking back at older commits. So if we get rid of tarx in sonobuoy and commit it, then have a separate commit where we update the test image to leverage that it would be fine.

We could do that: split it into two commits. We can just copy it here for this commit and follow up with just referencing sonobuoys. That way it is always in sync and after just this one commit I'll follow up and then testimage can depend directly on sonobuoy.

johnSchnake:
One failure and it is for a symlink issue; can someone else take a look and confirm I resolved it and this is now a false negative? I added an ultra-conservative fix but I think the codeQL logic just doesn't pick it up. Their suggested logic doesn't fix because the symlink path doesn't actually exist so evalSymlink fails. We are testing a hypothetical link and so I just prevent any path traversal.

vladimirvivien (approved):
One more suggestion 
Maybe do a relative import from within the test package to point to the location of TarDir?
Anyway, we can get this in now and think of a better/simpler way of solving this later.

johnSchnake:
> One more suggestion
> Maybe do a relative import from within the test package to point to the location of TarDir?
> Anyway, we can get this in now and think of a better/simpler way of solving this later.

After this PR I'll immediately put up another to reference it. Your description of it as another project entirely made the flow much more clear to me; if we depend on XYZ and want a change we just have to order those changes as needed.

PR #1270, opened 2021-05-18
Extend default timeout

### LINKED ISSUE #1269: Extend default timeout
Certain configurations (reasonable ones) like Windows or ones that include optional tests/etc may take longer to complete. We can afford to increase the default timeout so that users are not confused about the timeout failure.

This is slightly kicking the can, because it would be nice to have smarter logic about 'if seeing progress, dont timeout until some crazy limit like 24h' but this is a 1 line change and solves 80% of cases.

The original need for timeout was to avoid hanging in CI forever but I think numerous improvements make that less common/likely.

xref #1268

johnSchnake:
Doubles default server timeout. Will help avoid some user issues as
well as preempting some Windows issues. Only cost is increased downside
when truly hanging in a way that Sonobuoy doesn't detect (not common I
think).

Fixes #1269


vladimirvivien (approved):
LGTM

PR #1278, opened 2021-05-25
Add node selector option when generating plugin spec

### LINKED ISSUE #1275: Sonobuoy gen plugin should add node selector if desired
A simple flag when creating a plugin spec so that you can specify `--node-selector=kubernetes.io/os:windows` or the like. Maybe even just an `--windows` flag to make it dead simple and it would add that (and the arch one too?)

Default to linux if no podSpec is present. Discussed in linked issue.

johnSchnake:
By default jobs can run on Linux nodes (thats fine) but
if we expect users to specify the node selectors for
Windows we should make it simple to add that bit.

Fixes #1275

**Release note**:
```
Adds a new flag: sonobuoy gen plugin --node-selector foo=bar to facilitate creating windows plugins more easily.
```

### DIFF
--- a/cmd/sonobuoy/app/gen_plugin_def.go
+++ b/cmd/sonobuoy/app/gen_plugin_def.go
@@ -53,6 +53,11 @@ type GenPluginDefConfig struct {
 	// Allows validation during flag parsing by having a custom type.
 	driver pluginDriver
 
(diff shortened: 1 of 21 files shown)


vladimirvivien (approved):
LGTM

PR #1280, opened 2021-06-02
Remove coloring for codeblocks in blog

johnSchnake:
The darkgrey was chosen, presumablu, because small bits
of code look good. However, in a large code block the background
is black and so the dark grey is nearly invisible.

Changing it to lightgrey or the like makes the blocks look good
but not the short snippets.

Rather than modify the blocks separately, removing this property
in general makes both look good in my opinion.

### DIFF
--- a/site/themes/sonobuoy/assets/scss/_components.scss
+++ b/site/themes/sonobuoy/assets/scss/_components.scss
@@ -351,7 +351,6 @@
         }
         code {
             border: 2px solid #EFEFEF;
-            color: $darkgrey;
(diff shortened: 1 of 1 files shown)


johnSchnake:
Compare https://sonobuoy.io/simplified-results-reporting-with-sonobuoy/ (current) to this change https://deploy-preview-1280--sonobuoy.netlify.app/simplified-results-reporting-with-sonobuoy/

FYI @jonasrosland to be aware for any other sites you manage.

Tiny change so I'm just going to merge.

PR #1284, opened 2021-06-03
Fix nodeSelector logic

johnSchnake:
The logic for this was previously broken, but when fixed, it
was found to actually be implemented incorrectly causing it to
overwrite the nodeSelector in all cases, not just when the podSpec
was empty.

Now it just updates the nodeSelector to linux boxes when there is
no podspec given by the user. If they want to specify a podspec to
force it to a different os (or any os) then they can do so.

Tests added to prevent this issue from regression.

### DIFF
--- a/cmd/sonobuoy/app/testdata/pluginDef-default-podspec.golden
+++ b/cmd/sonobuoy/app/testdata/pluginDef-default-podspec.golden
@@ -1,5 +1,7 @@
 podSpec:
   containers: []
+  nodeSelector:
+    kubernetes.io/os: linux
   restartPolicy: Never
   serviceAccountName: sonobuoy-serviceaccount
   tolerations:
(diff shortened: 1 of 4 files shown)


johnSchnake:
FWIW: I didnt like these tests as-is and thought they should just use golden files to check the whole pod spec generated. However, avoiding import cycles is difficult and as you try and untangle things it just becomes more of a mess. I think we just hold our nose and keep them as-is. I think Bridget had the same experience I did: she added the podspec stuff and after fighting the code organization for a while just settled on copying the frustrating/verbose pattern and moving on.

johnSchnake:
Also, this is a hard issue to find/debug locally. The plugin will look fine, the sonobuoy gen output looks good, but when you actually run the plugin this was injected at the last second.

I'm going to create a ticket to clean that up at some point and make a more simple way to test from plugin file X what gets sent to k8s. May be a lot of refactoring (as I noted it was a tangled mess in spots) but in the end we will get more clarity. This bug caused me to waste much too much time today so it I'd like to avoid it in the future.

wilsonehusin (approved):

PR #1295, opened 2021-06-10
Use embed to improve readability

johnSchnake:
Using the new embed package here allows the IDE to use
language aware coloring/tools/etc. Also makes it easy
to swap copy the file and swap out with a different one
for dev/testing.

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -5,14 +5,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-go@v2
-        with:
-          go-version: 1.15
       - name: Run unit tests
         env:
           VERBOSE: true
         run: |
-          source ./scripts/build.sh; unit_local
+          source ./scripts/build.sh; unit
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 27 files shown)


.github/workflows/ci-test.yaml
```diff
@@ -5,14 +5,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-go@v2
-        with:
-          go-version: 1.15
```
johnSchnake, line 10: This was bypassing the go version we had in our build scripts since it was trying to run local unit tests instead of dockerized. Local is really just for dev

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/vmware-tanzu/sonobuoy
 
-go 1.14
```
johnSchnake, line 3: Gets us embed

pkg/client/testdata/custom-systemd-logs-image.golden
```diff
@@ -1,4 +1,3 @@
-
```
johnSchnake, line 1: All these were just artifacts of how the string was defined before;  now there isn't a need for the random empty line at the start.

johnSchnake:
I think the go mod tidy failure is just from the different versions of go being used here. When I run it locally it has no issue and doesn't modify the go.sum/go.mod files.

johnSchnake:
Going to merge based on positive comments so far. This is somewhat blocking since another PR of mine wanted/needed to go to this version of go too.

PR #1296, opened 2021-06-11
Enables plugins to specify config map information

johnSchnake:
The most common configuration option that we don't easily support is
a plugin specification that also includes config map info. By supporting
this we can easily support different deployment configurations without
specifying the entire `sonobuoy gen` yaml manually and sharing that.

This is particularly useful for specifying the kube-image-repo-list
as part of the plugin itself rather than as a command line option that
makes it harder to share.

### DIFF
--- a/cmd/sonobuoy/app/gen_plugin_def.go
+++ b/cmd/sonobuoy/app/gen_plugin_def.go
@@ -19,6 +19,7 @@ package app
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/vmware-tanzu/sonobuoy/pkg/client"
 	"github.com/vmware-tanzu/sonobuoy/pkg/client/results"
@@ -60,6 +61,9 @@ type GenPluginDefConfig struct {
 
 	// If set, the default pod spec used by Sonobuoy will be included in the output
 	showDefaultPodSpec bool
+
+	// configMapFiles is the list of files to read/store as configmaps for the plugin.
+	configMapFiles []string
 }
 
 // NewCmdGenPluginDef ...
@@ -117,6 +121,11 @@ func NewCmdGenPluginDef() *cobra.Command {
 	genPluginSet.StringArrayVarP(
 		&genPluginOpts.def.Spec.Args, "arg", "a", []string{},
 		"Arg values to set on the plugin. Can be set multiple times (e.g. --arg 'arg 1' --arg arg2)",
+	)
+
+	genPluginSet.StringArrayVar(
+		&genPluginOpts.configMapFiles, "configmap", nil,
+		`Specifies files to read and add as configMaps. Will be mounted to the plugin at /tmp/sonobuoy/configs/<filename>.`,
 	)
 
 	AddShowDefaultPodSpecFlag(&genPluginOpts.showDefaultPodSpec, genPluginSet)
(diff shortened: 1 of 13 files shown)


pkg/plugin/manifest/manifest.go
```diff
@@ -60,21 +60,30 @@ func (s *SonobuoyConfig) DeepCopy() *SonobuoyConfig {
...
+	m2 := &Manifest{
 		SonobuoyConfig: *m.SonobuoyConfig.DeepCopy(),
 		Spec:           *m.Spec.DeepCopy(),
 		PodSpec:        m.PodSpec.DeepCopy(),
 		objectKind:     objectKind{m.gvk},
 	}
+	if m.Configs != nil {
+		m2.Configs = map[string]string{}
+		for k, v := range m.Configs {
+			m2.Configs[k] = v
```
johnSchnake, line 83: Need to range over the inner values too; its a map of maps.

pkg/plugin/manifest/manifest.go
```diff
@@ -60,21 +60,30 @@ func (s *SonobuoyConfig) DeepCopy() *SonobuoyConfig {
...
 type Manifest struct {
-	SonobuoyConfig SonobuoyConfig `json:"sonobuoy-config"`
-	Spec           Container      `json:"spec"`
-	ExtraVolumes   []Volume       `json:"extra-volumes,omitempty"`
-	PodSpec        *PodSpec       `json:"podSpec,omitempty"`
+	SonobuoyConfig SonobuoyConfig    `json:"sonobuoy-config"`
+	Spec           Container         `json:"spec"`
+	ExtraVolumes   []Volume          `json:"extra-volumes,omitempty"`
+	PodSpec        *PodSpec          `json:"podSpec,omitempty"`
+	Configs        map[string]string `json:"configs,omitempty"`
```
johnSchnake, line 67: Ensure I'm consistent about config/configs; unless another word entirely works better. Does get confusing because of the `sonobuoyConfig`

pkg/plugin/manifest/manifest.go
johnSchnake: False; this one isnt

pkg/plugin/manifest/manifest.go
johnSchnake: Using ConfigMap instead; I think thats more clear even if not completely accurate

pkg/client/gen_test.go
```diff
@@ -402,13 +402,27 @@ func TestGenerateManifestGolden(t *testing.T) {
 			},
 			goldenFile: filepath.Join("testdata", "single-node-selector.golden"),
 		}, {
-			name: "Multiple node selectors can be added",
```
johnSchnake, line 405: Never meant to edit this test; be sure to add it back

cmd/sonobuoy/app/gen_plugin_def.go
```diff
@@ -186,6 +195,18 @@ func genPluginDef(cfg *GenPluginDefConfig) ([]byte, error) {
...
+	if len(cfg.configMapFiles) > 0 {
+		cfg.def.ConfigMap = map[string]string{}
+	}
+	for _, v := range cfg.configMapFiles {
+		fData, err := os.ReadFile(v)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to read file %q", v)
+		}
+		base := filepath.Base(v)
+		cfg.def.ConfigMap[base] = string(fData)
```
vladimirvivien, line 207: Possibility of overwriting file content if `base` happens to be same name from multiple locations.

vladimirvivien (commented):
This looks good.

cmd/sonobuoy/app/gen_plugin_def.go
johnSchnake: True but wouldnt that be the case for all configmaps? This is for a particular plugin and each plugins configmaps are put into separate objects, so it shouldn't conflict between plugins.

johnSchnake:
Rebased; merging after green.

scripts/build_funcs.sh
```diff
@@ -60,7 +60,7 @@ integration() {
         --env SONOBUOY_CLI="$SONOBUOY_CLI" \
         --network host \
         "$BUILD_IMAGE" \
-    go test ${VERBOSE:+-v} -timeout 3m -tags=integration "$GOTARGET"/test/integration/...
+    go test ${VERBOSE:+-v} -timeout 3m -tags=integration "$GOTARGET"/test/integration/... -run TestConfigmaps
```
johnSchnake, line 63: This line hid a broken build by mistake. Fixing...

PR #1298, opened 2021-06-15
Split windows and linux builds for speed

johnSchnake:
Waiting for the windows images to build each time when iterating
on tests/features is painful. We really only need the windows
images if we are sure things are passing and someone may actually
want to download/use/test the windows images.


johnSchnake:
You can see in the actions that it still builds for PRs so if you are iterating on windows it doesnt stop you from getting the images, it just makes sure not to waste cpu/storage by building them right away if unit/integration tests are going to fail (which only depend on linux)

vladimirvivien (approved):
LGTM

PR #1300, opened 2021-06-16
Improve daemonset restart logic

### LINKED ISSUE #1299: 409 from server shouldn't be error (sonobuoy-worker)
See https://github.com/vmware-tanzu/sonobuoy-plugins/issues/50

409 indicates the plugin already reported results. If that is the case we should probably just exit 0 rather than nonzero. Can still log it as a warning though.

johnSchnake:
- Modifies default sleep from 1h to indefinite
- Changes 409 results to log instead of error in
order to prevent daemonsets from restarting.

- Fixes #1299

**Release note**:
```
Improved daemonset logic in order to avoid unnecessary restarts and log entries.
```


vladimirvivien (approved):
LGTM

PR #1303, opened 2021-06-16
Expands timeout error message to try and clarify action

johnSchnake:
When a plugin times out, users can sometimes be confused what they
should do next. Was there a real problem? Rerun?

At the very least we need to instruct them to check the logs and
see if they can find more clues.

- Fixes # 1152

**Release note**:
```
Clarifies plugin timeout error message to instruct the user to check logs and that there are no results to view.
```


vladimirvivien (approved):
Looks good.

PR #1304, opened 2021-06-16
Removes numerous globals

johnSchnake:
Mainly globals for flag definitions replaced with variables
and making methods return closures around those variables rather
than accessing the globals.

There are other globals but this avoids one of the most common
types of bugs we've ran into and closes a long-standing issue
to fix these.

Fixes: 1052


cmd/sonobuoy/app/gen_config.go
```diff
@@ -13,10 +13,11 @@ import (
 // NewCmdGenConfig creates the `config` command which will print out
 // the default sonobuoy config in a json format.
 func NewCmdGenConfig() *cobra.Command {
+	var f genFlags
 	var GenCommand = &cobra.Command{
 		Use:   "config",
 		Short: "Generates a sonobuoy config for input to sonobuoy gen or run.",
-		Run:   genConfigCobra,
+		Run:   genConfigCobra(f),
```
vladimirvivien, line 20: Other command values use a pointer param in their helper function for Run.

cmd/sonobuoy/app/logs.go
```diff
@@ -32,58 +32,61 @@ const (
...
+		for {
+			n, err := logreader.Read(b)
+			if err != nil && err != io.EOF {
+				errlog.LogError(errors.Wrap(err, "error reading logs"))
+				os.Exit(1)
+			}
+			fmt.Fprint(logConfig.Out, string(b[:n]))
+			if err == io.EOF {
+				return
+			}
```
vladimirvivien, line 89: Consider `io.Copy(logConfig.out, logreader)` since `fmt.Fprintf` is not used for any sort of formatting.

vladimirvivien (commented):
Looks great, left couple comments

cmd/sonobuoy/app/logs.go
johnSchnake: Its just an `Fprint`, not `Fprintf` and `io.Copy` would need a reader where here its just some raw bytes.

johnSchnake:
Making change based on input and merging once green. Then moving on to fix some issues that _slightly_ touch on some of this code

PR #1309, opened 2021-06-17
Allow plugins to have variable k8s version

### LINKED ISSUE #1307: Surface k8s version and allow plugins to use that variable
Plugin definitions should be able to utilize the same "auto" type of logic that the e2e plugin does. We resolve the k8s API version then let the plugin image be relative to that.

We should allow something like a variable in the plugin specs. E.g. specify `image: myrepo/image:$k8s_version` even if it is just a hardcoded string replacement later on. We can make it more unique to sonobuoy to avoid any conflict: `$sonobuoy_k8s_version` for instance.

This way we can 1) allow other plugins to use this 2) pull the e2e plugin definition out of sonobuoy's code itself

This helps support windows e2e as well.

We can also add it automatically to the plugin env (which would also pave the way to surface other values that may be useful to the plugins)

johnSchnake:
- Allows plugins to specify image using a variable SONOBUOY_K8S_VERSION
- Exposes this as an env var on all plugins

Fixes #1307

**Release note**:
```
Plugins can specify their image using a variable $SONOBUOY_K8S_VERSION which will be resolved at runtime. This will value will also be set as the environment variable SONOBUOY_K8S_VERSION.
```


cmd/sonobuoy/app/gen.go
```diff
@@ -88,14 +95,16 @@ func GenFlagSet(cfg *genFlags, rbac RBACMode) *pflag.FlagSet {
...
-	AddKubeConformanceImageVersion(&cfg.kubeConformanceImageVersion, genset)
 	AddSSHKeyPathFlag(&cfg.sshKeyPath, genset)
 	AddSSHUserFlag(&cfg.sshUser, genset)
 
 	AddPluginSetFlag(&cfg.plugins, genset)
 	AddPluginEnvFlag(&cfg.pluginEnvs, genset)
 
 	AddNodeSelectorsFlag(&cfg.nodeSelectors, genset)
+
+	AddKubeConformanceImageVersion(&cfg.k8sVersion, genset)
```
johnSchnake, line 106: Binding bot to the same variable; not sure if that will behave correctly. Need to make sure that either sets the value.

cmd/sonobuoy/app/args.go
```diff
@@ -359,13 +362,15 @@ func AddPluginListFlag(p *[]string, flags *pflag.FlagSet) {
...
-	)
+// AddKubernetesVersionFlag initialises an image version flag.
+func AddKubernetesVersionFlag(imageVersion *image.ConformanceImageVersion, flags *pflag.FlagSet) {
+	help := "Use default Conformance image, but override the version. "
+	help += "Default is 'auto', which will be set to your cluster's version if detected, erroring otherwise. "
+	help += "'ignore' will try version resolution but ignore errors. "
+	help += "'latest' will find the latest dev image/version upstream."
+
+	*imageVersion = image.ConformanceImageVersionAuto
+	flags.Var(imageVersion, "kubernetes-version", help)
```
johnSchnake, line 373: Moving towards this flag/naming since this version is more generic than just the conformance image.

The preexisting flag should work as well but I marked it as deprecated. Need to make sure its documented well in the release.

johnSchnake:
Moved logic into the client so it is more reusable (even if someone vendors the client) and it pushes the handling to later which means we dont have to differentiate between static/dynamic plugins so it wont have to change when those distinctions go away.

Also added a static config so that we can add integration tests for `gen` which makes testing this and all sorts of things way more easy.

pkg/client/testdata/single-node-selector.golden
```diff
@@ -71,12 +71,12 @@ data:
       env:
       - name: CHROOT_DIR
         value: /node
-      - name: RESULTS_DIR
-        value: /tmp/results
       - name: NODE_NAME
         valueFrom:
           fieldRef:
             fieldPath: spec.nodeName
+      - name: RESULTS_DIR
```
johnSchnake, line 78: Moving these into alphabetical order is what I would have expected. Not exactly sure which part caused this to change but IMO its just a good thing.

cmd/sonobuoy/app/args.go
```diff
@@ -117,6 +117,9 @@ func AddKubeConformanceImageVersion(imageVersion *image.ConformanceImageVersion,
 
 	*imageVersion = image.ConformanceImageVersionAuto
 	flags.Var(imageVersion, "kube-conformance-image-version", help)
+	if err := flags.MarkDeprecated("kube-conformance-image-version", "Use --kubernetes-version instead."); err != nil {
+		panic("Failed to setup flags properly")
```
johnSchnake, line 121: This would happen every time you run so I dont think its a risk at all to panic here.

cmd/sonobuoy/app/gen.go
johnSchnake: Confirmed either works; if both set the latest applies. Not going to test for that though since it is obviously not desirable behavior to persist.

johnSchnake:
Its definitely not as large a PR as the line count implies; most of that is testing. Please take a look though @vladimirvivien since it is a much more visible feature.

johnSchnake:
Added another goldenfile test with a plugin that uses it in their image instead of just the env vars. I think this is good to go now.

vladimirvivien (approved):
LGTM

cmd/sonobuoy/app/args.go
```diff
@@ -359,13 +362,15 @@ func AddPluginListFlag(p *[]string, flags *pflag.FlagSet) {
...
-// AddKubernetesVersionFlag adds a string flag for the user to specify the cluster version to assume instead of reaching
-// out to the cluster to auto-detect it.
-func AddKubernetesVersionFlag(version *string, flags *pflag.FlagSet) {
-	flags.StringVar(
-		version, "kubernetes-version", "",
-		"Version to assume for Kubernetes. If empty, the cluster will be queried for its version",
-	)
+// AddKubernetesVersionFlag initialises an image version flag.
+func AddKubernetesVersionFlag(imageVersion *image.ConformanceImageVersion, flags *pflag.FlagSet) {
+	help := "Use default Conformance image, but override the version. "
```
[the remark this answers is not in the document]
johnSchnake: Definitely could. I just copied the code that was there before.

PR #1310, opened 2021-06-18
Update the gen plugin e2e command to take a configmap

johnSchnake:
- Adds new flag to that command. It is somewhat forced in there but
it is different than the other genset and it shouldnt get all of the
gen plugin flags either IMO.
- Adds integration test for any/all commands that check for static
output. This may be gen plugin or other info that doesnt require
timestamps/UUIDs. We can extend this with a hidden field for forcing
static UUIDs or something so that we can test all gen logic this way
potentially.
- Does NOT add this to the systemd-logs plugin since there isn't really
a use case for that at this time.


cmd/sonobuoy/app/gen_plugin_def.go
```diff
@@ -213,24 +213,42 @@ func genPluginDef(cfg *GenPluginDefConfig) ([]byte, error) {
...
+		}
+		for _, v := range *configMapFiles {
+			fData, err := os.ReadFile(v)
+			if err != nil {
+				return errors.Wrapf(err, "failed to read file %q", v)
+			}
+			base := filepath.Base(v)
+			m.ConfigMap[base] = string(fData)
+		}
+
```
johnSchnake, line 251: Just copied code from the other handling of this. IMO fine to duplicate the code, partially since hopefully this is short lived when we pull out the e2e plugin entirely.

scripts/build_funcs.sh
```diff
@@ -60,7 +60,7 @@ integration() {
         --env SONOBUOY_CLI="$SONOBUOY_CLI" \
         --network host \
         "$BUILD_IMAGE" \
-    go test ${VERBOSE:+-v} -timeout 3m -tags=integration "$GOTARGET"/test/integration/... -run TestConfigmaps
```
johnSchnake, line 63: Accidentally added in a previous commit.

test/integration/sonobuoy_integration_test.go
```diff
@@ -443,3 +446,44 @@ func pwd(t *testing.T) string {
 	}
 	return pwd
 }
+
+// TestExactOutput is to test things which can expect exact output; so do not use it
+// for things like configs which include timestamps or UUIDs.
+func TestExactOutput(t *testing.T) {
```
johnSchnake, line 452: Will be quite useful testing other commands. We have goldenfiles for some gen logic but using it based on just cmdline logic is way more wholistic AND simple.

test/integration/testdata/gen-plugin-e2e-configmap
```diff
@@ -0,0 +1,44 @@
...
+    value: \[Conformance\]
+  - name: E2E_SKIP
+    value: \[Disruptive\]|NoExecuteTaintManager
+  - name: E2E_PARALLEL
+    value: "false"
+  - name: E2E_USE_GO_RUNNER
+    value: "true"
+  - name: E2E_EXTRA_ARGS
+    value: --progress-report-url=http://localhost:8099/progress
+  image: k8s.gcr.io/conformance:v1.20.2
```
johnSchnake: img is based on my current cluster. Should just hardcode the value to make this static.

scripts/build_funcs.sh
johnSchnake: This is actually hiding another bug; on fixing that now.

johnSchnake: 💯 No bug, just a timeout issue since we keep adding new tests. A good thing.  See #1311

johnSchnake:
Reasonably small, tightly scoped, and has tests. Merging.

PR #1311, opened 2021-06-18
Extend timeout and fix integration test invocation

johnSchnake:
The tests are simply getting longer in total; the 3m timeout
is not for individual tests but for the whole suite.

Using default of 10m and fixing an accidental filter on the tests.


johnSchnake:
One line change and fixes the build. Merging.

PR #1312, opened 2021-06-21
Add configurable log levels for logrus

### LINKED ISSUE #1189: Make log level configurable
### Description

Sonobuoy currently logs to stderr for `DEBUG` level and above. Ideally we'd like this to be configurable at runtime.

Implementation can either use environment variables (i.e. `LOG_LEVEL=DEBUG`) or global flags (i.e. `--debug` / `--verbose=3`)

### Notes

There are `TODO` marks to downgrade log levels once this feature is achieved (e.g. [`pkg/image/docker/client.go`](https://github.com/vmware-tanzu/sonobuoy/blob/185c859371c7dfa6858906ab6e1f220c51cfaf0b/pkg/image/docker/client.go#L39-L50))

johnSchnake:
- Adds string flag to set log level for logrus
- Removes todo to update some of the logging to debug
- Marks debug flag as hidden/deprecated
- Marks klog flags as hidden
- If debug flag is set, simply sets the log level to debug.

Fixes #1189

**Release note**:
```
Numerous logging flags. from klog have been marked as hidden to avoid clutter in the help menu. In addition, the --debug flag has been replaced with a --level flag that can be set to "debug" for the same effect as well as finer grain control over future logging.
```

### DIFF
--- a/cmd/sonobuoy/app/root.go
+++ b/cmd/sonobuoy/app/root.go
@@ -62,9 +62,19 @@ func NewSonobuoyCommand() *cobra.Command {
 	cmds.AddCommand(NewCmdResults())
 	cmds.AddCommand(NewCmdSplat())
 
-	klog.InitFlags(nil)
-	cmds.PersistentFlags().AddGoFlagSet(flag.CommandLine)
+	initKlog(cmds)
+	cmds.PersistentFlags().StringVar(&errlog.LogLevel, "level", "info", "Log level. One of {panic, fatal, error, warn, info, debug, trace}")
+
+	// Previously just had debug flag but in desire to have fine grained control over output we opted to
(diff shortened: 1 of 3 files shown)


vladimirvivien (commented):
Sonobuoy uses two loggers (klog and logrus) ?

johnSchnake:
So I think the way/reason it "uses two loggers" is that sonobuoy makes extensive use of the kubernetes client code which uses klog. Then for our own logging it was easier (I guess?) to use logrus for all our code. Nothing ever used klog but in order to aid debugging the klog flags were added in case you wanted to see the deeper k8s code logs.

johnSchnake:
Merging based on offline conversations.

PR #1330, opened 2021-06-25
Add sleep after systemd-logs again

### LINKED ISSUE #1322: v0.52  sonobuoy-systemd-logs deamonset NotReady
`sonobuoy run` on v0.52  don't deploy the sonobuoy-systemd-logs deamonset correctly and it goes into NotReady / Crashloopbackoff

sonobuoy-systemd-logs deamonset should be deployed in an ready state

```
❯ diff -u v0.51_systemd-logs.yaml v0.52-systemd-logs.yaml
--- v0.51_systemd-logs.yaml     2021-06-23 14:30:12.769534884 +0200
+++ v0.52-systemd-logs.yaml     2021-06-23 14:30:27.413461363 +0200
@@ -6,8 +6,7 @@
   command:
   - /bin/sh
   - -c
-  - /get_systemd_logs.sh && while true; do echo "Sleeping for 1h to avoid daemonset
-    restart"; sleep 3600; done
+  - /get_systemd_logs.sh
   env:
   - name: CHROOT_DIR
     value: /node
```

v0.51 works as aspect.

**Environment:**

- Sonobuoy version: v0.52
- Kubernetes version: (use `kubectl version`): v1.21.1
- OS (e.g. from `/etc/os-release`):  debian 10 / flatcar

johnSchnake:
The previous change which removed this sleep did also
ensure that the worker doesn't exit from errors so it doesn't
restart but the plugin container itself still exited and caused
a misleading pod status of CrashLoopBackOff.

That doesn't prevent anything for working but is confusing and
the sleep should be restored.

Fixes #1322

**Release note**:
```
Fixes an issue where the default systemd-logs plugin command would cause the pod to enter the CrashLoopBackOff state. This bug did not prevet the plugin from submitting results properly.
```

### DIFF
--- a/pkg/client/gen.go
+++ b/pkg/client/gen.go
@@ -298,7 +298,7 @@ func SystemdLogsManifest(cfg *GenConfig) *manifest.Manifest {
 			Container: corev1.Container{
 				Name:            "systemd-logs",
 				Image:           cfg.SystemdLogsImage,
-				Command:         []string{"/bin/sh", "-c", `/get_systemd_logs.sh`},
+				Command:         []string{"/bin/sh", "-c", `/get_systemd_logs.sh; while true; do echo "Plugin is complete. Sleeping indefinitely to avoid container exit and automatic restarts from Kubernetes"; sleep 3600; done`},
 				ImagePullPolicy: corev1.PullPolicy(cfg.ImagePullPolicy),
 				Env: []corev1.EnvVar{
 					{Name: "CHROOT_DIR", Value: "/node"},
--- a/pkg/client/testdata/custom-systemd-logs-image.golden
+++ b/pkg/client/testdata/custom-systemd-logs-image.golden
@@ -34,7 +34,8 @@ data:
       command:
       - /bin/sh
       - -c
-      - /get_systemd_logs.sh
+      - /get_systemd_logs.sh; while true; do echo "Plugin is complete. Sleeping indefinitely
+        to avoid container exit and automatic restarts from Kubernetes"; sleep 3600; done
       env:
       - name: CHROOT_DIR
         value: /node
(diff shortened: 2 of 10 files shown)


johnSchnake:
- Previously used `&&` and have changed it to `;` so it doesn't matter the exit status of the script.
- Improved error message for more clarity

johnSchnake:
This will eventually be obsolete but it is going to be a common question I think. Wouldn't hurt to fix in the default path currently; as we work to extract plugins from the core logic this will probably still be the default for a while or at least a fallback option.

vladimirvivien (commented):
Would `podSpec: restartPolicy=never` also work for preventing pod from restarting ?

johnSchnake:
Thats not a valid restartPolicy for daemonsets. The only allowed value is `Always`.

johnSchnake:
I nearly tried to take the sleeps here _back_ out but knew this had gone back and forth. Again, I found that the way daemonsets work we need to leave this sleep in.

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy

Makes it clear that the restart policy of the Pod (which has to be Always) applies to _every container individually_ so even if we have the worker properly kept running, the log gathering container will keep restarting oddly.

johnSchnake:
I nearly tried to take the sleeps here _back_ out but knew this had gone back and forth. Again, I found that the way daemonsets work we need to leave this sleep in.

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy

Makes it clear that the restart policy of the Pod (which has to be Always) applies to _every container individually_ so even if we have the worker properly kept running, the log gathering container will keep restarting oddly.

PR #1333, opened 2021-06-25
Prevent logs -f from repeating

johnSchnake:
The retryWatcher in client go seems to be broken and results
in the logs being restarted from the same spot each time despite
the version info being set.

Using a typical watch seems to fix the issue.

**Release note**:
```
Fixed a bug where following logs (sonobuoy logs -f) would result in the same log messages being repeated.
```


vladimirvivien (approved):
LGTM

PR #1336, opened 2021-06-28
Add more descriptive status output

### LINKED ISSUE #1328: Improve initial status check information
Very frequently people run into problems early in the run and the only thing they have to show for the bug is:

```
time="2021-06-25T08:03:46-04:00" level=error msg="error attempting to run sonobuoy: pod has status \"Pending\""
```

Unless they, on their own, run a series of kubectl commands.

We should improve the logic around initial status checks to clarify the state and recommend actions if problems are detected.

johnSchnake:
Whenever we do not have an aggregation status, we can still
direct the user what to do next (usually debug with kubectl).

When possible, we should be able to tell them we can't pull the
image etc to shorten their cycles.

Fixes #1328

**Release note**:
```
Improved output of `sonobuoy status` when the pod was not in the expected state to make it more clear how to proceed.
```

### DIFF
--- a/cmd/sonobuoy/app/status.go
+++ b/cmd/sonobuoy/app/status.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/vmware-tanzu/sonobuoy/pkg/client"
 	"github.com/vmware-tanzu/sonobuoy/pkg/errlog"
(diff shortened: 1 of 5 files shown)


cmd/sonobuoy/app/status.go
```diff
@@ -122,6 +122,12 @@ func getStatus(f *statusFlags) func(cmd *cobra.Command, args []string) {
 }
 
 func exitCode(status *aggregation.Status) int {
+	// Allow status==nil to be non-error path. Explicit errors have been handled
```
vladimirvivien, line 125: Love the insightful comments!

cmd/sonobuoy/app/status.go
```diff
@@ -145,6 +151,34 @@ func humanReadableStatus(str string) string {
 	}
 }
 
+func printPodInfo(w io.Writer, pod *corev1.Pod) error {
+	switch pod.Status.Phase {
+	case (corev1.PodRunning):
+		fmt.Fprintf(w, "The Sonobuoy aggregator is running but we were unable to obtain the plugin status information.\n\nIf this state persists, use kubectl to debug further.\n")
```
vladimirvivien, line 157: Consider "we" -> "is"

vladimirvivien (commented):
LGTM, left one suggestion.

PR #1338, opened 2021-06-29
Simplify config handling

johnSchnake:
In the quest for removing code and eventually the special handling
of the e2e/systemdlogs plugins, I had to start untangling some of the
different configuration management.

It was always more complex than desired but we never had enough
of a reason to spend the time and untangle it. We have that
motivation now since in order to remove those plugins from the core
I have to wire up numerous types and flags differently which impacts
all of the other config 'resolution' code.

In this PR, I took the 4 fields from the genFlags type which were
duplicated in the config (it had a config itself as well as these
four fields for convenience setting from the cmdline). Now it properly
merges the different flags/configs appropriately regardless where they
are set.

**Release note**:
```
Adds the new condition that if you are using `sonobuoy gen|run` with a custom config file (`--config`) you must provide it before setting other flags that modify the configuration. Not expected to impact many users, if any.
```


johnSchnake:
Some yak shearing to occur here; other PRs to precede this one and ensure we add some extra tests and such

cmd/sonobuoy/app/gen.go
```diff
@@ -229,99 +223,3 @@ func getClient(kubeconfig *Kubeconfig) (*kubernetes.Clientset, error) {
...
-	emptyResources := false
-	if len(dst.Resources) == 0 && dst.Resources != nil {
-		emptyResources = true
-	}
-
-	// Workaround to differentiate a false value and a nil value
-	// Only override dst.Limits.PodLogs.SonobuoyNamespace when it's nil
-	// See https://github.com/imdario/mergo/issues/89
-	var sonobuoyNamespace *bool
-	if dst.Limits.PodLogs.SonobuoyNamespace != nil {
```
johnSchnake, line 313: If not using mergo, this isnt an issue I think

johnSchnake: Same goes for the others. We start with the default config, load the given config if desired, update only the values they specify, and move on. No weird merging logic any more at all.

The price for this is simply an error (with instructions) if you do something like `--timeout 10 --config custom.json` and you just need to provide the config first so it gets layered correctly. Probably no one will be impacted.

cmd/sonobuoy/app/gen_config.go
```diff
@@ -14,12 +14,14 @@ import (
...
 func NewCmdGenConfig() *cobra.Command {
 	var f genFlags
+
 	var GenCommand = &cobra.Command{
 		Use:   "config",
 		Short: "Generates a sonobuoy config for input to sonobuoy gen or run.",
 		Run:   genConfigCobra(&f),
 		Args:  cobra.ExactArgs(0),
 	}
+	GenCommand.Flags().AddFlagSet(GenFlagSet(&f, EnabledRBACMode))
```
johnSchnake, line 24: Need to init the flags properly. Has the benefit of having all the relevant gen flags available to make a slightly customized config file. Really not different than even just running gen command and grepping for the config. Not sure if I could write that in fewer LOC though.

cmd/sonobuoy/app/gen_test.go
```diff
@@ -1,363 +0,0 @@
...
-
-	"github.com/kylelemons/godebug/pretty"
-	v1 "k8s.io/api/core/v1"
-)
-
-const (
-	rawInput = "not nil"
-)
-
-func TestResolveConfig(t *testing.T) {
```
johnSchnake, line 37: No more wonky resolution function you have to know to call 🎉

cmd/sonobuoy/app/images_test.go
```diff
@@ -99,9 +99,9 @@ func TestGetClusterVersion(t *testing.T) {
 			output, err := getClusterVersion(tc.input, Kubeconfig{})
 			switch {
 			case err != nil && !tc.expectErr:
-				t.Fatalf("Expected no error but got %v", err)
+				t.Errorf("Expected no error but got %v", err)
```
johnSchnake, line 102: When developing here I found I needed the next messages and not just a fatal here.

cmd/sonobuoy/app/run.go
```diff
@@ -63,10 +63,6 @@ func (r *runFlags) Config() (*client.RunConfig, error) {
 		GenFile:    r.genFile,
 	}
 
-	if r.genFile != "" && givenAnyGenConfigFlags(&(r.genFlags), allowedGenFlagsWithRunFile) {
```
johnSchnake, line 66: All moved into a prerun method which natively has all the flags available. No longer need to put the `genFlags` flagset inside the `genFlags` object (I probably have the casing wrong, but thats why I'm so happy to get rid of it!

cmd/sonobuoy/app/sonobuoyconfig.go
```diff
@@ -41,11 +42,15 @@ func (c *SonobuoyConfig) String() string {
 // Type is needed for pflag.Value.
 func (c *SonobuoyConfig) Type() string { return "Sonobuoy config" }
 
-// Set attempts to read a file, then deserialise the json into a config.Config struct.
+// Set attempts to read a file, then deserialize the json into a config.Config struct.
 func (c *SonobuoyConfig) Set(str string) error {
+	if !reflect.DeepEqual(c.Config, *config.New()) {
```
johnSchnake, line 47: These 4 lines replace a 50 line method and 200+ lines of tests. Extreme simplification.

cmd/sonobuoy/app/sonobuoyconfig.go
```diff
@@ -56,12 +61,7 @@ func (c *SonobuoyConfig) Set(str string) error {
 	return nil
 }
 
-// Get will return the config.Config if one is available, otherwise nil.
+// Get will return the config.Config.
 func (c *SonobuoyConfig) Get() *config.Config {
-	// Don't just return zero structs
```
johnSchnake, line 61: Just initialize the method to the default which is normally what you do with flags. Not sure why we did it this way at first.

scripts/build_funcs.sh
```diff
@@ -184,6 +184,8 @@ native() {
     LDFLAGS="-s -w -X $GOTARGET/pkg/buildinfo.Version=$GIT_VERSION -X $GOTARGET/pkg/buildinfo.GitSHA=$GIT_REF_LONG"
     args=(-ldflags "${LDFLAGS}" "$GOTARGET")
     CGO_ENABLED=0 GOOS="$HOST_GOOS" GOARCH="$HOST_GOARCH" go build -o sonobuoy "${args[@]}"
+    mkdir -p ./build/$HOST_GOOS/$HOST_GOARCH
+    cp ./sonobuoy ./build/$HOST_GOOS/$HOST_GOARCH/sonobuoy
```
johnSchnake, line 188: Just incidental improvement, not really related but doesnt deserve its own PR to improve a small local method.

test/integration/sonobuoy_integration_test.go
```diff
@@ -48,16 +48,16 @@ func findSonobuoyCLI() (string, error) {
 
 // runSonobuoyCommandWithContext runs the Sonobuoy CLI with the given context and arguments.
 // It returns any encountered error and the stdout and stderr from the command execution.
-func runSonobuoyCommandWithContext(ctx context.Context, t *testing.T, args string) (bytes.Buffer, bytes.Buffer, error) {
-	var stdout, stderr bytes.Buffer
+func runSonobuoyCommandWithContext(ctx context.Context, t *testing.T, args string) (bytes.Buffer, error) {
+	var combinedOutput bytes.Buffer
```
johnSchnake, line 52: When testing anything we really care about what the users see; rarely exactly where it goes. Separating out the stdout/stderr was much harder to test 'what does the user see...'

test/integration/sonobuoy_integration_test.go
```diff
@@ -513,18 +512,15 @@ func TestExactOutput(t *testing.T) {
 			desc:       "gen with config then flags targeting subfields",
 			cmdLine:    "gen --config=testdata/subfieldTest.json -n cmdlineNS --image-pull-policy=Always --sonobuoy-image=cmdlineimg --timeout=99",
 			expectFile: "testdata/gen-config-then-flags.golden",
-		}, {
-			desc:       "gen with flags targeting subfields then config",
-			cmdLine:    "gen -n=cmdlineNS --image-pull-policy=Always --sonobuoy-image=cmdlineimg --timeout=99 --config=testdata/subfieldTest.json",
-			expectFile: "testdata/gen-flags-then-config.golden",
```
johnSchnake, line 519: Moved this test case below, didnt remove it.

test/integration/sonobuoy_integration_test.go
```diff
@@ -543,3 +539,30 @@ func TestExactOutput(t *testing.T) {
 		})
 	}
 }
+
+func TestOutputIncludes(t *testing.T) {
```
johnSchnake, line 543: Important to test it like this for clarity. TestExactOutput does what it says, tests all the exact output. This is just testing that some output includes....xyz.

Here we are testing an error message. This is important since we don't want to have golden files of our help output which would cause us to have to update golden files anytime we update unrelated flags, commands, help, etc.

test/integration/testdata/gen-config-no-flags.golden
```diff
@@ -52,7 +52,7 @@ rules:
...
...isions","cronjobs","customresourcedefinitions","daemonsets","deployments","endpoints","ingresses","jobs","leases","limitranges","mutatingwebhookconfigurations","namespaces","networkpolicies","nodes","persistentvolumeclaims","persistentvolumes","poddisruptionbudgets","pods","podlogs","podsecuritypolicies","podtemplates","priorityclasses","replicasets","replicationcontrollers","resourcequotas","rolebindings","roles","servergroups","serverversion","serviceaccounts","services","statefulsets","storageclasses","validatingwebhookconfigurations","volumeattachments"],"Filters":{"Namespaces":".*","LabelSelector":""},"Limits":{"PodLogs":{"Namespaces":"","SonobuoyNamespace":true,"FieldSelectors":[],"LabelSelector":"","Previous":false,"SinceSeconds":null,"SinceTime":null,"Timestamps":false,"TailLines":null,"LimitBytes":null,"LimitSize":"","LimitTime":""}},"QPS":30,"Burst":50,"Server":{"bindaddress":"0.0.0.0","bindport":8080,"advertiseaddress":"","timeoutseconds":12345},"Plugins":null,"PluginSearchPath":["./plugins.d","/etc/sonobuoy/plugins.d","~/sonobuoy/plugins.d"],"Namespace":"configfileNS","WorkerImage":"configImg","ImagePullPolicy":"Never","ImagePullSecrets":"","ProgressUpdatesPort":"8099"}
```
johnSchnake, line 55: The change in questopn is that the pluginSelection is now null instead of e2e and systemd-logs.

This is a noticable but probably inconsequential change.

The issue was that pluginSelection used to be necessary since systemdlogs and e2e were ALWAYS present in the original sonobuoy code so you had to specify which to run.

Then we added custom plugins so we started adding custom plugins to the 2 default ones, and still had to say which to run.

Then we defaulted to uploading only the ones we wanted to run which made this field generally meaningless. You could, in theory, do a sonobuoy gen with 100 plugins but then "select" only a few of them manually in the YAML but we have never heard of someone doing that.

test/integration/testdata/gen-config-no-flags.golden
```diff
@@ -98,7 +98,7 @@ data:
       - name: SONOBUOY_K8S_VERSION
         value: v1.20.0
       image: k8s.gcr.io/conformance:v1.20.0
-      imagePullPolicy: Always
```
johnSchnake, line 101: This wasn't matching the config value above, so this is an improvement tbh

test/integration/sonobuoy_integration_test.go
```diff
@@ -513,18 +512,15 @@ func TestExactOutput(t *testing.T) {
...
 	}
 	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
-			output := mustRunSonobuoyCommandWithContext(ctx, t, tc.cmdLine)
+			// Allow errors here since we also may test stderr
+			output, _ := runSonobuoyCommandWithContext(ctx, t, tc.cmdLine)
+
 			binaryVersion := mustRunSonobuoyCommand(t, "version --short")
 			binaryVer := strings.TrimSpace(binaryVersion.String())
-			t.Logf("version output of binary, after trimming space: %q\n", binaryVer)
```
johnSchnake, line 527: Just needed this for a while; after looking at more logs it isnt quite necessary.

test/integration/testdata/gen-flags-then-config.golden
```diff
@@ -1,205 +0,0 @@
----
```
johnSchnake, line 1: Dont need this since it is covered by the TestOutputIncludes test now and doesnt use goldenfiles.

test/integration/testdata/gen-no-uuid.golden
```diff
@@ -98,7 +98,7 @@ data:
       - name: SONOBUOY_K8S_VERSION
         value: v1.20.0
       image: k8s.gcr.io/conformance:v1.20.0
-      imagePullPolicy: Always
```
johnSchnake, line 101: There was a disconnect between the intended defaults and the _actual_ defaults. This fixes it.

test/integration/testdata/gen-plugin-e2e-kube-flag-still-works.golden
```diff
@@ -1,3 +1,4 @@
+Flag --kube-conformance-image-version has been deprecated, Use --kubernetes-version instead.
```
johnSchnake, line 1: Hey! Good news this showed up; just noticed it on a self-review but we do want this shown. In sonobuoy gen it is unlikely they'll see it though.

johnSchnake:
@zubron I tagged you in this even though you dont work primarily on sonobuoy anymore; I just thought that we might benefit from another set of eyes and I know you had tried to untangle some of these things in the past. I might have even fought to keep some of the weird logic. No more!

@vladimirvivien Tiny tiny tiny concession of logic in order to remove 600 LOC and simplify code flows. I hope you'll agree that its worth it.

vladimirvivien (commented):
/lgtm

I really appreciate the self-review. Makes looking over this (larger PRs) easier.

vladimirvivien (approved):

cmd/sonobuoy/app/gen.go
```diff
@@ -229,99 +223,3 @@ func getClient(kubeconfig *Kubeconfig) (*kubernetes.Clientset, error) {
...
-	//  - Otherwise use the supplied config and mode to figure out the plugins to run.
-	//    This only works for e2e/systemd-logs which are internal plugins so then "Set" them
-	//    as if they were provided on the cmdline.
-	// Gate the logic with a nil check because tests may not specify flags and intend the legacy logic.
-	if g.genflags == nil || !g.genflags.Changed("plugin") {
-		// Use legacy logic; conf.SelectedPlugins or mode if not set
-		if conf.PluginSelections == nil {
-			modeConfig := g.mode.Get()
-			if modeConfig != nil {
-				conf.PluginSelections = modeConfig.Selectors
```
johnSchnake, line 261: This actually should have been pulled out and put in https://github.com/vmware-tanzu/sonobuoy/pull/1343 but that finishes the job and removes the selectors (this is all that it was used for).

Just putting an FYI here

PR #1339, opened 2021-06-29
Default config.New to not provide UUID

johnSchnake:
This is all on the client-side and the server will fill it in
even if it is empty. By not setting it so early, it becomes
much easier to test things. Users, if they want a static UUID
can save it explicitly as always.

**Release note**:
```
Configs generated on the client-side will no longer include UUIDs. This means that `sonobuoy gen` output will have an empty UUID and successive runs with that `sonobuoy gen` output will be able to be differentiated by their UUIDs.
```


johnSchnake:
Debugging a test issue; the version (from buildinfo) is giving me a problem for some reason even though I was trying to swap that out in a way that would prevent it from mattering.

johnSchnake:
Had a good understanding of the problem during PR state and fixed it. Otherwise no issues; if anyone still provide a UUID it works fine and if they submit one without a UUID it still works fine. Merging.

PR #1340, opened 2021-06-29
Extra gen tests

johnSchnake:
For an upcoming PR I'm trying to ensure existing behavior is established
in these tests rather than unit tests which are more difficult to
preserve when we modify large chunks of underlying types/logic.

No external changes.


scripts/build_funcs.sh
```diff
@@ -197,7 +182,7 @@ build_binaries() {
 # Builds sonobuoy using the local goos/goarch.
 native() {
     LDFLAGS="-s -w -X $GOTARGET/pkg/buildinfo.Version=$GIT_VERSION -X $GOTARGET/pkg/buildinfo.GitSHA=$GIT_REF_LONG"
-    args=("${VERBOSE:+-v}" -ldflags "${LDFLAGS}" "$GOTARGET")
+    args=(-ldflags "${LDFLAGS}" "$GOTARGET")
```
johnSchnake, line 185: Without verbose this actually failed. Just a local helper so I just removed it.

scripts/build_funcs.sh
```diff
@@ -35,21 +35,6 @@ WIN_AMD64_BASEIMAGE=mcr.microsoft.com/windows/nanoserver
...
-    go test $GOTARGET/cmd/... $GOTARGET/pkg/... -update > tmp_update.out
-    cat tmp_update.out | grep github | grep -v ok
-    cat tmp_update.out | grep github | grep ok
-    rm tmp_update.out
-
-    # Integration tests take longer and need kind (usually). Just run the test we need.
-    go build -o sonobuoy
-    go test $GOTARGET/test/integration -update -tags integration -run TestExactOutput
-}
-
```
johnSchnake, line 52: Just moved lower to use the native build method

johnSchnake:
No issues and just extra test coverage; merging.

PR #1342, opened 2021-06-29
Update github actions workflows to show history

johnSchnake:
Ran into a problem where it wasnt clear if the action
was checking out the expected commit. Adds this log line
to the unit tests and integration tests just to have
an extra debug point.

A few points of formatting touched up as well.


johnSchnake:
Love this. Makes it clear that you're actually checked out onto a theoretical merge branch and you can ensure you have the commits you think you need. Merging.

PR #1343, opened 2021-07-01
Simplify the way mode and plugin selection is processed

johnSchnake:
- Mode will no longer remove systemd-logs when in quick mode. This
requires extra configuration code and provides little to no benefit
since the logs gathering plugin is so fast. Meanwhile, by removing it
we simplify the code in a way that makes upcoming changes regarding
e2e flags more simple.

- PluginSelection (the field in the config file) will continue
to be used for filtering which plugins are run on the server, but
will no longer be the appropriate place to list plugins in order to
run them. Modern sonobuoy code had simply been supporting this legacy
logic from when sonobuoy usually only ran our default plugins and
never had other fields to track them. To be clear: the CLI doesnt even use
the pluginSelection field typically; it was just the precursor to staticPlugins
and dynamic plugins (which should, soon, become just plugins)

**Release note**:
```
- Mode will no longer remove systemd-logs when in quick mode. To run only the e2e plugin, add the flag `-p e2e`

- Sonobuoy will no longer generate plugin yaml based off of the sonobuoy config file's "plugins" field. This field can just be left blank (indicating to run all loaded plugins) and you should either specify no plugins (taking the defaults) or specify the plugins you want with the `-p` flag.
```

### DIFF
--- a/pkg/client/gen.go
+++ b/pkg/client/gen.go
@@ -102,20 +102,13 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
 		}
 	}
 
-	// Support legacy logic for the time being.
+	// If the user didnt provide plugins at all fallback to our original
+	// defaults. Legacy logic was that the user could specify plugins via the
+	// config.PluginSelection field but since the CLI handles custom plugins
+	// we moved to the list of actual plugin data. The PluginSelection is only
+	// a server-side capability to run a subset of available plugins.
 	if len(cfg.DynamicPlugins) == 0 && len(cfg.StaticPlugins) == 0 {
-		if conf.PluginSelections != nil {
-			// Empty (but non-nil) means run nothing. Setting any value means run
-			// those explicitly.
-			for _, v := range conf.PluginSelections {
-				cfg.DynamicPlugins = append(cfg.DynamicPlugins, v.Name)
-			}
-		} else {
-			// Nil plugin selection now means to run all plugins that are loaded.
-			// If the user didnt provide plugins at all fallback to our original
-			// defaults.
-			cfg.DynamicPlugins = []string{e2ePluginName, systemdLogsName}
-		}
+		cfg.DynamicPlugins = []string{e2ePluginName, systemdLogsName}
 	}
 
 	plugins := []*manifest.Manifest{}
--- a/pkg/client/gen_test.go
+++ b/pkg/client/gen_test.go
@@ -192,7 +192,7 @@ func TestGenerateManifestGolden(t *testing.T) {
 			},
 			goldenFile: filepath.Join("testdata", "systemd-logs-default.golden"),
 		}, {
-			name: "Enabling SSH (legacy plugin choice)",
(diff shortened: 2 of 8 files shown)


pkg/client/gen_test.go
```diff
@@ -192,7 +192,7 @@ func TestGenerateManifestGolden(t *testing.T) {
 			},
 			goldenFile: filepath.Join("testdata", "systemd-logs-default.golden"),
 		}, {
-			name: "Enabling SSH (legacy plugin choice)",
+			name: "Enabling SSH (via legacy plugin choice)",
```
johnSchnake, line 195: Just more clear to me

pkg/client/gen_test.go
```diff
@@ -203,15 +203,15 @@ func TestGenerateManifestGolden(t *testing.T) {
 			},
 			goldenFile: filepath.Join("testdata", "ssh.golden"),
 		}, {
-			name: "Empty array leads to 0 plugins (legacy plugin choice)",
+			name: "Empty array leads to default plugins, not 0",
```
johnSchnake, line 206: This is the "breaking change" that I dont think matters.

I was the original proponent of the 'run 0 plugins' case but I've gotten over it and I dont think anyone ever actually ran it that way. It also makes it harder for users to know the difference between empty and nil.

This idea that 0/nil both have the same behavior is much more intuitive.

pkg/client/testdata/default-plugins-via-selection.golden
```diff
@@ -24,6 +24,69 @@ metadata:
   namespace: sonobuoy
 ---
 apiVersion: v1
+data:
```
johnSchnake, line 27: Most of these changes to golden files are simply adding 1 of the default plugins to the YAML since these tests were just setting the PluginSelection field and not the `DynamicPlugins` field.

johnSchnake:
Don't let the line count fool you; most removed/edited code but a bunch of extra lines in the goldenfiles was addedl

pkg/client/gen.go
```diff
@@ -102,20 +102,13 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
...
-			// those explicitly.
-			for _, v := range conf.PluginSelections {
-				cfg.DynamicPlugins = append(cfg.DynamicPlugins, v.Name)
-			}
-		} else {
-			// Nil plugin selection now means to run all plugins that are loaded.
-			// If the user didnt provide plugins at all fallback to our original
-			// defaults.
-			cfg.DynamicPlugins = []string{e2ePluginName, systemdLogsName}
-		}
```
johnSchnake, line 118: This is the main bit of legacy code we're getting rid of. Just the more simply default now so it is easy to grok. I had to read this block of code many times to understand what/why it was doing that.

johnSchnake:
https://github.com/vmware-tanzu/sonobuoy/pull/1338 actually changed the way the mode acted; I meant to pull that functionality into this PR but oh well.

This one has even less impact then so I'll merge it when tests are green and keep going on this track...

PR #1344, opened 2021-07-01
Wire e2e focus/skip/mode/parallel into existing env structure

johnSchnake:
Removing types that are e2e specific while keeping our flags
for convenience/shorthand.


cmd/sonobuoy/app/args.go
```diff
@@ -168,62 +172,48 @@ const (
...
-	)
-	e2eFlags.String(
-		e2eSkipFlag, defaultMode.E2EConfig.Skip,
-		"Specify the E2E_SKIP flag to the conformance tests. Overrides --mode.",
-	)
-	e2eFlags.String(
-		e2eParallelFlag, defaultMode.E2EConfig.Parallel,
-		"Specify the E2E_PARALLEL flag to the conformance tests. Overrides --mode.",
-	)
+
```
johnSchnake, line 175: The e2econfig type will eventually go away is the intent; but the registry config flag is very different so I'll do that in another PR

cmd/sonobuoy/app/args.go
```diff
@@ -401,3 +391,76 @@ func AddExtractFlag(flag *bool, flags *pflag.FlagSet) {
 		"If true, extracts the results instead of just downloading the results",
 	)
 }
+
+type e2eFocusFlagType PluginEnvVars
```
johnSchnake, line 395: So most of the 'added' lines are like these: we remove logical complexity throughout the codebase and for this flag-specific stuff, wire up these flag types. The flag methods of get/set/etc handle all the complexity in a very narrowly scoped point in code/time.

So in these cases, each of the focus/skip/parallel/mode, you get a new flag type which defines a custom action when you set it. These end up setting env vars but you could imagine them doing other things if needed.

cmd/sonobuoy/app/args.go
```diff
@@ -168,62 +172,48 @@ const (
...
-		e2eFocusFlag, defaultMode.E2EConfig.Focus,
-		"Specify the E2E_FOCUS flag to the conformance tests. Overrides --mode.",
-	)
-	e2eFlags.String(
-		e2eSkipFlag, defaultMode.E2EConfig.Skip,
-		"Specify the E2E_SKIP flag to the conformance tests. Overrides --mode.",
-	)
-	e2eFlags.String(
-		e2eParallelFlag, defaultMode.E2EConfig.Parallel,
-		"Specify the E2E_PARALLEL flag to the conformance tests. Overrides --mode.",
```
johnSchnake, line 183: Again a bit of a simplification here that I dont think loses anything. We had to nest flags within our types so we could do checking later if the flag was provided or a default value. Now we just say: let the flags figure it out.

If they provide the flags in a conflicting way (--mode=quick --e2e-focus=foo) then the flags will sort of fight over the value; last to set it wins.

I think this is reasonable and I dont see a need to bend over backwards to support wonkily-written commands like that.

cmd/sonobuoy/app/args_test.go
```diff
@@ -1,89 +0,0 @@
...
-
-import (
-	"testing"
-
-	"github.com/spf13/pflag"
-
-	ops "github.com/vmware-tanzu/sonobuoy/pkg/client"
-)
-
-func TestGetE2EConfig(t *testing.T) {
```
johnSchnake, line 27: Mode is just a flag setting env vars; we have gen tests and smoke tests etc. No loss of coverage imo.

cmd/sonobuoy/app/gen.go
```diff
@@ -97,11 +97,19 @@ func GenFlagSet(cfg *genFlags, rbac RBACMode) *pflag.FlagSet {
 }
 
 func (g *genFlags) Config() (*client.GenConfig, error) {
-	e2ecfg, err := GetE2EConfig(g.mode, g.e2eflags)
+	e2ecfg, err := GetE2EConfig(g.e2eflags)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not retrieve E2E config")
 	}
 
+	// TODO(jschnake) we should probably just set the dynamic plugins here, but
```
johnSchnake, line 105: TODO was old

test/integration/testdata/gen-config-no-flags.golden
```diff
@@ -52,7 +52,7 @@ rules:
...
...finitions","daemonsets","deployments","endpoints","ingresses","jobs","leases","limitranges","mutatingwebhookconfigurations","namespaces","networkpolicies","nodes","persistentvolumeclaims","persistentvolumes","poddisruptionbudgets","pods","podlogs","podsecuritypolicies","podtemplates","priorityclasses","replicasets","replicationcontrollers","resourcequotas","rolebindings","roles","servergroups","serverversion","serviceaccounts","services","statefulsets","storageclasses","validatingwebhookconfigurations","volumeattachments"],"Filters":{"Namespaces":".*","LabelSelector":""},"Limits":{"PodLogs":{"Namespaces":"","SonobuoyNamespace":true,"FieldSelectors":[],"LabelSelector":"","Previous":false,"SinceSeconds":null,"SinceTime":null,"Timestamps":false,"TailLines":null,"LimitBytes":null,"LimitSize":"","LimitTime":""}},"QPS":30,"Burst":50,"Server":{"bindaddress":"0.0.0.0","bindport":8080,"advertiseaddress":"","timeoutseconds":12345},"Plugins":[{"name":"e2e"},{"name":"systemd-logs"}],"PluginSearchPath":["./plugins.d","/etc/sonobuoy/plugins.d","~/sonobuoy/plugins.d"],"Namespace":"configfileNS","WorkerImage":"configImg","ImagePullPolicy":"Never","ImagePullSecrets":"","ProgressUpdatesPort":"8099"}
```
johnSchnake, line 55: Going to double check the changes here; seems the plugins returned and I'll have to make sure I understand why. Not sure if its because of some new defaults or something else.

johnSchnake: In this case, it is using `sonobuoy gen --config custom` and in that custom config the plugins are null explicitly. Need to make sure we arent modifying that.

johnSchnake: Hm, so it makes sense with the simplification of plugin logic. If you dont specify any plugins, we revert to our default. So maybe this is a good thing and I should just update the test to check a run which specifies a random plugin

test/integration/testdata/subfieldTest.json
```diff
@@ -1 +1 @@
...
...omresourcedefinitions","daemonsets","deployments","endpoints","ingresses","jobs","leases","limitranges","mutatingwebhookconfigurations","namespaces","networkpolicies","nodes","persistentvolumeclaims","persistentvolumes","poddisruptionbudgets","pods","podlogs","podsecuritypolicies","podtemplates","priorityclasses","replicasets","replicationcontrollers","resourcequotas","rolebindings","roles","servergroups","serverversion","serviceaccounts","services","statefulsets","storageclasses","validatingwebhookconfigurations","volumeattachments"],"Filters":{"Namespaces":".*","LabelSelector":""},"Limits":{"PodLogs":{"Namespaces":"","SonobuoyNamespace":true,"FieldSelectors":[],"LabelSelector":"","Previous":false,"SinceSeconds":null,"SinceTime":null,"Timestamps":false,"TailLines":null,"LimitBytes":null,"LimitSize":"","LimitTime":""}},"QPS":30,"Burst":50,"Server":{"bindaddress":"0.0.0.0","bindport":8080,"advertiseaddress":"","timeoutseconds":12345},"Plugins":[{"name":"configpluginval"}],"PluginSearchPath":["./plugins.d","/etc/sonobuoy/plugins.d","~/sonobuoy/plugins.d"],"Namespace":"configfileNS","WorkerImage":"configImg","ImagePullPolicy":"Never","ImagePullSecrets":"","ProgressUpdatesPort":"8099"}
```
johnSchnake, line 1: Changing this to configpluginval for plugins in order to differentiate it.

test/integration/testdata/gen-config-then-flags.golden
```diff
@@ -52,7 +52,7 @@ rules:
...
...customresourcedefinitions","daemonsets","deployments","endpoints","ingresses","jobs","leases","limitranges","mutatingwebhookconfigurations","namespaces","networkpolicies","nodes","persistentvolumeclaims","persistentvolumes","poddisruptionbudgets","pods","podlogs","podsecuritypolicies","podtemplates","priorityclasses","replicasets","replicationcontrollers","resourcequotas","rolebindings","roles","servergroups","serverversion","serviceaccounts","services","statefulsets","storageclasses","validatingwebhookconfigurations","volumeattachments"],"Filters":{"Namespaces":".*","LabelSelector":""},"Limits":{"PodLogs":{"Namespaces":"","SonobuoyNamespace":true,"FieldSelectors":[],"LabelSelector":"","Previous":false,"SinceSeconds":null,"SinceTime":null,"Timestamps":false,"TailLines":null,"LimitBytes":null,"LimitSize":"","LimitTime":""}},"QPS":30,"Burst":50,"Server":{"bindaddress":"0.0.0.0","bindport":8080,"advertiseaddress":"","timeoutseconds":99},"Plugins":[{"name":"configpluginval"}],"PluginSearchPath":["./plugins.d","/etc/sonobuoy/plugins.d","~/sonobuoy/plugins.d"],"Namespace":"cmdlineNS","WorkerImage":"cmdlineimg","ImagePullPolicy":"Always","ImagePullSecrets":"","ProgressUpdatesPort":"8099"}
```
johnSchnake, line 55: This is what the config says; load it as instructed. TBH its a bad configuraion since in this run they're saying in the config to run `configplugival` but dont provide that plugin anywhere.

I dont think this really impacts many folks and would only do so once until they learn how pieces work together.

TLDR; I think this is ok.

test/integration/testdata/gen-config-no-flags.golden
```diff
@@ -52,7 +52,7 @@ rules:
...
...omresourcedefinitions","daemonsets","deployments","endpoints","ingresses","jobs","leases","limitranges","mutatingwebhookconfigurations","namespaces","networkpolicies","nodes","persistentvolumeclaims","persistentvolumes","poddisruptionbudgets","pods","podlogs","podsecuritypolicies","podtemplates","priorityclasses","replicasets","replicationcontrollers","resourcequotas","rolebindings","roles","servergroups","serverversion","serviceaccounts","services","statefulsets","storageclasses","validatingwebhookconfigurations","volumeattachments"],"Filters":{"Namespaces":".*","LabelSelector":""},"Limits":{"PodLogs":{"Namespaces":"","SonobuoyNamespace":true,"FieldSelectors":[],"LabelSelector":"","Previous":false,"SinceSeconds":null,"SinceTime":null,"Timestamps":false,"TailLines":null,"LimitBytes":null,"LimitSize":"","LimitTime":""}},"QPS":30,"Burst":50,"Server":{"bindaddress":"0.0.0.0","bindport":8080,"advertiseaddress":"","timeoutseconds":12345},"Plugins":[{"name":"configpluginval"}],"PluginSearchPath":["./plugins.d","/etc/sonobuoy/plugins.d","~/sonobuoy/plugins.d"],"Namespace":"configfileNS","WorkerImage":"configImg","ImagePullPolicy":"Never","ImagePullSecrets":"","ProgressUpdatesPort":"8099"}
```
johnSchnake, line 55: same msg:

This is what the config says; load it as instructed. TBH its a bad configuraion since in this run they're saying in the config to run `configplugival` but dont provide that plugin anywhere.

I dont think this really impacts many folks and would only do so once until they learn how pieces work together.

TLDR; I think this is ok.

cmd/sonobuoy/app/args.go
vladimirvivien: Agreed.
I would only suggest that if both `mode` and `e2e-focus` flags are provided together to, maybe, output a warning.

cmd/sonobuoy/app/args.go
```diff
@@ -401,3 +395,84 @@ func AddExtractFlag(flag *bool, flags *pflag.FlagSet) {
 		"If true, extracts the results instead of just downloading the results",
 	)
 }
+
+type e2eFocusFlagType PluginEnvVars
+
+func (i *e2eFocusFlagType) String() string {
+	if i != nil && (map[string]map[string]string)(*i)["e2e"] != nil {
```
vladimirvivien, line 402: Suggestion: make this `(map[string]map[string]string)` a named type so your type conversion is simpler

```
type FocusFlag map[string]map[string]string

func (i *e2eFocusFlagType) String() string {
	if i != nil && FocusFlag(*i)["e2e"] != nil {
```

vladimirvivien (commented):
Left couple of comments/suggestions.
Ping me if ready for approval.

cmd/sonobuoy/app/args.go
johnSchnake: Its actually already a named type so that was easy to swap. For some reason I thought when indexing it I had to use the underlying map type, but you're right

cmd/sonobuoy/app/args.go
johnSchnake: OK; added a warning in a persistentPrerun method now. Otherwise we have to place it in a bunch of different places. Its a simple check and would apply anywhere those flags are present/provided.

johnSchnake:
I also reworked the boilerplate a bit so I trimmed down that code so the focus/skip/parallel all reuse the same code path. Then I also included the ssh-user in this set of changes because it was exactly the same, just changed an env var.

vladimirvivien (approved):
/lgtm

PR #1347, opened 2021-07-02
Add generic plugin transformations and use for kube-repo-list

johnSchnake:
In an effort to generalize some features to more than just e2e,
I implemented a way to tie command line flags directly to plugins
which may be loaded at a later time.

In this case, we use the transformation logic along with the new
configmap handling for plugins to add a config map and env var
to the e2e plugin whenever the user provides the flag for a custom
e2e repo configuration.

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -18,18 +18,18 @@ package app
 
 import (
 	"fmt"
-	"io/ioutil"
+	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 
-	"github.com/vmware-tanzu/sonobuoy/pkg/client"
 	"github.com/vmware-tanzu/sonobuoy/pkg/image"
+	"github.com/vmware-tanzu/sonobuoy/pkg/plugin/manifest"
 
-	"gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
-	ops "github.com/vmware-tanzu/sonobuoy/pkg/client"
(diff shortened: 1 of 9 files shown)


cmd/sonobuoy/app/args.go
```diff
@@ -468,3 +431,39 @@ func (m *Mode) Set(str string) error {
 	m.name = lcase
 	return nil
 }
+
+type addConfigMapFlag struct {
```
johnSchnake, line 435: Right now the set command is extremely tied to specifically the e2e custom repo config since it also adds the env var.

I think it could be generalized some but that can definitely be done in a follow up PR if we generalize the logic even more (only if we want), for instance to add configmaps to plugins on the fly. I don't think that feature is typically needed though and `sonobuoy gen plugin` can also be used as needed or direct yaml manipulation.

pkg/client/gen.go
```diff
@@ -156,35 +169,33 @@ func (*SonobuoyClient) GenerateManifest(cfg *GenConfig) ([]byte, error) {
 
 	cfg.PluginEnvOverrides, plugins = applyK8sVersion(cfg.KubeVersion, cfg.PluginEnvOverrides, plugins)
 
-	if cfg.PluginEnvOverrides != nil {
```
johnSchnake, line 160: Not sure why the nil check was there; we use range in the next line so nils are not a problem.

johnSchnake:
This approach will be expandable to solve any other issues with plugin-specific flags and could potentially be used to add entirely new features we haven't considered yet. I hope you agree its a good approach to this problem.

I have a followup to this that utilizes it for the sshkey logic (adding volumes to the e2e plugin to utilize a secret)

johnSchnake:
Merging based on offline conversations.

PR #1348, opened 2021-07-03
Ssh path flag

johnSchnake:
This allows the `gen` logic to be even more simple
and avoid e2e specific logic.

### DIFF
--- a/pkg/client/gen.go
+++ b/pkg/client/gen.go
@@ -373,34 +373,6 @@ func E2EManifest(cfg *GenConfig) *manifest.Manifest {
 	}
 	m.PodSpec.PodSpec.NodeSelector = map[string]string{"kubernetes.io/os": "linux"}
 
-	// Add volume mount, volume, and 3 env vars (for different possible platforms) for SSH capabilities.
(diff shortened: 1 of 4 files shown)


johnSchnake:
Just needs rebased after the first PR merges for transforms; I fixed up the test error there but dont want to worry about constantly rebasing.

pkg/client/gen.go
```diff
@@ -373,34 +373,6 @@ func E2EManifest(cfg *GenConfig) *manifest.Manifest {
 	}
 	m.PodSpec.PodSpec.NodeSelector = map[string]string{"kubernetes.io/os": "linux"}
 
-	// Add volume mount, volume, and 3 env vars (for different possible platforms) for SSH capabilities.
-	if len(cfg.SSHKeyPath) > 0 {
```
johnSchnake, line 377: Without using the transforms; this was conditional logic in the e2e manifest itself. Now we can still extract the plugin (without the ssh logic at all) and the transform can act on it at the CLI level of the code and not the gen/plugin lookup state.

PR #1349, opened 2021-07-04
Default to assuming the conformance image can support progress

johnSchnake:
This was introduced in 1.17 which was numerous k8s releases ago.

We previously said we supported 3 releases back in any given Sonobuoy
release so I think it is fair to cut this out now.

It also helps in the ongoing plugin separation changes since it
makes one less thing variable about the e2e tests specifically.

**Release note**:
```
Sonobuoy will now assume the e2e plugin can support sending progress updates; if you have a cluster operating on 1.16.x or earlier, you'll need to modify the E2E_EXTRA_ARGS to remove the --progress-report-url. You can do this by adding `--plugin-env=e2e.E2E_EXTRA_ARGS=` to the command line.
```


johnSchnake:
Merging based on conversations offline.

PR #1350, opened 2021-07-04
Rework the gen plugin commands in order to reuse more code

johnSchnake:
- Adds a new method that allows us to easily reuse all the `gen`
logic and then just pull out the plugin we want.
- Removes the ability to specify configmaps to gen plugin e2e;
this seems like a trivial loss since I just added it in order
to achieve some other goal related to the windows plugins and
it never filled much of a real need.

This PR is meant to precede another related to how the k8s-version
flags are handled. In that PR, it broke the `gen plugin e2e` command
since this command wasn't really processing the plugin in the
same way that `gen` would.

**Release note**:
```
Removed the ability to specify configmaps to `gen plugin e2e`. That was a very recent addition, had no obvious benefit, and was counter to the general intent of the `gen plugin e2e` command which was to present you with the e2e plugin that `gen` would have
```

### DIFF
--- a/test/integration/testdata/gen-plugin-e2e-configmap.golden
+++ b/test/integration/testdata/gen-plugin-e2e-configmap.golden
@@ -1,7 +1,3 @@
-config-map:
-  tiny-configmap.yaml: |
-    a: 1
-    b: 2
 podSpec:
   containers: []
   nodeSelector:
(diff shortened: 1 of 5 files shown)


test/integration/testdata/gen-plugin-e2e-configmap.golden
```diff
@@ -24,21 +20,22 @@ spec:
   command:
   - /run_e2e.sh
   env:
+  - name: E2E_EXTRA_ARGS
```
johnSchnake, line 23: These are just properly sorted now

test/integration/testdata/gen-plugin-e2e-configmap.golden
```diff
@@ -24,21 +20,22 @@ spec:
...
   - name: E2E_PARALLEL
     value: "false"
+  - name: E2E_SKIP
+    value: \[Disruptive\]|NoExecuteTaintManager
   - name: E2E_USE_GO_RUNNER
     value: "true"
-  - name: E2E_EXTRA_ARGS
-    value: --progress-report-url=http://localhost:8099/progress
+  - name: SONOBUOY_K8S_VERSION
+    value: v123.456.789
```
johnSchnake, line 34: This value should have been added but it never was part of the unique `gen plugin e2e` flow

johnSchnake:
Merging based on conversations offline.

Issue #1292, opened 2021-06-08
Plugin caching/management


johnSchnake:
**Describe the solution you'd like**
It would be good to improve the way sonobuoy handles multiple plugins.

Currently users can run plugins from a file or URL which is great, but its not as if they are going to keep a list of URLs for the plugins they want. They can download them but if it is just a matter of downloading, storing, and updating them then sonobuoy should be able to do that.

We've talked about something like adding a `.sonobuoy` folder where users can do things like:

```
// download a plugin from a url
sonobuoy plugin get <url>

// run a plugin you have locally; its it .sonobuoy and not having to point to a file directly now
sonobuoy run -p foo

...
// list plugins we have locally
sonobuoy plugins list
```

Need a DD about this to flush out edge cases about updates or filenames/plugin names.

**Anything else you would like to add:**
Unsure if it would be possible or desireable to move the e2e plugin itself out of sonobuoy and support it this way too. Benefits would be:

 - could have a few preconfigured plugins instead of the `--mode` flag
 - ideally we get away from the static/dynamic plugin code. That is a mess and very confusing but was necessary to maintain historical functionality for all the flags
 - may get away from some extra flags; again that just makes the code easier to maintain and consistent between all the plugins

vladimirvivien:
Awesome start of a real plugin management feature in Sonobuoy @johnSchnake

johnSchnake:
Actively working on https://docs.google.com/document/d/1RzwYwI43TBFyHBxZ4mKDtfTAq6wJgXsfW7sFCHJjXxw/edit?usp=sharing . Anyone, feel free to comment.

PR #1356, opened 2021-07-12
Adds plugin installation functionality

johnSchnake:
Adds a known location, ~/.sonobuoy or SONOBUOY_DIR env,
which stores plugin files (*.yaml). When running plugins this location is searched
first so that users can more easily call custom plugins from any location.

Adds:
 - plugins list command
 - plugin show command
 - plugin install command
 - plugin uninstall command
 - tests for the above

Partially fixes: #1292

### DIFF
--- a/cmd/sonobuoy/app/pluginCache.go
+++ b/cmd/sonobuoy/app/pluginCache.go
@@ -1,0 +1,272 @@
+/*
+Copyright the Sonobuoy contributors 2021
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package app
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/vmware-tanzu/sonobuoy/pkg/plugin/loader"
+	"github.com/vmware-tanzu/sonobuoy/pkg/plugin/manifest"
+	kuberuntime "k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	SonobuoyDirEnvKey = "SONOBUOY_DIR"
+)
+
+var (
+	defaultSonobuoyDir = filepath.Join("~", ".sonobuoy")
+)
+
+func NewCmdPlugin() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "plugin",
+		Aliases: []string{"plugins"},
+		Short:   "Manage your installed plugins",
+		Hidden:  true,
+	}
+
+	listCmd := &cobra.Command{
+		Use:   "list",
+		Short: "List all installed plugins",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return listInstalledPlugins(getPluginCacheLocation(cmd))
+		},
+	}
+
+	showCmd := &cobra.Command{
+		Use:   "show <plugin filename>",
+		Short: "Print the full definition of the named plugin file",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return showInstalledPlugin(getPluginCacheLocation(cmd), filenameFromArg(args[0], ".yaml"))
+		},
+	}
+
+	installCmd := &cobra.Command{
+		Use:   "install <save-as-filename> <source filename or URL>",
+		Short: "Install a plugin so that it can be run via just its filename rather than a full path or URL.",
+		Args:  cobra.ExactArgs(2),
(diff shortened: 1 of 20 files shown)


johnSchnake:
Have basic list/show/install commands with an env override for the installation location.

TODO:
 - [ ] lots of tests
 - [ ] clarify help messages for usage
 - [ ] rework slightly so the methods dont output the CLI output. Ensure the functional code is in its own methods and the CLI stuff is separate.
 - [ ] more? Recheck the DD

johnSchnake:
Also added an uninstall command and am making the commands hidden for now. This way all this can merge before actually tying it into the actual plugin loading. So this is just a stealth feature a.t.m. where sonobuoy will save/list/show plugins but it doesn't impact runs at all.

The next PR can actually tie it into the loading of plugins (thats the part that we'll feature gate with an env var).

johnSchnake:
@vladimirvivien Since these are hidden commands at first we don't have to be perfect but I wanted eyes on this first main chunk of the functionality. Anything that doesn't make sense or that I failed to implement from the DD?

The followup PR will make these non-hidden and actually modify the loading of plugins. The point of merging this first is to get the bulk of the code out of the way so we can more easily code/review the implementation and tweaks on this foundational PR.

cmd/sonobuoy/app/pluginCache.go
```diff
@@ -0,0 +1,266 @@
...
+
+// getPluginCacheLocation will return the location of the plugin cache (defaults to ~/.sonobuoy)
+// If no override is set in the env, the default is assumed. If we are unable to expand to an abosolute
+// path then we return the empty string signalling the feature should be disabled.
+func getPluginCacheLocation(cmd *cobra.Command) string {
+	usePath := os.Getenv(SonobuoyDirEnvKey)
+	if len(usePath) == 0 {
+		usePath = defaultSonobuoyDir
+	}
+	expandedPath, err := homedir.Expand(usePath)
```
vladimirvivien, line 99: Suggestion: you could save a dependency here. If I recall correctly, that `home-dir` package essentially does the following.

```go
func ExpandPath(path string) (string, error) {
	home, err := os.UserHomeDir()
	if err != nil {
		return "", err
	}

	if path[0] == '~' {
		path = filepath.Join(home, path[1:])
	}
	return path, nil
}
```

cmd/sonobuoy/app/pluginCache.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		logrus.Errorf("failed to expand sonobuoy directory %q: %v", usePath, err)
+		return ""
+	}
+
+	if _, err := os.Stat(expandedPath); err != nil && os.IsNotExist(err) {
+		logrus.Debugf("sonobuoy plugin location %q does not exist, creating it.", expandedPath)
+		if err := os.Mkdir(expandedPath, 0777); err != nil {
+			logrus.Errorf("failed to create directory for installed plugins %q: %v", expandedPath, err)
```
vladimirvivien, line 108: Should this if-block `return ""` dir failed to be created ?

cmd/sonobuoy/app/pluginCache.go
```diff
@@ -0,0 +1,266 @@
...
+		first = false
+	}
+
+	return nil
+}
+
+// loadPlugins loads all plugins from the given directory (recursively) and
+// returns a map of their filename and contents. Returns the first error encountered
+// but continues to load and return as many plugins as possible.
+func loadPlugins(installedDir string) (map[string]*manifest.Manifest, error) {
```
vladimirvivien, line 139: Suggestion: maybe this should return a `[]error` to collect all errs as they happen (since the walk continues). Not sure if that would be useful for the caller of this func .

cmd/sonobuoy/app/pluginCache.go
```diff
@@ -0,0 +1,266 @@
...
+		if firstErr == nil {
+			firstErr = err
+		}
+		logrus.Errorf("error walking the path %q: %v\n", installedDir, err)
+	}
+
+	return pluginMap, firstErr
+}
+
+func loadPlugin(installedDir, reqFile string) (*manifest.Manifest, error) {
```
vladimirvivien, line 167: Could this be done from the pre-loaded `map[string]*manifest.Manifest` from above ?

vladimirvivien (commented):
This looks great @johnSchnake 
Left a few comments/suggestions.

cmd/sonobuoy/app/pluginCache.go
johnSchnake: Maybe I will; originally I used home-dir since it mentioned something about cgo but that was very outdated. https://github.com/mitchellh/go-homedir/issues/20#issuecomment-467476994 mentions the stdlib vs home-dir usage and I may just prefer the stdlib in our case.

cmd/sonobuoy/app/pluginCache.go
johnSchnake: I'm always a little split on that but I tend to usually prefer the 'returns first error' since programatically, in most cases, we just need to know error/pass and the user still gets logs for each error. Makes it easier/prettier to handle rather than a []error to handle.

cmd/sonobuoy/app/pluginCache.go
johnSchnake: I thought about doing it that way too; it could definitely leverage it if we wanted but the main difference was simply that this exits early if found. That's not a huge issue since users won't have 1 million plugins or anything. You're probably right that the simplification is preferred.

johnSchnake:
Fixed up from comments and added a new test to confirm that when listing plugins we continue despite failures (which we didnt actually do before).

Merging and working on the followup that actually ties this into sonobuoy logic

PR #1357, opened 2021-07-15
pkg/tarball: fix dropped test error

alrs:
Signed-off-by: Lars Lehtonen <lars.lehtonen@gmail.com>

This fixes a dropped test error.

**Release note**:
```
release-note NONE
```


johnSchnake:
Thanks for the catch and fix 👍

PR #1363, opened 2021-07-22
Loads plugins from installation directory if feature enabled 

johnSchnake:
Loads plugins from installation directory if feature enabled 

 - Adds feature gates via env var
 - Uses feature gate to make this new functionality optional
 - If enabled, will search in the installation directory for
the plugin first before looking in the pwd

Fixes: vmware-tanzu#1292

**Release note**:
```
Setting SONOBUOY_ALL_FEATURES or SONOBUOY_PLUGIN_INSTALLATION to `true` will enable sonobuoy to load plugins from the installation directory before looking in the pwd. It is expected that this will become the default behavior in the future.
```

### DIFF
--- a/cmd/sonobuoy/app/pluginCache.go
+++ b/cmd/sonobuoy/app/pluginCache.go
@@ -21,6 +21,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -38,12 +39,19 @@ var (
 	defaultSonobuoyDir = filepath.Join("~", ".sonobuoy")
 )
 
+// pluginNotFoundError is a custom type so we can tell whether or not loading the plugin
+// from the installation directory FAILED or if it just wasn't found.
+type pluginNotFoundError struct{ e error }
+
+func (p *pluginNotFoundError) Error() string {
+	return p.e.Error()
+}
+
 func NewCmdPlugin() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "plugin",
 		Aliases: []string{"plugins"},
 		Short:   "Manage your installed plugins",
-		Hidden:  true,
 	}
 
 	listCmd := &cobra.Command{
@@ -51,7 +59,7 @@ func NewCmdPlugin() *cobra.Command {
 		Short: "List all installed plugins",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return listInstalledPlugins(getPluginCacheLocation(cmd))
+			return listInstalledPlugins(getPluginCacheLocation())
 		},
 	}
 
@@ -60,7 +68,7 @@ func NewCmdPlugin() *cobra.Command {
 		Short: "Print the full definition of the named plugin file",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return showInstalledPlugin(getPluginCacheLocation(cmd), filenameFromArg(args[0], ".yaml"))
+			return showInstalledPlugin(getPluginCacheLocation(), filenameFromArg(args[0], ".yaml"))
 		},
 	}
 
@@ -69,7 +77,7 @@ func NewCmdPlugin() *cobra.Command {
 		Short: "Install a plugin so that it can be run via just its filename rather than a full path or URL.",
 		Args:  cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return installPlugin(getPluginCacheLocation(cmd), filenameFromArg(args[0], ".yaml"), args[1])
+			return installPlugin(getPluginCacheLocation(), filenameFromArg(args[0], ".yaml"), args[1])
 		},
 	}
 
@@ -78,7 +86,7 @@ func NewCmdPlugin() *cobra.Command {
 		Short: "Uninstall a plugin. You can continue to run any plugin via specifying a file or URL.",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return uninstallPlugin(getPluginCacheLocation(cmd), filenameFromArg(args[0], ".yaml"))
+			return uninstallPlugin(getPluginCacheLocation(), filenameFromArg(args[0], ".yaml"))
 		},
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/sonobuoy/app/pluginCache.go
```diff
@@ -38,20 +39,27 @@ var (
...
 		Short:   "Manage your installed plugins",
-		Hidden:  true,
 	}
 
 	listCmd := &cobra.Command{
 		Use:   "list",
 		Short: "List all installed plugins",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return listInstalledPlugins(getPluginCacheLocation(cmd))
```
johnSchnake, line 54: Thought I'd have a flag potentially to impact cache usage but opted out of that. Removed unnecessary arg.

cmd/sonobuoy/app/pluginCache.go
```diff
@@ -119,10 +127,16 @@ func listInstalledPlugins(installedDir string) error {
 	}
 
 	pluginFiles, _ := loadPlugins(installedDir)
+	filenames := []string{}
+	for filename, _ := range pluginFiles {
+		filenames = append(filenames, filename)
+	}
+	sort.StringSlice(filenames).Sort()
```
johnSchnake, line 134: Need to sort these to make the output predictable

test/integration/sonobuoy_integration_test.go
```diff
@@ -723,27 +689,99 @@ func TestPluginComplexCmds_LocalGolden(t *testing.T) {
...
+	os.Setenv("SONOBUOY_DIR", tmpDir)
+	os.Setenv("SONOBUOY_ALL_FEATURES", "true")
+
+	_, e := runSonobuoyCommandWithContext(ctx, t, "gen -p hello-world.yaml --kubernetes-version=v123.456.789")
+	if e == nil {
+		t.Fatal("Expected a failure since no plugin was installed but got none")
+	}
+
+	_ = mustRunSonobuoyCommandWithContext(ctx, t, "plugin install hello-world.yaml ./testdata/plugins/good/hello-world.yaml")
+	output := mustRunSonobuoyCommandWithContext(ctx, t, "gen -p hello-world.yaml --kubernetes-version=v123.456.789")
```
johnSchnake, line 721: Not the prettiest kind of tests, but it was unintutive to put this sort of thing into a TDT. Here we build up from no plugin, installed plugin, installed and local,  disabling the feature, and removing the local file.

The only difference (so we can tell which plugin we are loading) is via the source-url field which we swap out for the file in the pwd

test/integration/sonobuoy_integration_test.go
```diff
@@ -723,27 +689,99 @@ func TestPluginComplexCmds_LocalGolden(t *testing.T) {
...
-				fileData, err := ioutil.ReadFile(tc.expectFile)
-				if err != nil {
-					t.Fatalf("Failed to read golden file %v: %v", tc.expectFile, err)
-				}
-				if diff := pretty.Compare(string(fileData), outString); diff != "" {
-					t.Errorf("Expected manifest to equal goldenfile: %v but got diff:\n\n%v", tc.expectFile, diff)
-				}
-			}
-		})
+	installedPluginFile := "./testdata/plugin-loading-installed.golden"
```
johnSchnake, line 703: We repeatedly check against these files so saving them at the top.

test/integration/sonobuoy_integration_test.go
```diff
@@ -723,27 +689,99 @@ func TestPluginComplexCmds_LocalGolden(t *testing.T) {
...
+	localPluginFile := "./testdata/plugin-loading-local.golden"
+
+	tmpDir, err := ioutil.TempDir("", "sonobuoy_plugin_test_*")
+	if err != nil {
+		t.Fatal("Failed to create tmp dir")
+	}
+	defer os.RemoveAll(tmpDir)
```
johnSchnake, line 710: Tempdir rather than pointng to testdata. This avoids pollution of those files/folders if we install/uninstall anything

test/integration/sonobuoy_integration_test.go
```diff
@@ -723,27 +689,99 @@ func TestPluginComplexCmds_LocalGolden(t *testing.T) {
...
+
+	os.Setenv("SONOBUOY_DIR", tmpDir)
+	os.Setenv("SONOBUOY_ALL_FEATURES", "true")
```
johnSchnake, line 713: Need to have a defer here so it gets unset even if it fails in the middle of the test

test/integration/sonobuoy_integration_test.go
```diff
@@ -723,27 +689,99 @@ func TestPluginComplexCmds_LocalGolden(t *testing.T) {
...
+		t.Fatal("Expected a failure since no plugin was installed but got none")
+	}
+
+	// Copy file to pwd
+	input, err := ioutil.ReadFile("./testdata/plugins/good/hello-world.yaml")
+	if err != nil {
+		t.Fatalf("Failed to read plugin to test pwd loading")
+	}
+
+	err = ioutil.WriteFile("hello-world.yaml", bytes.Replace(input, []byte("foo.com"), []byte("localfile.com"), -1), 0644)
```
johnSchnake, line 736: This is the point of this test partially. We need to differentiate which plugin was loaded so they need to be unique. The one that is in the pwd will have this source instead fo the foo.com value.

cmd/sonobuoy/app/pluginList.go
```diff
@@ -171,3 +200,9 @@ func loadManifest(bytes []byte) (*manifest.Manifest, error) {
 	err := kuberuntime.DecodeInto(manifest.Decoder, bytes, &def)
 	return &def, errors.Wrap(err, "couldn't decode yaml for plugin definition")
 }
+
+// isNotExist returns true if the cause of the error was that the plugin did not exist.
+func isNotExist(e error) bool {
```
vladimirvivien, line 205: The doc for `pluginNotFoundErr` says
```
pluginNotFoundError is a custom type so we can tell whether or not loading the plugin from the installation directory FAILED or if it just wasn't found.
```
Could this not be done with Go io errors ?

test/integration/sonobuoy_integration_test.go
vladimirvivien: The tests look laborious, but it's a good thing to have them to boost confidence in code.

vladimirvivien (commented):
One comment. Looks good overall.

cmd/sonobuoy/app/pluginList.go
johnSchnake: > Go io errors

What do you mean by that? You usually either have an exact error you check against (e.g. `if err==notFoundErr`) or a type which you can typecast and check. The only other complication is with wrapped errors but that doesn't really change those two approaches.

vladimirvivien: Oh I see what not found means, it has nothing to do with actual file op, but a look up in the cache.  Never mind.

```go
func loadPlugin(installedDir, reqFile string) (*manifest.Manifest, error) {
	// Ignore error since it may not have to do with this plugin at all. Only return error if not found.
	// The loadPlugins will log errors as needed for visibility anyways.
	manMap, _ := loadPlugins(installedDir)
	reqManifest := manMap[filepath.Join(installedDir, reqFile)]
	if reqManifest != nil {
		return reqManifest, nil
	}
	return nil, &pluginNotFoundError{fmt.Errorf("failed to find plugin file %v within directory %v", reqFile, installedDir)}
}
```

vladimirvivien (approved):

Issue #1365, opened 2021-07-22
Edit docs links in GitHub are broken. 


salaboy:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Documentation links to edit the page in github are broken. 
https://sonobuoy.io/docs/v0.52.0/gen/

Edit in Github button on the top right corner is broken.

**What did you expect to happen:**

Be able to send a PR to the docs

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Sonobuoy version: 
- Kubernetes version: (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`): 
- Sonobuoy tarball (which contains * below)

PR #1366, opened 2021-07-22
Update sonobuoy.io button to edit docs and report issues

johnSchnake:
- Removes button to edit docs. Would work well since we have master
docs and versioned docs that we have to reconcile. In addition it was failing
to get the user email address anyways.
 - Fixes the link for reporting issues
 - Fixes the body/title for the issue being reported

Fixes: #1365


johnSchnake:
FYI @jonasrosland 

The .Title value wasn't working and the base URL was wrong. Updated values accordingly and just removed the button to edit it at all.

PR #1385, opened 2021-08-10
Add vulnerability scanning to CI

### LINKED ISSUE #1384: Do vulnerability scanning by default
Using a tool like trivy we could scan the resulting images from builds with little effort and avoid adding vulnerabilities into our images.

johnSchnake:
Some CVEs were reported to us and were found with trivy. They
have a githubAction we can utilize to trivially add this to our
CI flow, adding to the security of the resulting product.

- Fixes #1384

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -75,6 +75,23 @@ jobs:
           VERBOSE: true
         run: |
           source ./scripts/build.sh; stress
+  vulnerability-scan-linux:
+    runs-on: ubuntu-latest
+    needs: [build-linux, unit-test-linux]
+    steps:
+      - name: Download binaries and prebuilt images
+        uses: actions/download-artifact@v2
+        with:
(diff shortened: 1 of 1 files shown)


johnSchnake:
Started this PR from an old release so that we would be able to see the error case.

johnSchnake:
Tried to force a failure by starting from an old version, but CI will actually test as if you merged onto master so we were already picking up the fixes and getting 0 issues.

As a result, I added an import for go-etherium which had a vulnerability. My reason for choosing that one is that there is 0 confusion as to whether or not that code should be there. As opposed to if I just slightly downgraded something and accidentally merged it, which would be harder to spot.

johnSchnake:
See the failure here https://github.com/vmware-tanzu/sonobuoy/actions/runs/1117804690 ; now I'll remove the eth code and see it resolve.

johnSchnake:
CI is confused for some reason; leaving things as pending and when investigating it still shows the commit that I removed. Trying in trivial change/force push

Issue #1388, opened 2021-08-16
Missing Conformance tests skipped by disruptive with _certified-conformance_ mode


BobyMCbobs:
**What steps did you take and what happened:**
When I ran
```shell
sonobuoy run --mode=certified-conformance
```
Conformance tests that have the disruptive label were ignored with the _certified-conformance_ mode.

**What did you expect to happen:**
Disruptive tests should be run with the _certified-conformance_ mode as they are required by Conformance.

**Anything else you would like to add:**
In the sonobuoy-plugins-0 ConfigMap, there
```
$ kubectl -n sonobuoy get cm sonobuoy-plugins-cm -o=jsonpath='{.data.plugin-0\.yaml}' | yq e .spec.env[3] -P -
name: E2E_SKIP
value: \[Disruptive\]|NoExecuteTaintManager
```

I also noticed the logic
https://github.com/vmware-tanzu/sonobuoy/blob/master/cmd/sonobuoy/app/args.go#L457
for choosing the args

A workaround that I had was to run
```
sonobuoy run --mode=certified-conformance --plugin-env e2e.E2E_SKIP=''
```

**Environment:**

- Sonobuoy version: v0.53.1
- Kubernetes version: (use `kubectl version`): v1.21.2
- Kubernetes installer & version: `kind` v0.11.1
- Cloud provider or hardware configuration: nil
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04
- Sonobuoy tarball (which contains * below)

tedli:
I think it's because of this line,
https://github.com/vmware-tanzu/sonobuoy/blob/0f7c4a222f334c9b957843f36410713fa14f1946/cmd/sonobuoy/app/args.go#L203
`mode` is hardcoded set to `NonDisruptiveConformance` to act as a default behavior, however, this caused the `E2E_SKIP` env set.
Then by specifying `--mode` to `certified-conformance` could trigger mode set again, but without clearing the already set `E2E_SKIP`.

There goes this problem.

johnSchnake:
confirmed the issue; will patch asap and push a new release. Will check various versions to find when the regression landed (though I have a guess) and push a new version.

PR #1390, opened 2021-08-16
Fix regression regarding certified-conformance mode

johnSchnake:
When the mode is set as certified conformance mode, the skip value
does not get properly reset and as a result skips the disruptive
tests.

Fixes: #1388

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -466,8 +466,14 @@ func (m *Mode) Set(str string) error {
 		if err := m.env.Set(fmt.Sprintf("e2e.E2E_FOCUS=%v", quickFocus)); err != nil {
 			return fmt.Errorf("failed to set flag with value %v", str)
 		}
+		if err := m.env.Set("e2e.E2E_SKIP"); err != nil {
+			return fmt.Errorf("failed to set flag with value %v", str)
+		}
 	case CertifiedConformance:
 		if err := m.env.Set(fmt.Sprintf("e2e.E2E_FOCUS=%v", conformanceFocus)); err != nil {
+			return fmt.Errorf("failed to set flag with value %v", str)
+		}
+		if err := m.env.Set("e2e.E2E_SKIP"); err != nil {
 			return fmt.Errorf("failed to set flag with value %v", str)
 		}
 	default:
--- a/test/integration/sonobuoy_integration_test.go
+++ b/test/integration/sonobuoy_integration_test.go
@@ -495,7 +495,7 @@ func TestExactOutput_LocalGolden(t *testing.T) {
 			expectFile: "testdata/gen-variable-image.golden",
 		}, {
 			desc:       "gen doesnt provide UUID",
-			cmdLine:    "gen",
+			cmdLine:    "gen --kubernetes-version=ignore",
 			expectFile: "testdata/gen-no-uuid.golden",
 		}, {
 			desc:       "gen config doesnt provide UUID",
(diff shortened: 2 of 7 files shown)


test/integration/sonobuoy_integration_test.go
```diff
@@ -495,32 +495,36 @@ func TestExactOutput_LocalGolden(t *testing.T) {
...
+			cmdLine:    "gen --kubernetes-version=ignore",
 			expectFile: "testdata/gen-no-uuid.golden",
 		}, {
 			desc:       "gen config doesnt provide UUID",
 			cmdLine:    "gen config",
 			expectFile: "testdata/gen-config-no-uuid.golden",
 		}, {
 			desc:       "gen with config testing fields that also have flags",
-			cmdLine:    "gen --config=testdata/subfieldTest.json",
+			cmdLine:    "gen --config=testdata/subfieldTest.json --kubernetes-version=ignore",
```
johnSchnake, line 506: These tests were just failing locally because there was no cluster; I was trying to avoid that. I can remove this if needed, but I hated that when running `update_local` from the bash scripts it had an issue with numerous of these tests because it was previously catching that I had a v1.20.0 k8s cluster locally.

cmd/sonobuoy/app/args.go
```diff
@@ -466,10 +466,16 @@ func (m *Mode) Set(str string) error {
 		if err := m.env.Set(fmt.Sprintf("e2e.E2E_FOCUS=%v", quickFocus)); err != nil {
 			return fmt.Errorf("failed to set flag with value %v", str)
 		}
+		if err := m.env.Set("e2e.E2E_SKIP"); err != nil {
```
johnSchnake, line 469: Actual code change here and right below; just set remove those values if we don't want them.

Confirmed that setting them this way (without the `=foo` ending) will just remove the value.

test/integration/testdata/gen-config-then-flags.golden
```diff
@@ -96,8 +96,8 @@ data:
       - name: E2E_USE_GO_RUNNER
         value: "true"
       - name: SONOBUOY_K8S_VERSION
-        value: v1.20.0
-      image: k8s.gcr.io/conformance:v1.20.0
+        value: ignore
+      image: k8s.gcr.io/conformance:ignore
```
johnSchnake, line 100: The ignore mode is just for this sort of thing; when doing `gen` locally and you don't have a cluster to find. Rather than erroring out it just swallows those errors and sets this string to unblock things. Good for testing and hacking.

test/integration/testdata/gen-issue-1388.golden
```diff
@@ -0,0 +1,203 @@
...
+      driver: Job
+      plugin-name: e2e
+      result-format: junit
+    spec:
+      command:
+      - /run_e2e.sh
+      env:
+      - name: E2E_EXTRA_ARGS
+        value: --progress-report-url=http://localhost:8099/progress
+      - name: E2E_FOCUS
```
johnSchnake, line 90: You can see here, no skip value is set.

cmd/sonobuoy/app/args.go
```diff
@@ -466,10 +466,16 @@ func (m *Mode) Set(str string) error {
...
+			return fmt.Errorf("failed to set flag with value %v", str)
+		}
 	case CertifiedConformance:
 		if err := m.env.Set(fmt.Sprintf("e2e.E2E_FOCUS=%v", conformanceFocus)); err != nil {
 			return fmt.Errorf("failed to set flag with value %v", str)
 		}
+		if err := m.env.Set("e2e.E2E_SKIP"); err != nil {
+			return fmt.Errorf("failed to set flag with value %v", str)
```
vladimirvivien, line 477: Is this setting E2E_SKIP to v or is it clearing it?

vladimirvivien (commented):
Left single comment.

BobyMCbobs:
Woohoo :100: Thanks for the changes!
@johnSchnake

PR #1395, opened 2021-08-18
Absorb e2e functionality into rerun-failed flag

johnSchnake:
- Removes e2e command since its results analysis code is
duplicated by the newer results command and its rerun functionality
is being moved in this PR. The flags conflict and so rather than
wire them to work at the same time, I wanted to simplify things
by removing a code path with little value.
 - Adds --rerun-failed flag to gen/run
 - Copies data for relevant tests cases: no failures, not a tarball,
happy path, and missing results.

**Release note**:
```
The `sonobuoy e2e` command has been removed. To inspect tarball results use `sonobuoy results` and for rerunning the failed tests, a new flag `--rerun-failed` has been added to `sonobuoy gen` and `sonobuoy run`. To use the flag, pass the tarball path to that new flag. This rearrangement of code means that you can verify what will be submitted when trying to rerun tests (via gen).
```


johnSchnake:
It looks by the LOC like this is a wash, but most of the 'added' lines are the .golden files.

site/content/docs/master/snapshot.md
```diff
@@ -28,8 +28,6 @@ The results of plugins can be inspected without being extracted. By default, it
 sonobuoy results $output [--plugin <name>] [--mode report|detailed|dump]
```
vladimirvivien, line 28: Curious:
In this command example it looks like  `--plugin` and `mode` are optional.
If a plugin is not provided, does it assume the `e2e` by default ?

test/integration/testdata/gen-rerunfailed-missing.golden
```diff
@@ -0,0 +1,44 @@
+Error: invalid argument "testdata/results-missing-e2e.tar.gz" for "--rerun-failed" flag: failed to find file "plugins/e2e/sonobuoy_results.yaml" in archive
```
vladimirvivien, line 1: I always wonder about the golden files.  They seem like a like a lot of work to maintain.  I guess they make validation easy.

vladimirvivien (commented):
Super stoked that this is taking shape and the UX is getting simpler!!
Looks good. Left couple comments.

site/content/docs/master/snapshot.md
johnSchnake: No, it prints the info for each plugin if not specified.

So if you try `sonobuoy results tarball` on a default sonobuoy run, you'll see a little summary of e2e and the systemd-logs.

test/integration/testdata/gen-rerunfailed-missing.golden
johnSchnake: They are _soooooooooooooooooooo_ easy to maintain in comparison to other types of tests.

Its very easy to make new test cases: copy/paste the test case struct and just change the details (name of file, command to test, etc). Then run `update_local` and check that it looks good.

Then no matter what UX or logic changes happen, if you want to make a small or big change you just have to run the `update_local` command again and verify the output looks good.

Otherwise you're stuck manually trying to verify random parts of large output and miss lots of little things too. For instance, this PR was failing the goldenfile tests at first because the help menu apparently was listing the `image-pull-policy` values but not sorting them so about 1/3 of the time they'd swap order.

They are definitely verbose, but since you dont write/maintain the contents _manually_ they are so nice.

johnSchnake: If you werent aware, I dont ever run commands and copy/paste into those files. The tests are made to handle an `--update` flag. If that it set then rather than testing the output against the goldenfile, it saves its output into the goldenfile.

So you can write 40 test cases of output you expect to be OK, run the tests with the update flag (use the update_local script function), and it creates all those files for you. After a quick look, you have documented the exact output of 40 test cases.

For `gen` commands especially this is really nice since the muli-yaml document would be painful to have to try and parse for certain details (there may be some tests that still do that and I think (1) they miss important data that could be tested and (2) I think they comment about how awkward it is)

vladimirvivien: @johnSchnake that is great, thanks for the explanation. 
I never made it into that part of the code, maybe at some point we can do a walk through so I can understand it a bit.

vladimirvivien (approved):

PR #1398, opened 2021-08-24
Use experimental features in test

johnSchnake:
Tweaks integration tests to use our experimental features by default
and therefore updates misc parts of the test file to set/pass
the appropriate env vars.


cmd/sonobuoy/app/pluginCache.go
```diff
@@ -274,12 +274,11 @@ func uninstallPlugin(installedDir, filename string) error {
...
-	home, err := os.UserHomeDir()
-	if err != nil {
-		return "", err
-	}
-
 	if len(path) > 0 && path[0] == '~' {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return "", err
+		}
```
johnSchnake, line 281: The way it was before would cause errors if HOME wasn't set even if the user was using a path that didnt include `~`

test/integration/testdata/plugin-list-good-bad.golden
```diff
@@ -1,7 +1,13 @@
...
+time="STATIC_TIME_FOR_TESTING" level=error msg="failed to load definition from file \"*STATIC_FOR_TESTING*/p4.yaml\": couldn't load plugin definition for file *STATIC_FOR_TESTING*/p4.yaml: couldn't decode yaml for plugin definition: couldn't get version/kind; json parse error: json: cannot unmarshal string into Go value of type struct { APIVersion string \"json:\\\"apiVersion,omitempty\\\"\"; Kind string \"json:\\\"kind,omitempty\\\"\" }"
 Installed plugin plugin into file *STATIC_FOR_TESTING*/p1.yaml from source ./testdata/plugins/good/hello-world.yaml
+time="STATIC_TIME_FOR_TESTING" level=error msg="failed to load definition from file \"*STATIC_FOR_TESTING*/p2.yaml\": couldn't load plugin definition for file *STATIC_FOR_TESTING*/p2.yaml: couldn't decode yaml for plugin definition: couldn't get version/kind; json parse error: json: cannot unmarshal string into Go value of type struct { APIVersion string \"json:\\\"apiVersion,omitempty\\\"\"; Kind string \"json:\\\"kind,omitempty\\\"\" }"
```
johnSchnake, line 6: This change seems to be correct for now but I'd like to tweak that in a follow up. Here, the output is from when we are installing a different plugin. If we are installing a well formed plugin, there is no reason we should be trying to load all the bad plugins and erroring about them.

Still an experimental feature so I dont mind merging with this as-is though

test/integration/sonobuoy_integration_test.go
```diff
@@ -49,14 +53,21 @@ func findSonobuoyCLI() (string, error) {
...
 
 	command := exec.CommandContext(ctx, sonobuoy, strings.Fields(args)...)
 	command.Stdout = &combinedOutput
 	command.Stderr = &combinedOutput
 
-	t.Logf("Running %q\n", command.String())
+	// Test with features on so we can get more feedback. Low risk that
+	// this will hide default behavior but in that case we may just make the
+	// experimental feature the default then.
+	command.Env = []string{"SONOBUOY_ALL_FEATURES=true", "KUBECONFIG=" + os.Getenv("KUBECONFIG"), "HOME=" + testHome}
```
johnSchnake, line 66: This seems also to be a good exercise in realizing the requirements of sonobuoy/our tests.

Had failed to explicitly copy kubeconfig here and so test run failed to connect to cluster.

FYI the reason we have to do all this is because if you _dont_ set the command env then it uses the current os env. Since we are setting it for each command, we have to set everything, we aren't just adding to the os.env

PR #1402, opened 2021-08-27
Replace master with main

johnSchnake:
In numerous scripts/docs we still use `master` instead of `main`.

This change should finish the conversion in language.


johnSchnake:
Green tests; going to merge. May find some broken links in the future but we'll just patch them as we go. I think this hit most , if not all, of what we needed.

PR #1421, opened 2021-09-17
 Add sleep parameter to job plugins worker container

### LINKED ISSUE #1415: Why aren't plugin logs gathered?
Run a plugin; check results tarball for pod logs and you typically only find the `sonobuoy` aggregator in the sonobuoy namespace. Where are the plugins?

They're already cleaned up.

I noticed this a few times in the past but I sweat it didn't used to be like that. Potentially it was an unrealized race condition? We should wait to cleanup the pods until we get the logs, not just exit as soon as it finishes. I think this is an option on the sonobuoy worker sidecar.

johnSchnake:
Without this, the container exits after submitting results
and leads to the pod being shut down before we can query
the logs. In some plugins (like e2e) it isn't that big of
an issue, but for plugins that don't separately save their
logs or if we want to debug the worker container, it is
vital.

Fixes #1415 

### DIFF
--- a/cmd/sonobuoy/app/worker.go
+++ b/cmd/sonobuoy/app/worker.go
@@ -57,12 +57,16 @@ func NewCmdWorker() *cobra.Command {
 }
 
 func newGlobalCmd() *cobra.Command {
+	var sleep int64
+
 	cmd := &cobra.Command{
 		Use:   "global",
 		Short: "Submit results scoped to the whole cluster",
-		RunE:  runGatherGlobal,
+		RunE:  runGatherGlobal(&sleep),
 		Args:  cobra.ExactArgs(0),
 	}
+
+	cmd.Flags().Int64Var(&sleep, "sleep", -1, "After sending results, keeps the process alive for N seconds to avoid restarting the container. If N<0, Sonobuoy sleeps forever.")
 
 	return cmd
 }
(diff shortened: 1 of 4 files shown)


johnSchnake:
Timeouts on the real integration runs now. Debugging; may be some broken cleanup that keeps the aggregator running since the plugins didnt exit themselves.

johnSchnake:
Ah! apparently, when the change went in we added the flag to the `sonobuoy worker node` command and not the `global` one that the jobs use. So the flag was missing, the worker would fail to start right, results would never get sent to the aggregator, hence the hang.

johnSchnake:
Changing this to draft since I had gone such a rabbit hole here; trying just the UID change here https://github.com/vmware-tanzu/sonobuoy/pull/1425 and moved a bit of a dev helper to #1424 . I also re-remembered that I couldn't remove the bash-sleeps from the daemonsets (see notes on #1300 )

johnSchnake:
I dont know that this is necessary; need to check but I think the pod enters the terminated state but we can still get the logs.

If we keep the worker alive arbitrarily then it reads as `NotReady` since one container is terminated and the other not. That seems more confusing.

johnSchnake:
Confirmed ^ is explicitly true; closing.

PR #1422, opened 2021-09-17
Add default security context to aggregator

### LINKED ISSUE #1410: Should specify security context
From a conversation with @btrieger :

> Most of the admission controllers for pod security check to see that a securityContext is set as they can't see what is in the dockerfile itself...

He ran into Sonobuoy being blocked by OPA Gatekeeper. This should be able to be resolved if we simply explicitly state the RunAsUser in the podspec for the aggregator.

johnSchnake:
There is no elevetated priviledge that the aggregator needs on the node
so we should be fine to add this security context which makes it
clear that we are not operating as root in any sense.

Fixes #1410

** Notes **
In addition to CI, I ran this locally against kind and it worked fine on quick mode. I was able to get/download/inspect results without issue.

**Release note**:
```
The aggregator pod is now run with a security context specified which explicitly makes it run as non-root user/group/fsgroup. This should avoid sonobuoy being blocked by certain security sweeps.
```

### DIFF
--- a/pkg/client/gen.tmpl.yaml
+++ b/pkg/client/gen.tmpl.yaml
@@ -102,6 +102,10 @@ metadata:
 {{- end }}
 {{- end }}
 spec:
+  securityContext:
+    runAsUser: 1000
+    runAsGroup: 3000
+    fsGroup: 2000
 {{- if .NodeSelectors }}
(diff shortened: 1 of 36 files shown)


vladimirvivien (approved):
LGTM

Issue #1426, opened 2021-09-17
Should we simplify the config and get more query data


johnSchnake:
```
 level=info msg="networksets not specified in non-nil Resources. Skipping networksets query."
time="2021-06-23T10:48:25Z" level=info msg="secrets not specified in non-nil Resources. Skipping secrets query."
time="2021-06-23T10:48:25Z" level=info msg="events not specified in non-nil Resources. Skipping events query."
time="2021-06-23T10:48:25Z" level=info msg="horizontalpodautoscalers not specified in non-nil Resources. Skipping horizontalpodautoscalers query."
time="2021-06-23T10:48:25Z" level=info msg="e2e-test-crd-publish-openapi-2339-crds not specifi
```

Lots of resources like that. I made the changes so that we dont have to explicitly list every data type we want to get. I we just provide the empty list I think we get everything (except secrets).

This would make the config much more grockable and then we'd get all the things!

Trivial change I think, just change the default file and golden files. Need to see the impact on size though.

johnSchnake:
So I ran a few cases and compared them:

```
-rw-r--r--   1 jschnake  staff   468K Sep 18 08:39 allandkubesystem.tar.gz
-rw-r--r--   1 jschnake  staff   419K Sep 18 08:30 allthethings.tar.gz
-rw-r--r--   1 jschnake  staff   342K Sep 18 08:32 defaultthings.tar.gz
-rw-r--r--   1 jschnake  staff   333K Sep 18 08:34 noresources.tar.gz
```

No resources to our default is ~10k
Default to all resources is 67k
Adding kube-system logs added 50k

By percentages the 50k and 67k is non-trivial but:
1. we're talking about KB, not MB or GB. No one will care.
2. These were quick mode results so the e2e logs are MUCH shorter than a full run. A larger run is consumed by the plugin logs anyways so these do become trivial values.

I say we make both changes so that the query is, by default, the most useful without sacrificing security (want to make sure its clear that we are explicitly not getting secrets).

Issue #1416, opened 2021-09-16
Consider getting kube-system logs by default.


johnSchnake:
Most users dont realize that you can tweak things about the queries that sonobuoy runs. E.g. you can get the logs for kube-system which are really useful when debugging scheduling/timeout issues.

Would it take up that much space to get them? Not sure of the size increase but its a good thing to try.

johnSchnake:
Related to #1426 in that they both try and get more good, useful info into the tarball

PR #1429, opened 2021-09-18
Increase default query data

johnSchnake:
Adding kube-system to the logs that we get as well as
defaulting to getting more of the API resources.

Fixes: #1426
Fixes: #1416


pkg/config/config.go
```diff
@@ -66,53 +66,13 @@ const (
 var (
 	// DefaultImage is the URL of the docker image to run for the aggregator and workers
 	DefaultImage = "sonobuoy/sonobuoy:" + buildinfo.Version
+
 	// DefaultResources is the default set of resources which are queried for after plugins run. The strings
 	// are compared against the resource.Name given by the client-go discovery client. The non-standard values
 	// that are included here are: podlogs, servergroups, serverversion. The value 'nodes', although a crawlable
 	// API value, also is used to query against the healthz and configz endpoints on the node.
-	DefaultResources = []string{
```
vladimirvivien, line 73: Are you dynamically discovering the resources, is that why these are no longer needed?

vladimirvivien (approved):
LGTM

pkg/config/config.go
johnSchnake: Yes we dynamically walk the API server. That code isn't easy to follow but thats been done for a long time. The difference is that prior to this change we dynamically discover all the ytpes and then only get this explicit list of types. Now we'll default to getting them all (except for secrets).

PR #1430, opened 2021-09-19
Removed managed fields from query output

### LINKED ISSUE #1428: Unnecessary empty values in yaml
In the queried values; lots are empty and should be ignored. Is there a simple way to ignore them in the serialization?

E.g.
```
"fieldsV1": {
            "f:count": {},
            "f:firstTimestamp": {},
            "f:involvedObject": {
              "f:apiVersion": {},
              "f:kind": {},
              "f:name": {},
              "f:namespace": {},
              "f:resourceVersion": {},
              "f:uid": {}
            },
            "f:lastTimestamp": {},
            "f:message": {},
            "f:reason": {},
            "f:source": {
              "f:component": {}
```

johnSchnake:
It is a confusing waste of info and makes it much harder to look
at the output. In addition, it is not normally provided by kubectl
so users aren't even aware that it exists (I wasn't) and so it
makes the rest of the data seem more confusing than it needs to be.

Fixes #1428

You can use this command to test locally if you want:

```
sonobuoy run --mode=quick --sonobuoy-image=schnake/sonobuoy:devonly1632013611 --wait && sonobuoy retrieve -x nomanagedoutput
```

**Release note**:
```
Removes managedfields from the query data since this is generally meaningless to any user, and very often mostly empty, nested fields that clog up the meaningful data.
```

### DIFF
--- a/pkg/discovery/queries.go
+++ b/pkg/discovery/queries.go
@@ -63,7 +63,7 @@ func timedListQuery(outpath string, file string, f listQuery) (time.Duration, er
 	}
 
 	if len(list.Items) > 0 {
-		err = errors.WithStack(SerializeObj(list.Items, outpath, file))
+		err = errors.WithStack(SerializeObj(list, outpath, file))
 	}
 	return duration, err
 }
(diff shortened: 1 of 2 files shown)


vladimirvivien (approved):
LGTM

PR #1439, opened 2021-09-22
Default our experimental features to true

### LINKED ISSUE #1436: Turn on extra wait output as default
An experimental feature that I've used a fair amount and it is useful in most cases.

People may complain about some of the deletion output since it shows the namespace object itself which has lots of info that _most_ people wont need. However, its easy to ignore and still shows progress.

### LINKED ISSUE #1437: Turn on the plugin management features by default
Experimental feature that I've tinkered with successfully and haven't seen any bad bugs. I'd consider this still in a sort of beta feature and as people use it we may have some bugs reported.

This is a very new feature and will require a blog post and help docs to describe its use.

johnSchnake:
Added logic so users can turn them off if it breaks
their logic in some way.

Fixes #1436
Fixes #1437

**Release note**:
```
Two new experimental features have now been enabled by default:
 - When running `sonobuoy run` or `sonobuoy delete` with the `--wait` flag, output is more verbose and shows you information about the ongoing process. It can be disabled by setting `SONOBUOY_WAIT_PROGRESS=false`
 - New commands have been added to address plugin management. This allows you to store plugin configurations and reference them more easily, as opposed to always requiring a full path to the file or URL. It can be disabled by setting `SONOBUOY_PLUGIN_INSTALLATION=false`.
```

### DIFF
--- a/cmd/sonobuoy/app/featureGates.go
+++ b/cmd/sonobuoy/app/featureGates.go
@@ -10,6 +10,24 @@ const (
 	FeatureWaitOutputProgressByDefault = "SONOBUOY_WAIT_PROGRESS"
 )
 
+var (
+	featureDefaultMap = map[string]bool{
+		FeaturePluginInstallation:          true,
+		FeatureWaitOutputProgressByDefault: true,
+	}
+)
+
+// featureEnabled returns if the named feature is enabled based on the current env and defaults.
 func featureEnabled(feature string) bool {
-	return os.Getenv(FeaturesAll) == "true" || os.Getenv(feature) == "true"
+	return featureEnabledCore(feature, os.Getenv(FeaturesAll), os.Getenv(feature), featureDefaultMap)
+}
+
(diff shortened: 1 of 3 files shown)


cmd/sonobuoy/app/featureGates.go
```diff
@@ -10,6 +10,18 @@ const (
...
+)
+
 func featureEnabled(feature string) bool {
-	return os.Getenv(FeaturesAll) == "true" || os.Getenv(feature) == "true"
+	// Allow features we default as true to be turned off while still relatively new so if major
+	// bugs are found we have workarounds.
+	if os.Getenv(feature) == "false" {
+		return false
+	}
+	return featureDefaultMap[feature] || os.Getenv(FeaturesAll) == "true" || os.Getenv(feature) == "true"
```
johnSchnake, line 26: I'm going to add a small unit test for this. Super easy to test and without it super easy to miss one of the variations.

PR #1449, opened 2021-10-04
Add flag/config setting for security context

### LINKED ISSUE #1444: Windows issues with security context
The new security context ends up causing an issue when trying to launch windows pods due to https://github.com/kubernetes/kubernetes/issues/102849

Before the next release, we need to decide how to mitigate the problem for sonobuoy users. E.g. add a temporary flag to remove that security context or something.

johnSchnake:
We recently added security context to prevent policy from worrying
about Sonobuoy being a priviledged user. However, k8s for Windows nodes
does not support those settings. As a result we needed a change like this
to make it easier to un-set the default values.

I imagine we will add a few more modes here (e.g. windows) to round out
the functionality but currently we need just the default and `none` which
is the desired value currently for windows.

Fixes #1444

**Release note**:
```
When the Sonobuoy aggregator may land on a Windows node, add the `--security-context-mode=none` flag. This prevents the default security context values which are incompatible with Windows nodes.
```


johnSchnake:
This isn't that big of a change, just plumbing the value through config settings and a flag. Its just that it updates a lot of golden files since it is a new config value.

vladimirvivien (approved):
LGTM

PR #1472, opened 2021-10-18
Added new flag for adding new configurations for permissions

### LINKED ISSUE #1470: Enable running custom plugins without an admin kubeconfig
We (tagging @janipee) have a set of tests that we have added to custom plugins for sonobuoy. The tests themselves can be run with e.g. the default `edit` ClusterRole. These tests would need to be run in clusters where we do not have cluster-admin access (RBAC is enabled), and certain namespaces would be created for our use beforehand (i.e. we cannot create and delete namespaces).

Sonobuoy currently lists an admin `kubeconfig` in its prerequisites, but assuming that the only tests/plugins to be run are the custom ones mentioned previously (i.e. e2e and systemd-logs can be ignored), would it be feasible to introduce a non-admin mode to enable runs on locked-down clusters?

E.g. currently sonobuoy attempts to create its own namespace along with a ClusterRole containing
```
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - '*'
- nonResourceURLs:
  - '/metrics'
  - '/logs'
  - '/logs/*'
  verbs:
  - 'get'
```
Could this be made dependent on something like an `adminMode` parameter?

If this proposal sounds reasonable, we could work on it and submit a PR.

**Environment:**

- Sonobuoy version: v0.53.2
- Kubernetes version: (use `kubectl version`): Varies - v1.20 onwards

NikhilSharmaWe:
Signed-off-by: Nikhil Sharma nikhilsharma230303@gmail.com

I have added a new flag, for adding new configuration for permissions which are talked about in issue #1470
- Fixes #1470 

**Release note**:
```
release-note
```


NikhilSharmaWe:
Will the process to proceed will be similar to the steps you took in this [PR](https://github.com/vmware-tanzu/sonobuoy/pull/1449/) to add a new flag security-context-mode and its two values nonroot and none.

johnSchnake:
Yes, nearly identical except for what exactly it changes in the template yaml file.

NikhilSharmaWe:
Why only "nonroot" is added in this function. Why there is not another function for "none" mode. Also what is the difference between .Var and .StringVar function.  
https://github.com/johnSchnake/sonobuoy/blob/37348cf9e056bda20d1e0bfa33b9127499289499/cmd/sonobuoy/app/args.go#L172

johnSchnake:
That is just the definition of the flag; the default is specified there as `nonroot`, it isn't about defining allowed values.

Var and StringVar are just two different ways to define flags; one is for a flag that has a string value, `Var` is more generic and can handle flag types with custom logic (e.g. when the flag is set, dont just copy the value given but do something like look up a file or run some validation on the value)

NikhilSharmaWe:
@johnSchnake is the code added till now is reasonable. If it is what should be added in pkg/client/gen.tmpl.yaml and pkg/client/gen.template.go

NikhilSharmaWe:
@johnSchnake is the changes made till now are reasonable. And If it is, what should be added in pkg/client/gen.tmpl.yaml and pkg/client/gen.template.go and how to proceed.

NikhilSharmaWe:
What is the significance of this [statement](https://github.com/vmware-tanzu/sonobuoy/pull/1449/files#diff-89328cfdd33cc64f0c6bac6d6eba0c346c83b7090a6ce04f4c79520589473b1aR249) in your PR for adding security-context flag. Also, how can we implement the same task for AggregatorPermissons flag.

NikhilSharmaWe:
@johnSchnake made changes similar to you PR for security context. Are these changes correct.
If they are correct, I am unable to figure out what changes will be made in gen.tmpl.yaml and template.go. Also you made a whole new [file](https://github.com/johnSchnake/sonobuoy/blob/37348cf9e056bda20d1e0bfa33b9127499289499/test/integration/testdata/gen-security-context-none.golden) for adding security context, what is its use, also does something similar is required also for aggregator permissions.

johnSchnake:
@NikhilSharmaWe you'll need to sign your commit; if its the last commit all you need to do is use `git amend --signoff` but if its a commit or two ago you need to just do an interactive rebase and edit the unsigned commit.

johnSchnake (approved):

johnSchnake:
Congrats on the first PR to Sonobuoy 🎉

PR #1537, opened 2021-11-30
updated gen-aggregator-permissions-namespaced.golden

### LINKED ISSUE #1536: Aggregator permissions issue?
Trying to utilize the new aggregator permissions flag:

```
$ sonobuoy run --aggregator-permissions=namespaced --mode=quick

INFO[0000] create request issued                         name=sonobuoy namespace= resource=namespaces
INFO[0000] create request issued                         name=sonobuoy-serviceaccount namespace=sonobuoy resource=serviceaccounts
ERRO[0000] error attempting to run sonobuoy: failed to create object: failed to create API resource sonobuoy-serviceaccount-sonobuoy: the server does not allow this method on the requested resource
```

We did gen tests but may not have actually tried runs. Need to dive in; probably a small change to what the template does in the namespaced option.

Using gen, I think the issue is that even though we change the `kind` of the role, `roleref` still shows this:
```
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
```
so we probably just need to adjust that as well.

@NikhilSharmaWe

NikhilSharmaWe:

- Fixes #1536 

**Release note**:
```
release-note
```

### DIFF
--- a/test/integration/testdata/gen-aggregator-permissions-namespaced.golden
+++ b/test/integration/testdata/gen-aggregator-permissions-namespaced.golden
@@ -21,7 +21,7 @@ metadata:
   name: sonobuoy-serviceaccount-sonobuoy
 roleRef:
   apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
+  kind: RoleBinding
   name: sonobuoy-serviceaccount-sonobuoy
 subjects:
 - kind: ServiceAccount
@@ -240,4 +240,3 @@ spec:
   selector:
     sonobuoy-component: aggregator
   type: ClusterIP
-


NikhilSharmaWe:
@johnSchnake what could be the reason of test failing here.

johnSchnake:
@NikhilSharmaWe I'm out for the week so you can try and debug if you want. There were pretty obvious issues once I actually tried `sonobuoy run....` instead of `sonobuoy gen...` with that flag. For instance, we tried to create a namespace even though we don't have that permission in the namespaced situation.

Also, preflight checks fail if we try and run sonobuoy in a preexisitng namespace so we need to document that specifying that flag requires `--skip-prefilight` to be set as well.

johnSchnake:
I think the `gen` logic is cleaned up and now we just have the more systemic issues about how to run sonobuoy with fewer permissions. Closing.

PR #1544, opened 2021-12-15
Remove template and convert to code

### LINKED ISSUE #1545: Cleanup: Convert template to code
The gen.tmpl.yaml file is increasingly hard to read and modify. Simple changes to hardcoded values are easy, but if you need anything more than that you have to:
 - add a new field to the templateValues type
 - assign the new value to the object before processing
 - look up the go template format and deal with whitespaces in complex cases (loops, branches, etc)

If we convert it all to code then:
- logic branches are easier to understand 
- error handling more explicit and customizable
- logic is broken into much smaller, simpler chunks

The namespaced permissions issues (#1470) made this very painful since multiple parts of the template were changing due to a new variable.

johnSchnake:

The template was getting increasingly complex due to various
options. Even though dealing with the types can be verbose,
it allows us to better break up each object and apply arbitrary
logic with better formatting.

- removed template
- create each object and serialize it manually

The golden files were updated only in in the following ways:
- files end with '---' rather than start with it
- YAML quoting is dictated by the yaml pkg and not manually by us
- fields were ordered alphabetically as intended
- YAML arrays were written as dictated by internals and not by us

Related to #1541 
Fixes #1545

Lots of lines changed but you can see that they are all the same few changes over and over.

**Release note**:
```
N/A
```


johnSchnake:
Done for the night but I think this is ready. The only real changes to the golden files are surrounding quoting and ordering of fields.

test/integration/testdata/gen-config-then-flags.golden
```diff
@@ -25,7 +24,6 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: sonobuoy-serviceaccount
-  namespace: cmdlineNS
```
johnSchnake, line 28: Investigate this; can't repro at first glance

test/integration/testdata/gen-issue-1375.golden
```diff
@@ -25,7 +24,6 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: sonobuoy-serviceaccount
-  namespace: sonobuoy
```
johnSchnake, line 28: Similar as another change; not sure why SA is not getting a namespace here

johnSchnake: Oh this is in the clusterrole ref, not the SA. Will fix.

test/integration/testdata/gen-issue-1528.golden
```diff
@@ -109,8 +106,7 @@ data:
         image: foo/baz:0.1.2
         imagePullPolicy: Always
         name: sidecar
-        resources: {}
-        volumeMounts:
+          volumeMounts:
```
johnSchnake, line 109: Issue with the lines being replaced and leaving extra spaces

pkg/client/testdata/aggregatorpermissions-noncluster-admin.golden
```diff
@@ -185,6 +186,7 @@ metadata:
   labels:
     component: sonobuoy
     sonobuoy-component: aggregator
+    tier: analysis
```
johnSchnake, line 189: Currently confused about these lines; the template had them but it shows I'm adding them now. Want to figure that out before merging.

johnSchnake: Oh this is on the service not the pod; thats what was added. Removing that.

PR #1566, opened 2021-12-28
Continue with run even if unable to query nodes

### LINKED ISSUE #1470: Enable running custom plugins without an admin kubeconfig

### LINKED ISSUE #1542: Namespaced issue running plugins
When trying to run plugins, the code tries to get the cluster nodes. This is incompatible with the namespaced permissions.

We may need to have the config store the node info or something. Get the node list early, maybe from the client, and then persist it in the config.

I'm not sure if there are other permission issues with the run itself (as opposed to the queries which are completely broken with the permissions)

johnSchnake:
Early in the run we gather nodes so that we know what results to expect
for the daemonsets. The new logic is:
 - log a warning that this failed and indicate the daemonsets will not work
 - since we dont have node names we dont know what results to expect/wait for
for daemonsets, so we just dont expect or wait for daemonsets at all
 - when doing pod log queries, even though we cant query for all namespaces,
we know our current namespace and can query that if specified
 - when doing resource queries, we likewise query those for the current
namespace

I think this should resolve the two issues below:

Fixes #1542 
Fixes #1470 

I am able to get runs started and even get some extra debug info like podlogs (for the test namespace) and API info for objects in that namespace too.

### DIFF
--- a/pkg/discovery/query.go
+++ b/pkg/discovery/query.go
@@ -58,7 +58,7 @@ func QueryCluster(restConf *rest.Config, cfg *config.Config) error {
 		return errors.Wrap(err, "could not create kubernetes client")
 	}
 
-	// 1. Create the directory which will store the results, including the
+	// Create the directory which will store the results, including the
 	// `meta` directory inside it (which we always need regardless of
 	// config)
 	outpath := cfg.QueryOutputDir()
@@ -68,14 +68,7 @@ func QueryCluster(restConf *rest.Config, cfg *config.Config) error {
 		return errors.Wrap(err, "could not create directory to store results")
 	}
 
-	// 2. Get the list of namespaces and apply the regex filter on the namespace
-	logrus.Infof("Filtering namespaces based on the following regex:%s", cfg.Filters.Namespaces)
-	nslist, err := FilterNamespaces(kubeClient, cfg.Filters.Namespaces)
-	if err != nil {
-		return errors.Wrap(err, "could not filter namespaces")
-	}
-
-	// 5. Run the queries
+	// Run the queries
 	recorder := NewQueryRecorder()
 	clusterResources, nsResources, err := getAllFilteredResources(apiHelper, cfg.Resources)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/discovery/query.go
```diff
@@ -68,14 +68,7 @@ func QueryCluster(restConf *rest.Config, cfg *config.Config) error {
 		return errors.Wrap(err, "could not create directory to store results")
 	}
 
-	// 2. Get the list of namespaces and apply the regex filter on the namespace
```
johnSchnake, line 71: This section just got moved, not deleted.

pkg/discovery/query.go
```diff
@@ -94,39 +87,51 @@ func QueryCluster(restConf *rest.Config, cfg *config.Config) error {
 		logrus.Errorf("Failed to query server data: %v", err)
 	}
 
+	// Get the list of namespaces and apply the regex filter on the namespace
+	logrus.Infof("Filtering namespaces based on the following regex:%s", cfg.Filters.Namespaces)
+	nslist, err := FilterNamespaces(kubeClient, cfg.Filters.Namespaces, cfg.Namespace)
```
johnSchnake, line 92: New param in this function allows for some fallback logic to consider that hardcoded namespace name.

pkg/plugin/aggregation/run.go
```diff
@@ -80,7 +81,11 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 	// TODO: there are other places that iterate through the CoreV1.Nodes API
 	// call, we should only do this in one place and cache it.
 	nodes, err := client.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
-	if err != nil {
+	switch {
+	case apierrors.IsUnauthorized(err) || apierrors.IsForbidden(err):
+		logrus.Warningf("Unauthorized to query cluster nodes; continuing with empty node list. Daemonset plugins will be ignored as a result.")
```
johnSchnake, line 86: This is accurate; they will be ignored. Not failed, which is an important distinction. Not sure if there is a way to mark the plugin as failed since it is a daemonset and we dont have the node names.

PR #1567, opened 2021-12-30
New document for sonobuoy query

### LINKED ISSUE #1558: Document `sonobuoy query`
New refactoring actually resulted in a new command that si useful and honestly kind of fun to use. We need its own documentation page for this IMO or maybe a tiny blog post.

NikhilSharmaWe:
Documentation for `sonobuoy query` command
- Fixes #1558

**Release note**:
```
none
```


site/content/docs/main/sonobuoy-query.md
```diff
@@ -0,0 +1,18 @@
+# New `sonobuoy query` command
```
johnSchnake, line 1: Since we are putting this in docs and not a blog, don't refer to this as 'new'. Just make the title `Sonobuoy Query Command`

site/content/docs/main/sonobuoy-query.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Intro
+
+- `sonobuoy query` command runs queries against our cluster and helps in debugging by providing useful information about the cluster which helps us to find bugs and fix them before they can cause any severe harm to the cluster . These queries were run from the aggregator pod (made through sonobuoy run) in the past. But now we can run them locally, even without running sonobuoy or bother launching pods and waiting for processing. We have to just run sonobuoy query.
```
johnSchnake, line 5: - Replace "queries against our cluster" with "queries against your cluster"
- You have an extra space before the period in `severe harm to the cluster .`
- nit: when referring to things that are code/commands use backticks to format them, so "sonobuoy run" should become `sonobuoy run`
- "But now we can run..." -> "You can run them..."
- "without running sonobuoy" should capitalize Sonobuoy
- Last line I would recommend trimming down: "But now we can run them locally, even without running sonobuoy or bother launching pods and waiting for processing. We have to just run sonobuoy query." -> "Now you can run them locally without running Sonobuoy or launching pods. Just run `sonobuoy query` which is much faster and convenient."

site/content/docs/main/sonobuoy-query.md
```diff
@@ -0,0 +1,18 @@
...
+
+  ![sonobuoy query](img/sonobuoy-query-1.png)
+
+- The logs are generated via logrus (https://github.com/sirupsen/logrus), which contains three parts
+  `time, level, msg`.
+- `msg` provides information about the cluster which helps us to find bugs in our clusters and tackle them.
+
+- `level` specifies the level of logs. Default value is `info`, which basically put no filters while returning logs. We can specify it by the help of `--level` flag. It can take values like `{panic, fatal, error, warn, info, debug, trace}`. Each value logs the information according to there functionality, like in case of `panic` we get warnings and info about the cluster of which we should be careful about.
```
johnSchnake, line 13: You can trim this down a bit. Assume the user knows something about log levels. You have good info here but a lot that is unnecessary.

I'd trim it down to state that:
 - the level is the log level
 - We can specify it via the level flag and valid values are {...} and the default is info.

So I'd cut out describing what log levels are or trying to define what `info` means (which I dont think is accurate here).

site/content/docs/main/sonobuoy-query.md
```diff
@@ -0,0 +1,18 @@
...
+
+  ![sonobuoy query --level panic](img/sonobuoy-query-2.png)
```
johnSchnake, line 15: Even trimming down the above description of levels, you can still have the extra screenshot here. Its good to show that we actually can specify level and it changes output in a meaningful way.

site/content/docs/main/sonobuoy-query.md
```diff
@@ -0,0 +1,18 @@
...
+
+- While logging we can also specify **sonobuoy config file** which contains all info about our runs `(sonobuoy run)`, and we get the logs accordingly. We can do that by `--config` flag, we have to provide path to our config in the argument.
```
johnSchnake, line 17: This line is pretty inaccurate since the config has nothing to do with info about runs in this context (since we are not running sonobuoy pods at all).

I would just say:
```
While logging we can also specify **sonobuoy config file** in order to specify query configuration values. This allows you to tweak things like which namespaces to get pod logs for or what API resources to query. In most cases, the default is appropriate (gets logs for kube-system and queries all non-secret resources)
```

site/content/docs/main/sonobuoy-query.md
```diff
@@ -0,0 +1,18 @@
...
+- We can also save our output through `--output` flag, we have to specify path to the directory in which we want to save our data in.
```
johnSchnake, line 18: I'd move this line up higher as it is really important if you've played with the command.

Something like:
```
By default, Sonobuoy uses creates a temporary directory in the current directory to store the query information. You can capture this directory for scripting (its the only value output to stdout) or you can specify your own directory via the `--output` flag.
```

johnSchnake (changes requested):
Requesting a few changes; some rewordings to streamline the way it reads. Thanks for putting this together!

site/content/docs/main/sonobuoy-query.md
NikhilSharmaWe: @johnSchnake Should I keep the bit about `panic` value.
`Each value logs the information according to there functionality, like in case of `panic` we get warnings and info about the cluster of which we should be careful about.`

NikhilSharmaWe:
@johnSchnake thank you so much for making such an effort for reviewing. I think that you have wrote more than myself in this doc :sweat_smile: . I will try to make less mistakes next time.

site/content/docs/main/sonobuoy-query.md
johnSchnake: I would remove that; its not really accurate IMO. Panic means you only get information that was logged at the panic level. It depends on the developers about what they log at that level. We typically only log at fatal and below; panic is a pretty rare case I would think.

johnSchnake:
@NikhilSharmaWe I'm happy to help, I know you're starting out and lots of this is new. The best way to learn is just to keep trying at it. When I was a starting out I wouldn't have known how to navigate random repos, manage git PRs, technical writing, etc.

One of my most valuable pieces of advice I got starting out was just to try really hard to solve problems on your own, but once you really feel blocked and can't imagine how to proceed, always ask for help. Trying to learn is valuable, but at some point you're just wasting your time if you dont know what to do next (and have tried).

johnSchnake:
FYI be sure to click on the button to re-request a review from me when you're ready. Otherwise I'm not sure if you're still fixing it up.

NikhilSharmaWe:
@johnSchnake So, does the document need any more changes or is it fine now ?

site/content/docs/main/sonobuoy-query.md
```diff
@@ -0,0 +1,20 @@
+# `sonobuoy query` command
+
+## Intro
+
+- `sonobuoy query` command runs queries against your cluster and helps in debugging by providing useful information about the cluster which helps us to find bugs and fix them before they can cause any severe harm to the cluster. These queries were run from the aggregator pod (made through `sonobuoy run`) in the past. Now you can run them locally without running Sonobuoy or launching pods. Just run sonobuoy query which is much faster and convenient.
```
johnSchnake, line 5: nit: Dont set this as a bullet point; its just the intro paragraph.

johnSchnake (changes requested):
Nit left for the intro; but the only other thing I can think of before merging is to make this discoverable. If we merge this now, it exists in the docs but nothing links to it at all.

There may be a better place to put it, but right now I'm thinking we can just add something to the FAQ page (still under the main docs folder).

Add another Q/A such as:

```
Q: The information gathered on the cluster is useful for me, but do I have to run a plugin to obtain it?
A: No, you can run the cluster queries via the command `sonobuoy query`. Read more details about it <here>(link to new page).
```

Just match the formatting for the existing Q/As on that FAQ page.

johnSchnake (approved):
💯 Thanks, it's nice to have an extra help on these things.

PR #1576, opened 2022-01-05
Add conformance-lite mode

johnSchnake:
The goal of this mode is to maximize coverage while minimizing
test time. I was able to get 257 tests in 334s.

This is just done from looking at old runs and sorting the tests
by time and avoiding running the slowest ones. We do not specify
the fastest ones because there are so many of them.

- Fixes #

**Release note**:
```
release-note
```


johnSchnake:
Can be tweaked as necessary since this is an unofficial mode. Merging.

PR #1588, opened 2022-01-12
E2e dry run list

johnSchnake:
Adds tests lists (in gzipped format) to support the dry run helper `sonobuoy e2e`.

We will support online and offline querying of this data (offline is baked into the built, online can be updated out of band via github PRs).

You can also pipe data into the command and avoid using our data at all in case you're curious about some other data set (e.g. tests that failed or were skipped on your cluster).

Help page for internal users to repeat these steps when necessary. This should be done on a regular basis, but before Sonobuoy releases and after k8s releases would be the most usual times. Since users can pipe data in, we aren't ever fully blocking someone from using this feature on newer releases of k8s though.


site/content/docs/main/dryRun-listGenerator.md
```diff
@@ -0,0 +1,61 @@
+# Plugins used to help create test lists
```
johnSchnake, line 1: Steps for how I created this. Just an internal doc, not linked into the main readme or anything.

johnSchnake:
- I kind of hate having a 300k line diff but it is what it is I guess?
 - I think the files should have a suffix so they are easier to filter; but that just means that we'll also need to adjust the code in the feature to assume a file suffix when forming the URLs/paths

johnSchnake:
Q: Is this too many versions? Should we go even fewer? It is possible for test names to change slightly (usually a typo fixed or a tag added/removed) between point releases even if it is rare. Otherwise we would consider just using major.minor

johnSchnake:
Confirmed that this doubles the size of the binary since it is baked into it. I do not like that idea.

johnSchnake:
One option would be to stop putting it in the binary at all and instead the 'offline' mode could read from the SONOBUOY_DIR where we also put the plugins we can install.

That is another way to leverage that new location and also provides a way for users to modify the lists if they have changes or want to tag their own experimental version.

We'd have to give a way to download the data there (e.g. `sonobuoy e2e download-lists`) but then it would work fine.

We could also just tar up the test list for each version. Instead of being 950K its only 50K per set of tests then. Going to try that since it would make the LOC for the change smaller; would speed up downloads, and lowers binary size.

PR #1590, opened 2022-01-13
Adjust progress messages to allow for appending data

johnSchnake:
In the parallel ginkgo case, each thread doesn't know the
totals of how many tests will be run or have been run.

As a result we need to be able to handle partial updates such
as just saying one more test passed/failed.

This adds a few fields to the data type and helper functions/tests
to ensure that the appending works as intended.

Upstream change to utilze this will be next.

** Release Note **
```
Plugins can now report progress updates in more piecemeal fashion and the aggregator will combine the results. This fixes a bug where the e2e plugin couldn't report progress when in parallel mode. The fix will be in Sonobuoy, but the upstream fix will get put into releases going forward.
```


johnSchnake:
Testing this works as expected with a custom patched e2e conformance image. If I can get them working together then I'll merge this and make the upstream PR

PR #1609, opened 2022-02-04
Add a delay before consuming done file for postprocessing window

### LINKED ISSUE #1530: Customize "done" file name.
Need to wire the sonobuoy config to allow for other done file names to simplify having plugins, workers, and post-processors work together.

For instance, I am working on a basic post-processor POC and wanted to test it on the e2e output (most common plugin). However, the e2e plugin will always write the `done` file, triggering the worker to submit the results. This doesn't give the post-processor time to do anything. We could change the upstream e2e plugin to have an option and not write the done file, but I think this logic is smoother:
 - let the e2e plugin (or others) do whatever they want
 - The caller is the point of integration of all of these parts so they can integrate things intelligently.
 - They can tell tell the worker/post-processor to use a different done file; in this case the e2e plugin writing the done file can even trigger the postprocessor.

Alternatively we could just add another endpoint which the worker exposes that would double as the done file. Then we could just do a POST with the results location.

johnSchnake:
In order to facilitate interception and modification of results
by sidecars, wait some amount of time before consuming the done
file and sending results.

This allows sidecars to remove the done file, modify them as necessary,
and then rewrite the done file when appropriate.

- Fixes #1530

I used a modified e2e plugin to test this; notice the sidecar I added. The net result will be that results never get submitted to the aggregator. When we build/publish post-processors we can have better integration tests that this functionality works.

**Release note**:
```
The Sonobuoy worker (sidecar to the plugin images) will now wait 5s before submitting results. If the done file is removed during that time, it will begin polling again for it. This allows user-defined sidecars to intercept the results and add post-processing capabilities.
```

### DIFF
--- a/pkg/worker/worker.go
+++ b/pkg/worker/worker.go
@@ -42,18 +42,39 @@ const (
 	SonobuoyNSEnvKey              = "SONOBUOY_NS"
 	SonobuoyPluginPodEnvKey       = "SONOBUOY_PLUGIN_POD"
 	SonobuoyWorkerContainerEnvKey = "SONOBUOY_WORKER_CONTAINER"
+
+	// defaultWaitFileConsumptionDelay provides a minimum time window where sidecars may intercept results and remove
+	// the done file before Sonobuoy decides to process it.
+	defaultWaitFileConsumptionDelay = 5 * time.Second
+
+	// SonobuoyDoneFileDelayKey is the env key specifying where this value may be overwritten. Adding this in as a mitigation for problems
+	// caused by adding the delay. I anticipate we will not have a problem though since the default delay is reasonably short.
+	SonobuoyDoneFileDelayKey = "SONOBUOY_DONE_FILE_DELAY"
 )
 
 var (
 	// apiClient is used to check the status of this pod and other containers in it. If it is
 	// nil, those checks will be skipped.
 	apiClient kubernetes.Interface
+
+	// waitFileConsumptionDelay provides a minimum time window where sidecars may intercept results and remove
+	// the done file before Sonobuoy decides to process it.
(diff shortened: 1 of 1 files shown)


pkg/worker/worker.go
```diff
@@ -42,19 +42,39 @@ const (
...
+	waitFileConsumptionDelay time.Duration
 )
 
 func init() {
 	err := mime.AddExtensionType(".gz", "application/gzip")
 	if err != nil {
 		logrus.Error(err)
 	}
+
+	if v := os.Getenv(SonobuoyDoneFileDelayKey); len(v) > 0 {
```
johnSchnake, line 71: But how would someone actually modify this? In theory it could be overwritten, but we don't expose the worker env to the users and dont have any flags for that currently.

johnSchnake: I think that this just requires a small extension to the way we handle the env vars. Whenever we make the sonobuoy worker, perhaps we just copy over the env from the plugin itself. That way its easy for the user to set those values and it allows for per-plugin values rather than some global 'worker env' we set.

johnSchnake: We currently dont have the means to do that (the createworkercontainerdefintion method doesn't have access to that nor does the plugin interface).

So the change is _slightly_ non-trivial so I'm going to put that in a separate PR. We just have to have a separate call to add env values to the worker image.

johnSchnake:
The only failure is a HTTP 500 from something related to the windows image building. Merging.
```
error: failed to solve: failed to compute cache key: failed to copy: httpReadSeeker: failed open: unexpected status code https://mcr.microsoft.com/v2/windows/nanoserver/blobs/sha256:fe8c39bf0334000f0db1aed1c1eddb880d2af07f765f8ff0f91ace66c655cab9: 500 Internal Server Error
```

PR #1612, opened 2022-02-04
Issue1595: deal with plugins with unknown total tests count

### LINKED ISSUE #1595: Status should handle unknown tests counts better
If a test suite doesn't report the total number of tests (<=0 for total tests) then we shouldn't show "7/0 (1 failure)", it should show something that makes it clear that 7 tests have completed of an unknown number. Even if we just use `?` instead of the denominator it seems it would be better.

eturra:
Issue1595
Some plugins do not declare the total number of tests, so the progress update is confusing

- Fixes #1595

### DIFF
--- a/cmd/sonobuoy/app/status.go
+++ b/cmd/sonobuoy/app/status.go
@@ -31,7 +31,6 @@ import (
 
 	"github.com/vmware-tanzu/sonobuoy/pkg/client"
 	"github.com/vmware-tanzu/sonobuoy/pkg/errlog"
-	"github.com/vmware-tanzu/sonobuoy/pkg/plugin"
 	"github.com/vmware-tanzu/sonobuoy/pkg/plugin/aggregation"
 )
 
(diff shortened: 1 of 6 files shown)


johnSchnake (changes requested):
A few tests are failing:
```
--- FAIL: TestPrintStatus (0.00s)
    --- FAIL: TestPrintStatus/StatusSummary (0.00s)
    --- FAIL: TestPrintStatus/StatusShowAll (0.00s)
```

Those use the goldenfile pattern so if you're happy with how the output looks run:
```
go test github.com/vmware-tanzu/sonobuoy/cmd/... -update
```

And check the changes with `git diff` and take it from there. You can target the pkg more carefully if you want but that will get them (and just err on some other pkgs I'm sure).

johnSchnake:
You'll need to squash the commits down too. You can ping me if you need help with that.

Then the only github magic is that instead of just mentioning "Issue 1595"; if you put `Fixes #1595` it will automatically link/close the issue upon merging. It can be in the PR comment, it doesnt have to be in the commit message.

johnSchnake (approved):
Looks great! I confirmed on my cluster with both quick mode (seeing remaining) and conformance-lite (where total is -1 and so it is not printed).

Issue #1589, opened 2022-01-13
Sonobuoy results could show additional information


johnSchnake:
`sonobuoy results` currently shows plugins and their pass/fail/skip numbers.

However, when given a tarball to inspect, you very often may want to know more than just that. Sonobuoy even knows that since we add all those queries about the cluster into the process by default.

The question is, how can we automate more of the debugging steps I would take (as a knowledgeable Sonobuoy user) that might make debugging easier for ALL users who dont know where to look for logs/etc within the tarball of results.

For instance, in addition to plugin details, maybe we add another section along these lines:

```
Run details:
Node health: 5/5 (100%)
API Server version: v1.20.1
Errors detected in files:
podlogs/kube-system/kube-scheduler-kind-control-plane/logs/kube-scheduler.txt:17
podlogs/kube-system/kube-apiserver-kind-control-plane/logs/kube-apiserver.txt:2471
podlogs/kube-system/kube-controller-manager-kind-control-plane/logs/kube-controller-manager.txt:2043

Pods in an error state:
foo/bar: ErrorImgPull
...
```

It would be useful to at least know that nodes are up, API server responding, automatically grep for common error formats (e.g. klog), just to point people in the right direction for debugging.

johnSchnake:
Just brainstorming: if we expect to have users rely on this output for automated behavior in scripts, we could even consider a `--json` flag or something. Then we'd need to create a datatype for the information and just call different print routines depending on the mode.

PR #1615, opened 2022-02-08
Issue1589

eturra:
Add a summary of the health of nodes and pods to the results command

- Fixes the first part of #1589

There's another chunk ready for review, adding the log analysis part of issue 1589, I'll push it after this review.


johnSchnake (commented):
I'll have to comb through this a bit more but I built/tested locally and it works great:

```
Plugin: sonobuoy
Run details:
API Server version: v1.20.0
Node health: 3/3 (100%)
Nodes details:
kind-control-plane Ready:True: KubeletReady: kubelet is posting ready status
kind-worker Ready:True: KubeletReady: kubelet is posting ready status
kind-worker2 Ready:True: KubeletReady: kubelet is posting ready status

Pods health: 21/24 (87%)
Failed pods details:
Namespace:  secrets-627
secrets-627/pod-secrets-12f1832b-8a06-4c5a-a4d5-4f394d1c9d12 Ready:False: ContainersNotReady: containers with unready status: [dels-volume-test upds-volume-test creates-volume-test]

Namespace:  var-expansion-5477
var-expansion-5477/var-expansion-c62beb8c-6e93-498b-b362-8a7449cf4ead Ready:False: ContainersNotReady: containers with unready status: [dapi-container]

Namespace:  init-container-4990
init-container-4990/pod-init-79275f3b-ced0-4503-899d-b72fb8d70219 Ready:False: ContainersNotInitialized: containers with incomplete status: [init1 init2]
```

Ideas to improve the output slightly:
 - don't print the node details if there isn't a problem, especially in this summary mode. If its not important we should just hide it from the user. The detailed/dump modes can print all the finer details
 - I wouldn't write it as "Plugin: sonobuoy" and also have the "Run Details" line. I prefer the heading for "Run details" and think referring to it as a plugin will make people confused.

More detailed review to come.

johnSchnake:
Just tried detailed/dump modes too and thought it looked great. Looking at the json it does look like maybe those healthy nodes shouldn't have ended up in the list at all? For the pods you only kept details on the non-ready ones, but despite being healthy/ready the nodes are listed in the details and in the summary.

eturra:
I have incorporated your suggestions, let me know what it looks like now.

The idea of printing the "Plugin: sonobuoy" line was just to mimic the behaviour of a plugin, e.g. if a user sees the summary and wants to print just that, this would help the user figure out the `-p sonobuoy` option, but I guess it's a matter of documenting it, so I removed it.
I have decided also to collect the information about all pods and namespaces, regardless of their health, and just avoid printing them in the summary, and in case the information is useful, the "detailed" and "dump" modes would still have the full list for further processing.

Lastly, I have included the log scanning part too, with dummy regex examples, so you can review that par too.
I haven't squashed the commits so we can back the log scanning functionality out if necessary, 
but if you're happy with the rest, let me know, and I will squash before you accept the PR.

cmd/sonobuoy/app/results.go
```diff
@@ -143,6 +151,15 @@ func result(input resultsInput) error {
 			lastErr = err
 		}
 
+		//bypass if this plugin is called clusterHealthSummaryPluginName
+		if input.plugin == clusterHealthSummaryPluginName {
```
johnSchnake, line 155: I wouldn't put this here; there is a validation section where we make sure we don't have duplicates or reserved names being used. Its in `gen.go` (the client one, not the cmd one). `checkPluginNames` is the function name.

Technically someone could do sonobuoy gen, save the output and modify it to avoid this check then run it. I dont think its worth the time to really ensure people dont shoot themselves in the foot manually though. If we were that concerned we could just run that check here too rather than distributing this logic in multiple places.

cmd/sonobuoy/app/results.go
```diff
@@ -254,6 +327,75 @@ func printResultsDetails(treePath []string, o *results.Item, input resultsInput)
...
+	fmt.Println("Run Details:")
+	fmt.Printf("API Server version: %s\n", summary.APIVersion)
+
+	fmt.Printf("Node health: %d/%d", summary.NodeHealth.Healthy, summary.NodeHealth.Total)
+	//Print the percentage only if Total is not 0 to avoid division by zero errors
+	if summary.NodeHealth.Total != 0 {
+		fmt.Printf(" (%d%%)", 100 * summary.NodeHealth.Healthy / summary.NodeHealth.Total)
+	}
+	fmt.Println()
+	//Details of the failed pods. Checking the slice lenght to avoid trusting the Total
```
johnSchnake, line 342: typo: length

cmd/sonobuoy/app/results.go
```diff
@@ -254,6 +327,75 @@ func printResultsDetails(treePath []string, o *results.Item, input resultsInput)
...
+		fmt.Printf("Pods health: %d/%d", summary.PodHealth.Healthy, summary.PodHealth.Total)
+		//Print the percentage only if Total is not 0 to avoid division by zero errors
+		if summary.PodHealth.Total != 0 {
+			fmt.Printf(" (%d%%)", 100 * summary.PodHealth.Healthy / summary.PodHealth.Total)
+		}
+		fmt.Println()
+		if summary.PodHealth.Healthy < summary.PodHealth.Total {
+			pods := make(map[string][]*discovery.HealthInfoDetails)
+			fmt.Println("Details for failed pods:")
+			//First, scan all the pods and extract the not ready ones, and group them by namespace
```
johnSchnake, line 365: We also need to ensure that output is consistent, so we want to sort by namespace/name.

I also think that having the line with `Namespace: %v` is unnecessary sine we're writing it with the pod name.

If it were me I'd sort the slice once by "namespace/name" then iterate over it one time and print the info for the ones that aren't healthy.

That would remove the need for a loop and ensure that the output is perfectly consistent (which is important if we have tests that check the output).

cmd/sonobuoy/app/results.go
```diff
@@ -254,6 +327,75 @@ func printResultsDetails(treePath []string, o *results.Item, input resultsInput)
...
+				fmt.Println("Namespace: ", namespace)
+				for _, pod := range pods[namespace] {
+					fmt.Printf("%s/%s Ready:%s: %s: %s\n", pod.Namespace, pod.Name, pod.Ready, pod.Reason, pod.Message)
+				}
+				fmt.Println()
+			}
+		}
+	}
+
+	if len(summary.ErrorInfo) > 0 {
```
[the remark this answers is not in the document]
eturra: Yes, review the log output too, I merged the commit because its changes are overlapping a bit for the ClusterSummary struct and the printClusterHealthResultsSummary, and the summary test, but I kept it a separate commit, so it's easy to back it out if necessary.

pkg/discovery/testdata/podlogs/kube-system/coredns-74ff55c5b-cpfqs/logs/coredns.txt
```diff
@@ -0,0 +1,4 @@
+.:53
```
johnSchnake, line 1: For all these testdata files, please nest them down one extra layer to avoid polluting that directory itself. That way we have
```
testdata
|- healthTestsHappy
|- healthTestsWithErrors
|- healthTestsMissingFiles
|- otherTests
...
```

pkg/discovery/summary_test.go
```diff
@@ -0,0 +1,43 @@
...
..."name":"sonobuoy-systemd-logs-daemon-set-7a10f4effeb44d32-7n2cj","healthy":true,"ready":"Ready","namespace":"sonobuoy"},{"name":"sonobuoy-systemd-logs-daemon-set-7a10f4effeb44d32-jrn9s","healthy":true,"ready":"Ready","namespace":"sonobuoy"},{"name":"sonobuoy-systemd-logs-daemon-set-7a10f4effeb44d32-zqfjc","healthy":true,"ready":"Ready","namespace":"sonobuoy"}]},"api_version":"v1.20.0","error_summary":{"Errors":{"podlogs/kube-system/kube-apiserver-kind-control-plane/logs/kube-apiserver.txt":4,"podlogs/kube-system/kube-controller-manager-kind-control-plane/logs/kube-controller-manager.txt":44,"podlogs/kube-system/kube-scheduler-kind-control-plane/logs/kube-scheduler.txt":43},"Warnings":{"podlogs/kube-system/kube-apiserver-kind-control-plane/logs/kube-apiserver.txt":11,"podlogs/kube-system/kube-controller-manager-kind-control-plane/logs/kube-controller-manager.txt":13,"podlogs/kube-system/kube-proxy-8jj5r/logs/kube-proxy.txt":1,"podlogs/kube-system/kube-proxy-g79v6/logs/kube-proxy.txt":1,"podlogs/kube-system/kube-proxy-lrtdp/logs/kube-proxy.txt":1,"podlogs/kube-system/kube-scheduler-kind-control-plane/logs/kube-scheduler.txt":4,"podlogs/sonobuoy/sonobuoy/logs/kube-sonobuoy.txt":2}}}`)
```
johnSchnake, line 28: When you have this much data you really want to try and do something besides hardcode it like this. Its hard to consume (you can't easily cat the data or pipe it on the CLI), it makes your IDE a bit odd since its such a long line, and when/if changes are made it can be hard to update.

So thats what the "golden file" tests are for. That way the long, unwieldy data goes in its own file and if we want to update it we just run the test with the `-update` flag.

It can be odd to implement the first time so I'm happy to help if you're not sure how to apply that pattern (we use that pattern elsewhere so you can look at how its done though and dont have to start from scratch).

johnSchnake (changes requested):
Amazing job. Its been a while since I've had to review a big new feature PR like this too, I appreciate all the time and effort you put in.

I had a bunch of comments but they are mostly pretty mild comments regarding some code organization/naming/testing.

I still have to test it a little bit more (wanted to run with some errors in logs and such) but I think its already close to ready to merge after the fixups.

cmd/sonobuoy/app/results.go
```diff
@@ -157,6 +174,62 @@ func result(input resultsInput) error {
...
+		})
+		return err
+	}
+
+
+	// For summary and dump views, get the item as an object to iterate over.
+	clusterHealthSummary := discovery.ClusterSummary{}
+
+	err = r.WalkFiles(func(path string, info os.FileInfo, err error) error {
+		return results.ExtractFileIntoStruct(discovery.ClusterHealthFilePath(), path, info, &clusterHealthSummary)
```
johnSchnake, line 208: 2 things about this section:
 - the walk files is unnecessary; `r` has a a `.FileReader(<file>)` method that will give you the reader you can copy from
 - Can we move the `discovery.ClusterHealthFilePath()` to be defined on the reader? When looking at the previous point, I found there are similar methods named for important files in the tarball

pkg/discovery/summary_test.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+const (
+	//Root directory where all the file sin the tarball are located
+	//Simulates the directory where all the files resulting from discovery end up before being compressed into the tarball
+	//For the test, this directory is the location where the test files are located
+	tarballRootDir = "testdata"
+
+	//Name of the golden file that stores the expected result
+	SummaryGoldenFileName = "summary_test.golden"
```
johnSchnake, line 35: These constants are just test specific; just define them in the test. Thats just the norm for tests otherwise you explode with constants. I know its kind of the opposite of the other case in a normal method but in normal code those strings/constants are pretty rare but every single test case usually has 1 or more constants.

pkg/discovery/summary_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"flag"
+        "path/filepath"
+	"io/ioutil"
+)
+
+const (
+	//Root directory where all the file sin the tarball are located
+	//Simulates the directory where all the files resulting from discovery end up before being compressed into the tarball
+	//For the test, this directory is the location where the test files are located
+	tarballRootDir = "testdata"
```
johnSchnake, line 32: This is why I wanted you to nest the files down an extra layer. Otherwise the next test that gets added has to 
 - realize they need to nest each case
 - go back and nest these files into a folder so it doesnt pollute their test

I'd rather the next time we add tests like this we can just copy the pattern we used here and specify the tarball root in each test case (since it will be test specific)

pkg/discovery/summary.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	defer file.Close()
+	decoder := json.NewDecoder(file)
+	err = decoder.Decode(object)
+	//no need to check for err as we are returning it anyway
+	return err
+}
+
+//Read the server version from the working directory and return it as a string
+func ReadVersion(tarballRootDir string) (string, error) {
```
johnSchnake, line 85: [TODO, we can just make it in a clean-up PR]
It seems really odd that we have to re-impelemnt this (not your fault). We have these sort of methods already defined in `reader.go` but its always been targeted at a tarball and we read it without having to unzip it.

Seems like it would be really simple to update the methods to take a directory or tarball so we can centralize the file structure info/methods.

pkg/discovery/logsummary_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+        "encoding/json"
+	"bytes"
+        "path/filepath"
+	"io/ioutil"
+)
+
+const (
+	//Name of the golden file that stores the expected result
+	LogSummaryGoldenFileName = "logsummary_test.golden"
```
johnSchnake, line 29: Same as other test comment; keep those constants just in the test itself. Too many test constants to define in both places.

cmd/sonobuoy/app/results.go
johnSchnake: Oops; wrong 'Walk' section. I meant the one above this where you find the file by name yourself.

pkg/discovery/summary.go
```diff
@@ -0,0 +1,238 @@
...
+
+	//Filename of the cluster health, relative to MetaLocation
+	ClusterHealthFile = "clusterhealth.json"
+)
+
+// LoadJsonIntoStruct reads a json file into an object.
+// fileName needs to be the name of a file that can be opened
+// error will be returned if opening the file fails,
+// or if decoding its content as Json into the supplied "object" fails
+func LoadJsonIntoStruct(fileName string, object interface{}) (err error) {
```
johnSchnake, line 72: Is this the method you thought shouldn't be needed? Because you had an issue with an existing method? Could you refresh my memory on that?

cmd/sonobuoy/app/results.go
eturra: Regarding `ClusterHealthFilePath`:
in the reader I see the other methods there are `Reader`'s methods, should `ClusterHealthFilePath` be a `Reader`'s method too ?
I am asking because then I'd need to instantiate a Reader in `pkg/discovery/summary.go` just to get the file name:
is it Ok if I just move it to the `reader.go` file, together with the `ClusterHealthFile` const, but leave it as a "standalone" func  instead ?

pkg/discovery/summary.go
eturra: What I was doing was something like this: (this is a dummy example, I can go through the commits to see if i can find it if necessary)
```go
        file, _ := os.Open("test.txt")
        info, _ := file.Stat()
        _, ok := info.Sys().(io.Reader)
        if !ok {
                fmt.Println("Not a reader")
        }
```
This is just a simplification of the issue I was hitting, basically I was trying to open a file I wanted to process, and then I was passing it to `ExtractIntoStruct`, for example, and it was failing this test:

```go
                reader, ok := info.Sys().(io.Reader)
                if !ok {
                }
```
Maybe it's my fault, and there's a better way to get a `os.FileInfo` object that has a `Sys()` that is a reader, I just couldn't find out, so I took the shortcut of creating `LoadJsonIntoStruct`

I'd say, as you said in a comment above, if we can make the `Reader` transparently work out of a tgz or an uncompressed directory, this issue would go away, and there would be no need to have a separate `LoadJsonIntoStruct`

johnSchnake: OK, lets just ignore the issue for the moment and consider it an issue for cleanup when that gets done then. 👍

cmd/sonobuoy/app/results.go
```diff
@@ -254,6 +317,86 @@ func printResultsDetails(treePath []string, o *results.Item, input resultsInput)
...
+			}
+		}
+	}
+
+	if len(summary.ErrorInfo) > 0 {
+		fmt.Println("Errors detected in files:")
+		for errorType := range summary.ErrorInfo {
+			fmt.Printf("%s:\n", errorType)
+			for fileName := range summary.ErrorInfo[errorType] {
+				fmt.Printf("%s:%d\n", fileName, summary.ErrorInfo[errorType][fileName])
```
johnSchnake, line 392: I didnt realize it until i was trying to verify the output, but we shouldnt use `filename:errorCount`; it looks exactly like the format used for `filename:line`. I was thinking there was a bug identifying errors but it was giving me an error count and not the line number of each detected error.

I think it would be most clear to say `filename (%d found)`

johnSchnake: I think we can also put an issue to touch up the functionality slightly so we trace out what patterns did match/where. I'm looking at some of the output and having trouble identifying the warnings its saying it found. That doesn't have to be this PR though.

pkg/discovery/logsummary.go
```diff
@@ -0,0 +1,172 @@
...
+			scanner := bufio.NewScanner(file)
+			for scanner.Scan() {
+				line := scanner.Text()
+				for _, patternName := range patternsForFile {
+					for _, pattern := range patterns[patternName].matchPatterns {
+						if pattern.MatchString(line) {
+							if _, ok := logSummary[patternName]; !ok {
+								logSummary[patternName] = make(LogHitCounter)
+							}
+							logSummary[patternName][relFilePath]++
```
johnSchnake, line 153: If we match a pattern on a line, lets not keep increasing the count if it matches more patterns. I think we're double counting a fair number. Totally fine if a line hits as both a warning and an error but for instance:

```
W0211 17:21:27.983694       1 warnings.go:70] v1 ComponentStatus is deprecated in v1.19+
```
Is hitting 2x; once for W0211 and once for `warnings.go` hitting the `[wW]arn` regexp.

johnSchnake:
Double counting warnings/errors and a formatting issue then we can squash/merge this I think. We'll continue to improve/refine it over time but this is great and exceeds my expectations for what I thought we'd get on a first pass. 💯

johnSchnake:
When you squash, be sure to rebase onto upstream/main too

cmd/sonobuoy/app/results.go
eturra: Indeed, the first time I saw the description in the issue itself I also though these were supposed to be line numbers.
I've changed it to look like the output of `grep -l <regex> | sort | uniq -c | sort -n`, see if it looks good, otherwise I'll change it to what you suggested.
It would look something like this:
```
Errors:
22 podlogs/kube-system/kube-controller-manager-kind-control-plane/logs/kube-controller-manager.txt
22 podlogs/kube-system/kube-scheduler-kind-control-plane/logs/kube-scheduler.txt
 2 podlogs/kube-system/kube-apiserver-kind-control-plane/logs/kube-apiserver.txt
```

cmd/sonobuoy/app/results.go
johnSchnake: This was an earlier comment; in light of the other discussion about how that reader stuff should work on a directory/tarball we can just leave this as-is for now and include it in the cleanup work

johnSchnake:
I think this looks good; if you're happy lets go ahead and merge it. 

I have a local branch to try and get the reader type to handle working with the directory itself too; so if we get this in I can iterate on that a little easier.

rebase/Squash/sign-off 🚀

johnSchnake (approved):

PR #1631, opened 2022-02-18
Modify reader code to work on unzipped results as well

### LINKED ISSUE #1617: Make the tarball reader work from a directory as well
It seems really odd that we have to re-impelemnt this. We have these sort of methods already defined in `reader.go` but its always been targeted at a tarball and we read it without having to unzip it.

Seems like it would be really simple to update the methods to take a directory or tarball so we can centralize the file structure info/methods.

_Originally posted by @johnSchnake in https://github.com/vmware-tanzu/sonobuoy/pull/1615#discussion_r804129192_

johnSchnake:
The code for the results reader was made to read the tarball without
opening it but it should also work on unzipped results. This allows
more reuse of the methods, a centralized place to put results structure
information, and will allow us to expand `results` to process an unzipped
directory as well.

- Fixes #1617

### DIFF
--- a/cmd/sonobuoy/app/results.go
+++ b/cmd/sonobuoy/app/results.go
@@ -32,7 +32,7 @@ import (
 	"github.com/vmware-tanzu/sonobuoy/pkg/client/results"
 	"github.com/vmware-tanzu/sonobuoy/pkg/discovery"
 	"github.com/vmware-tanzu/sonobuoy/pkg/errlog"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 )
 
 const (
@@ -100,6 +100,13 @@ func NewCmdResults() *cobra.Command {
 // getReader returns a *results.Reader along with a cleanup function to close the
 // underlying readers. The cleanup function is guaranteed to never be nil.
 func getReader(filepath string) (*results.Reader, func(), error) {
+	fi, err := os.Stat(filepath)
+	if err != nil {
+		return nil, func() {}, err
+	}
+	if fi.IsDir() {
+		return results.NewReaderFromDir(filepath), func() {}, nil
+	}
 	f, err := os.Open(filepath)
 	if err != nil {
 		return nil, func() {}, errors.Wrapf(err, "could not open sonobuoy archive: %v", filepath)
@@ -141,7 +148,6 @@ func result(input resultsInput) error {
 	var lastErr error
 	for i, plugin := range plugins {
 		input.plugin = plugin
-
 
 		// Load file with a new reader since we can't assume this reader has rewind
 		// capabilities.
@@ -330,7 +336,7 @@ func sortErrors(errorSummary discovery.LogSummary) map[string][]string {
 		}
 		//Sort in descending order,
 		//And use the values in hitCounter for the sorting
-		isMore := func(i,j int) bool {
+		isMore := func(i, j int) bool {
 			valueI := hitCounter[sortedFileNamesList[i]]
 			valueJ := hitCounter[sortedFileNamesList[j]]
 			return valueI > valueJ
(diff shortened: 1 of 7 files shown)


cmd/sonobuoy/app/results.go
```diff
@@ -100,6 +100,13 @@ func NewCmdResults() *cobra.Command {
 // getReader returns a *results.Reader along with a cleanup function to close the
 // underlying readers. The cleanup function is guaranteed to never be nil.
 func getReader(filepath string) (*results.Reader, func(), error) {
+	fi, err := os.Stat(filepath)
+	if err != nil {
+		return nil, func() {}, err
+	}
+	if fi.IsDir() {
+		return results.NewReaderFromDir(filepath), func() {}, nil
+	}
```
johnSchnake, line 109: This is almost the only code needed to make `sonobuoy results` work on a directory. Easy win if we already wanted to do the other bits.

cmd/sonobuoy/app/results.go
```diff
@@ -341,22 +347,21 @@ func sortErrors(errorSummary discovery.LogSummary) map[string][]string {
...
 }
 
-
-// filterAndSortHealthInfoDetails takes a copy of a slice of HealthInfoDetails, 
+// filterAndSortHealthInfoDetails takes a copy of a slice of HealthInfoDetails,
 // discards the ones that are healthy,
-// then sorts the remaining entries, 
+// then sorts the remaining entries,
 // and finally sorts them by namespace and name
-func filterAndSortHealthInfoDetails(details []discovery.HealthInfoDetails) ([]discovery.HealthInfoDetails) {
```
johnSchnake, line 349: This is just `gofmt -s` changes. Automatic in my IDE.

pkg/client/results/reader.go
```diff
@@ -62,11 +65,17 @@ const (
 	defaultServerVersionFile  = "serverversion.json"
 	defaultServerGroupsFile   = "servergroups.json"
 
+	// CoreNodesFile is the filename of the core nodes json output, relative to nonNamespacedResourcesDir
+	CoreNodesFile = "core_v1_nodes.json"
+
+	// CorePodsFile is the filename of the core pod json output, relative to namespacedResourcesDir
+	CorePodsFile = "core_v1_pods.json"
+
```
johnSchnake, line 73: Just moved these from discovery

pkg/client/results/reader.go
```diff
@@ -192,10 +229,40 @@ func (t *tarFileInfo) Sys() interface{} {
 	return t.Reader
 }
 
+func cd(path string) (func(), error) {
```
johnSchnake, line 232: This is the secret sauce that was required to get this to work. The issue that my first version had (and probably yours) was that we specify the root directory and whenever we `Walk` the files we need the paths to all be relative to it; that messes things up since when "in" the tarball we dont have that issue.

So the easiest approach was to simply programmatically change our directory just while doing the walk functions. It doesn't mess with the shell at all, just what Go sees as the pwd.

pkg/client/results/reader.go
```diff
@@ -402,7 +481,26 @@ func (r *Reader) FileReader(filename string) (io.Reader, error) {
...
+
+// fileInfoToReader takes the given FileInfo object and tries to return a reader
+// for the data. In the case of normal FileInfo objects (e.g. from os.Stat())
+// you need to provide the full path to the file so it can be opened since the
+// FileInfo object only contains the name but not the directory.
+func fileInfoToReader(info os.FileInfo, path string) (io.Reader, error) {
+	switch v := info.Sys().(type) {
+	case io.Reader:
+		return info.Sys().(io.Reader), nil
+	case syscall.Stat_t, *syscall.Stat_t:
```
johnSchnake, line 497: It is frustrating; the docs says `.Sys()` should give you the backing data, but its just not the case for general FileInfo types. In this case you need to just call Open manually.

pkg/discovery/summary.go
```diff
@@ -149,25 +106,26 @@ func ReadPodHealth(tarballRootDir string) (HealthInfo, error) {
...
-	err := LoadJsonIntoStruct(fileName, &nodes)
-
+	r := results.NewReaderFromDir(tarballRootDir)
+	fileName := path.Join(ClusterResourceLocation, results.CoreNodesFile)
+	err := r.WalkFiles(func(path string, info os.FileInfo, err error) error {
+		return results.ExtractFileIntoStruct(fileName, path, info, &nodes)
+	})
 	if err != nil {
-		logrus.Errorf("Failed to read health sumamry: failed to read the node list from '%s': %s", fileName, err)
+		logrus.Errorf("Failed to read health summary: failed to read the node list from '%s': %s", fileName, err)
```
johnSchnake, line 128: Fixed typo

PR #1638, opened 2022-03-08
Issue1578

### LINKED ISSUE #1578: Better debug info when aggregator fails to start
A very common issue goes like this:

Q: Whats taking Sonobuoy so long to finish? It says it is waiting to get its initial status but nothing else happens. Is it hanging? Is this normal?
A: Oh, if that happens you need to check whats going on using `kubectl`

Preferred solution:
 - While we are waiting for the aggregator pod to start, we should monitor/display things like pod status. Most often the pod is in an imagePullBackoff loop, in some cases there are failed volume mounts (common if you are on a windows node but tried to use a securityContext still), etc. We shouldn't make the users have to dig into this every time.

eturra:
Produce some text describing the status of the aggregator pod is being deployed,
so that it can be monitored without additional `kubectl` commands

- Fixes #1578 

At the moment the change is spread across 3 commits that will be squashed when ready.


johnSchnake:
Sorry I hadn't reviewed this yet; taking look today

pkg/client/testdata/PrintPodStatus.json
```diff
@@ -0,0 +1,2818 @@
+{
```
johnSchnake, line 1: I'd really not prefer to have a 2k line test file here if we dont need it.

Lets just scope this to the ones that show our common use cases/errors.

pkg/client/run.go
```diff
@@ -302,3 +307,18 @@ func humanReadableStatus(str string) string {
 		return fmt.Sprintf("Sonobuoy is in unknown state %q. Please report a bug at github.com/vmware-tanzu/sonobuoy", str)
 	}
 }
+
+
+func getPodStatus(pod corev1.Pod) string {
```
johnSchnake, line 312: This isn't yet sufficient to what we need to show the users.

The most common use case is an ErrImagePull state and that isn't reflected here since that is on the container status:

```
    containerStatuses:
    - image: schnake/sonobuoy:987
      imageID: ""
      lastState: {}
      name: kube-sonobuoy
      ready: false
      restartCount: 0
      started: false
      state:
        waiting:
          message: Back-off pulling image "schnake/sonobuoy:987"
          reason: ImagePullBackOff
```

To repro this use case just run:
```
sonobuoy run --wait --sonobuoy-image=schnake/sonobuoy:987
```

And all I see is:
```
11:19:44 Waiting for the aggregator status to become Running. Currently the status is sonobuoy/sonobuoy: Pending: Ready: ContainersNotReady, containers with unready status: [kube-sonobuoy]
...
...
...
```

johnSchnake: In the above output I'll also say that you can see how its confusing because the "condition" is "Ready" but its false. As printed, it looks like youre saying it IS ready.

When printing the condition you should just tweak the format string so it reads like "status is pending: [Ready=false]: ContainersNotReady...

pkg/client/run_test.go
```diff
@@ -68,3 +72,42 @@ func TestRunInvalidConfig(t *testing.T) {
...
+		"kube-system/etcd-kind-control-plane: Running",
+		"kube-system/kindnet-hknk7: Running",
+		"kube-system/kindnet-qhzxn: Running",
+		"kube-system/kindnet-x9fsq: Running",
+		"kube-system/kube-apiserver-kind-control-plane: Running",
+		"kube-system/kube-controller-manager-kind-control-plane: Running",
+		"kube-system/kube-proxy-8jj5r: Running",
+		"kube-system/kube-proxy-g79v6: Running",
+		"kube-system/kube-proxy-lrtdp: Running",
+		"kube-system/kube-scheduler-kind-control-plane: Running",
```
johnSchnake, line 89: As my other comment says, most of this test case doesn't add any value. We dont need to check the running state 10 times.

johnSchnake (changes requested):

pkg/client/run.go
eturra: I've pushed a 4th draft, let me know what you think of it.
I've added also details for the containers that are not ready, so it would show `ImagePullBackOff`, for example.
To do that, I decided to add newlines (to separate the pod details from the container details), let me know what you think of that too (can we have newlines?)
It should look something like this (from the test output)

```
Status: Pending, Reason: ContainersNotReady, containers with unready status: [kube-sonobuoy]
Details of containers that are not ready:
kube-sonobuoy: waiting: ImagePullBackOff, Back-off pulling image "schnake/sonobuoy:987"
Status: Pending, Reason: ContainersNotReady, containers with unready status: [kube-sonobuoy]
Details of containers that are not ready:
kube-sonobuoy: waiting: ContainerCreating
Status: Pending, Reason: Unschedulable, 0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.
Status: Running
Status: Running
```
The "Status: Running" should Never Happen™, I left it there to have a return for any code path

johnSchnake (approved):
Looks good to me; this is great. Thanks @eturra

johnSchnake:
Oh, before I merge can you squash your commits?

johnSchnake:
Since its green and I'd like to commit and release today; I'm just going to squash through the UI. I hope you dont mind.

Issue #1605, opened 2022-01-28
issues with migrating from 0.51 to 0.56 in an air-gapped environment


johnqa:
**What steps did you take and what happened:**
I have a yaml file describing e2e testing generated with sonobuoy 0.51 in an air-gapped environment. Running tests using this file works perfectly.

I tried to upgrade to 0.56 in many ways and never managed to make it work:
1. If I just change in yaml file sonobuoy version from 0.51 to 0.56, I have unknown results for the execution of e2e plugin and the fix with pointing to a differente results dir is not working;
2. If I generate a new yaml file i have 
`WARN[0000] no pods found with label "sonobuoy-component=aggregator" in namespace sonobuoy 
WARN[0000] retrying with deprecated label "run=sonobuoy-master" 
WARN[0000] no pods found with label "run=sonobuoy-master" in namespace sonobuoy 
WARN[0000] Aggregator pod not found, using default pod name "sonobuoy": no pods found with label "run=sonobuoy-master" in namespace sonobuoy 
ERRO[0000] error attempting to run sonobuoy: could not retrieve sonobuoy pod ` 
and no pod is even started.
3. When I try to reconcile the files generated by 0.56 and the working one for 0.51 I have an issue with image being pulled form `k8s.gcr.io/e2e-test-images/agnhost:2.21` even if I do no changes in my repos list (that works with 0.51).

Is there a known migration process from 0.51 to 0.56 in order to have a properly yaml executable file? 

**Environment:**

- Sonobuoy version: 0.56
- Kubernetes version: (use `kubectl version`): 1.20.8
- Kubernetes installer & version: Rancher 2.5.9
- OS (e.g. from `/etc/os-release`): redhat

johnSchnake:
Sorry that you're having so many problems. Let me try and follow up on these questions/issues:

> If I just change in yaml file sonobuoy version from 0.51 to 0.56, I have unknown results for the execution of e2e plugin and the fix with pointing to a differente results dir is not working;

[...]

Issue #1223, opened 2021-03-18
Airgapped image config does not override all container registries in e2e conformance tests


jcstryker:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

So in our cluster we use an OPA Gatekeeper based admission webhook to enforce the sourcing of images from an internal mirror.

This puts us in the "airgapped" use case for sonobuoy. So we use the `--e2e-repo-config` flag to map the test images to internal sources.

However it seems that not all image registries are being overwritten, we have a few failing tests due to the use of outside images e.g.

- `Kubernetes e2e suite.[sig-apps] Deployment deployment should support rollover [Conformance]`
```
ReplicaFailure,Status:True,Reason:FailedCreate,Message:admission webhook "validation.gatekeeper.sh" denied the request: [denied by repo-is-artifactory] container <redis-slave> has an invalid image repo <gcr.io/google_samples/gb-redisslave:nonexistent>, allowed repos are
```
- `Kubernetes e2e suite.[sig-apps] Deployment deployment should support proportional scaling [Conformance]`
```
ReplicaFailure,Status:True,Reason:FailedCreate,Message:admission webhook "validation.gatekeeper.sh" denied the request: [denied by repo-is-artifactory] container <httpd> has an invalid image repo <webserver:404>, allowed repos are
```
- `Kubernetes e2e suite.[sig-apps] Daemon set [Serial] should rollback without unnecessary restarts [Conformance]`
```
Mar 18 21:03:23.674: INFO: At 2021-03-18 20:57:44 +0000 UTC - event for daemon-set: {daemonset-controller } FailedCreate: Error creating: admission webhook "validation.gatekeeper.sh" denied the request: [denied by repo-is-artifactory] container <app> has an invalid image repo <foo:non-existent>, allowed repos are 
```
- `Kubernetes e2e suite.[sig-api-machinery] AdmissionWebhook [Privileged:ClusterAdmin] should mutate pod and apply defaults after mutation [Conformance]`
```
Message: "admission webhook \"validation.gatekeeper.sh\" denied the request: [denied by repo-is-artifactory] container <webhook-added-init-container> has an invalid image repo <webhook-added-image>, allowed repos are 
```
Using this repo config yaml (produced by `sonobuoy gen default-image-config`
```
buildImageRegistry: artifactory-k8sgcr.<mirror>/build-image
dockerGluster: artifactory-dockerhub.<mirror>/gluster
[...]

PR #1639, opened 2022-03-09
Add sonobuoy images list command

sathieu:

`sonobuoy images` prints the image list, but there is no dry-run. This PR adds a `sonobuoy images list` command with dry-run flag.

- Related to #1605 and #1223

**Release note**:
```
Add sonobuoy images list command
```

### DIFF
--- a/cmd/sonobuoy/app/images.go
+++ b/cmd/sonobuoy/app/images.go
@@ -53,6 +53,24 @@ var (
 	transformSink = map[string][]func(*manifest.Manifest) error{}
 )
 
+func runListImages(flags imagesFlags) {
+	var client image.Client
+	if flags.dryRun {
+		client = image.DryRunClient{}
+	} else {
+		client = image.NewDockerClient()
+	}
+	version, err := getClusterVersion(flags.k8sVersion, flags.kubeconfig)
+	if err != nil {
+		errlog.LogError(err)
+		os.Exit(1)
+	}
+	if err := listImages(flags.plugins, version, client); err != nil {
+		errlog.LogError(err)
+		os.Exit(1)
+	}
+}
+
 func NewCmdImages() *cobra.Command {
 	var flags imagesFlags
 	// Main command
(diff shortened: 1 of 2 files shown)


cmd/sonobuoy/app/images.go
```diff
@@ -91,6 +92,39 @@ func NewCmdImages() *cobra.Command {
...
+			if err := listImages(flags.plugins, version, client); err != nil {
+				errlog.LogError(err)
+				os.Exit(1)
+			}
+		},
+		Args: cobra.ExactArgs(0),
+	}
+	AddKubeconfigFlag(&flags.kubeconfig, listCmd.Flags())
+	AddPluginListFlag(&flags.plugins, listCmd.Flags())
+	AddDryRunFlag(&flags.dryRun, listCmd.Flags())
```
johnSchnake, line 122: Any reason not to add the dryRun flag to the `images` command as well? Apparently we thought it had a dryrun option since the logic references but maybe that was just boilerplate.

johnSchnake (changes requested):
Thanks for the submission!

Just a small change requested.

I'm curious what your use case is though; the dryrun client for list doesn't do anything particularly useful since its just hardcoded values for a hardcoded version.

We _could_ add other versions and make that work but the concern is that it can get out of sync with the actual values and creates another dependency between us and the k8s releases.

In addition, the point of the `images` commands in general are to pull/push them to your airgapped registry; so if you can't reach the repos to pull/list them why list them in dry-run mode at all?

sathieu:
@johnSchnake I want to automate as much as possible. The script will do like this:

1. get the sonobuoy version from a var
2. sync the sonobuoy version to an offline registry
3. run sonobuoy to get the systemd-logs image and the conformance image
4. sync those two images
5. run the conformance image to list e2e images
6. sync those few images

The only missing part is step 3. This PR doesn't fill this need. Another flag should be added, or maybe the dry-run should not return e2e images at all. WDYT?

cmd/sonobuoy/app/images.go
```diff
@@ -91,6 +92,39 @@ func NewCmdImages() *cobra.Command {
...
+			}
+			version, err := getClusterVersion(flags.k8sVersion, flags.kubeconfig)
+			if err != nil {
+				errlog.LogError(err)
+				os.Exit(1)
+			}
+			if err := listImages(flags.plugins, version, client); err != nil {
+				errlog.LogError(err)
+				os.Exit(1)
+			}
```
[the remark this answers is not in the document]
sathieu: Done.

cmd/sonobuoy/app/images.go
sathieu: Done.

sathieu:
@johnSchnake I've pushed a newer commit with a new `--e2e-images` flag, defaulting to `true`. Please launch the CI and review :pray:.

johnSchnake:
For the systemd-logs image (and sonobuoy) you can run:
```
sonobuoy images -p systemd-logs
```
I would think the dry-run flag for the e2e plugin should just cause it to not run docker at all. So for e2e it should just figure out the image location and print that (and it seems the sonobuoy image is listed as well by default too).

Right now it seems to be hardcoded for the v1.19 version. As my previous comment mentioned, I dont think that provides much value.

I'm just not sure why step (5) is separated out from the e2e image itself. When you run `sonobuoy images` it also lists the conformance image itself.

So you can get all the images you need from:
`sonobuoy images -p systemd-logs && sonobuoy images -p e2e`

Either way, adding the dryrun flag and an explicit `list` command is fine with me though 👍 We can work out the proper behavior of dry-run in a follow-up.

sathieu:
@johnSchnake Good catch, I've improved the PR to only change two things:

- Add sonobuoy images list command (And add --druy-run flag to sonobuoy images command)
- Return no e2e images instead of outdated info

So, the script would look like this:

1. get the sonobuoy version from a var
2. download the sonobuoy binary
3. run `sonobuoy images list --dry-run` to get the sonobuoy, systemd-logs and conformance images
4. sync those 3 images
5. run the conformance image to list e2e images
6. sync those few images

Please retrigger the CI and review 🙏 .

NB: Some context [here](https://gitlab.com/kubitus-project/kubitus-installer/-/merge_requests/769).

johnSchnake:
Just a bad import to fix; once that is done it looks great. Ran some manual checks locally.

I'll keep an eye out for the next fixup and if everything is green I'll merge. Thanks for the submission, I think this is great.

sathieu:
@johnSchnake Fixed the import error. Thanks!

johnSchnake (approved):

PR #1673, opened 2022-04-14
add description flag and function to set description

### LINKED ISSUE #1647: Gen plugin should have flag for adding a description
When creating a plugin I should be able to specify most everything via flags since this is our preferred way of creating plugins from scratch. The description and source URL lack flags to set their value though.

DiptoChakrabarty:
Signed-off-by: DiptoChakrabarty <diptochuck123@gmail.com>

adds a flag to set a plugin description and underlying function logic to set description

- Fixes #1647 

**Release note**:
```
release-note
```

### DIFF
--- a/cmd/sonobuoy/app/gen_plugin_def.go
+++ b/cmd/sonobuoy/app/gen_plugin_def.go
@@ -130,7 +130,7 @@ func NewCmdGenPluginDef() *cobra.Command {
 	)
 
 	genPluginSet.StringVarP(
-		&genPluginOpts.def.SonobuoyConfig.Description, "description", "d", "",
+		&genPluginOpts.def.SonobuoyConfig.Description, "description", "desc", "",
 		"Description for the plugin",
 	)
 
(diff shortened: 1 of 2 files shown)


johnSchnake:
Hi and thanks for the PR, however I'm skeptical of the approach.
 - there are other fields that could be edited on the plugin, so why only a command to edit the description?
 - the PR description says its adding a flag but the PR actually adds another command specifically to modify the description field
 - the `sonobuoy plugin list` command shows where there file is located; if you want to edit it you can open the file and tweak anything you'd like about the plugin, that is the approach I would recommend for editing installed plugins

johnSchnake:
Let me know what you think about my comments above and we can decide a direction on this.

johnSchnake:
Seeing that you linked back to the pre-existing issue I see that this is just a misunderstanding of solution.

On the command `gen plugin` a flag should be able to edit the description just like all the other plugin building flags.

This is the file and approach you should take: https://github.com/vmware-tanzu/sonobuoy/blob/main/cmd/sonobuoy/app/gen_plugin_def.go#L107-L109

johnSchnake:
This fails with a panic since `d` is already used as a shorthand for the debug mode. Just remove the shorthand value for this. If you think `description` is too verbose you can change it to `desc`, I'm fine either way.

```
$ sonobuoy gen plugin -h
panic: unable to redefine 'd' shorthand in "plugin" flagset: it's already used for "description" flag

goroutine 1 [running]:
github.com/spf13/pflag.(*FlagSet).AddFlag(0xc000128c00, 0xc000292320)
...
...
```

PR #1681, opened 2022-04-22
Fix aggregator to wait forever when timeout is zero.

bostrt:
The help documentation states that a `--timeout 0` will wait indefinitely. This PR changes the type of `context` used when timeout is zero such that a timeout will not occur. 

Note that if `--timeout 0` is set right now (without this PR change) then the aggregator instantly times out waiting for workers.

https://github.com/vmware-tanzu/sonobuoy/blob/bbc49da2ac20431b76a0aa5d6a17fe5331bf0161/cmd/sonobuoy/app/args.go#L317-L323

**What this PR does / why we need it**: To match documentation and wait indefinitely for underlying jobs to complete. 

**Special notes for your reviewer**: I'm not familiar with Sonobuoy's set of tests and am not sure if this timeout==0 situation is covered. Any guidance would be appreciated. 

**Release note**:
```
none
```


johnSchnake:
Thanks for the submission! If you can sign your commit using `git amend --signoff` I'll approve. There is a vulnerability but that is already being addressed in another PR.

johnSchnake:
That vulnerability PR actually just merged; so if you want to rebase when you signoff you should see all green too.

bostrt:
Should be good-to-go now (pending tests). Thanks!

johnSchnake:
Looks great; the DCO check is hung as yellow for some reason but I can see the signoff.

Thanks for the PR!

PR #1690, opened 2022-05-06
Fix bug where 'retrieve' fails if filepath diectories DNE

### LINKED ISSUE #1689: Sonoboy retrieve fails saving on a different path
Running sonobuoy retrieve fails if the path specified using -f contains a path to a directory.

```
$ sonobuoy retrieve -f /tmp/work/results.tar.gz
error retrieving results: non-zero data ¸ (byte 184) read after tar EOF
$ sonobuoy retrieve 
202205050313_sonobuoy_9e1c1194-9068-4899-b8b2-000c8bb19fd1.tar.gz
$ sonobuoy retrieve -f /tmp/work/results.tar
error retrieving results: non-zero data ¸ (byte 184) read after tar EOF
$ sonobuoy retrieve -f results.tar
results.tar
```

Able to save the results tar ball on a directory different to where the  sonobouy binary  is located

**Environment:**

- Sonobuoy version: v0.56.4

johnSchnake:
The logic takes a path and a file; if the file has path values in it
that do not exist, then os.Create was failing but the MkdirAll logic
wasn't looking for paths in that.

To fix this, I just reconcile the directory/filename to get the full path
and process it accordingly sooner.

Fixes #1689

No tests really hit this but its a pretty trivial change. Just going forward with it having been tested manually. In my opinion this isn't worth a big test setup to check at this time.

**Release note**:
```
Fixed a bug where `sonobuoy retrieve -f <filename>` fails if the filename contains path elements that do not exist. 
```

### DIFF
--- a/pkg/client/retrieve.go
+++ b/pkg/client/retrieve.go
@@ -107,17 +107,24 @@ func (c *SonobuoyClient) RetrieveResults(cfg *RetrieveConfig) (io.Reader, <-chan
 
 /** Everything below this marker originally was copy/pasta'd from k8s/k8s. The modification are:
   exporting UntarAll, returning the list of files created, and the fix for undrained readers. Also
-  added the filename override as opposed to using a file prefix.  **/
+  added the filename override as opposed to using a file prefix and reconciled the filename/destDir
+  values.  **/
 
 // UntarAll expects a reader that contains tar'd data. It will untar the contents of the reader and write
 // the output into destDir. If filename specified, then files will be named "filename, filename-01, filename-02". It returns a list of all the
 // files it created.
 func UntarAll(reader io.Reader, destDir, filename string) (filenames []string, returnErr error) {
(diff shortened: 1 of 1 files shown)


pkg/client/retrieve.go
```diff
@@ -130,17 +134,15 @@ func UntarAll(reader io.Reader, destDir, filename string) (filenames []string, r
...
-		if err != nil {
-			returnErr = errors.Wrap(err, "failed to drain the tar reader")
+		err := drainReader(reader)
+		switch {
+		case err == nil:
+			return
+		case returnErr == nil && err != nil:
+			returnErr = err
+		case returnErr != nil && err != nil:
+			logrus.Warnf("Error while draining reader (%v) but there was a previous error that may be the root cause: %v", err, returnErr)
```
johnSchnake, line 144: This section was just refactored for clarity when hitting both errors. Found when debugging the original issue so I included it here.

PR #1693, opened 2022-05-07
Moved the logic for gathering logs to before plugin cleanup

johnSchnake:
The cleanup logic must occur after queries or else pod logs
won't be gathered (it will be a race, at best).

There is a test for this code but I think it was a race that
worked in our favor so it would pass in CI but with real plugins
it would not. Will follow up with other test modifications but want
to get this fix in asap.

- Fixes #

**Release note**:
```
release-note
```


scripts/build_funcs.sh
```diff
@@ -67,7 +67,7 @@ integration() {
         --env KUBECTL_CLI="$KUBECTL_CLI" \
         --network host \
         "$BUILD_IMAGE" \
-    go test ${VERBOSE:+-v} -tags=integration "$GOTARGET"/test/integration/...
+    go test ${VERBOSE:+-v} -tags=integration "$GOTARGET"/test/integration/... -run TestQuick
```
johnSchnake, line 70: Needs changed back before merge.

test/integration/sonobuoy_integration_test.go
```diff
@@ -125,6 +125,7 @@ func mustRunSonobuoyCommandWithContext(ctx context.Context, t *testing.T, ns, ar
 		} else {
 			t.Logf("Checked sonobuoy logs, got: %v", o.String())
 		}
+		t.FailNow()
```
johnSchnake, line 128: Noticed when testing locally this was missing. Will move it to its own PR though.

test/integration/sonobuoy_integration_test.go
```diff
@@ -439,7 +461,7 @@ func saveToArtifacts(t *testing.T, p string) (newPath string) {
 	var stdout, stderr bytes.Buffer
 
 	// Shell out to `mv` instead of using os.Rename(); the latter caused a problem due to files being on different devices.
-	cmd := exec.CommandContext(context.Background(), bash, "-c", fmt.Sprintf("mv -r %v %v", origFile, artifactFile))
+	cmd := exec.CommandContext(context.Background(), bash, "-c", fmt.Sprintf("mv %v %v", origFile, artifactFile))
```
johnSchnake, line 464: Rebasing onto main should have removed this from the diff. Will double check.

PR #1703, opened 2022-05-12
Remove unnecessary volume from gen plugin command output

### LINKED ISSUE #1649: Plugins dont need empty resources or volume mount
When creating a plugin via `sonobuoy gen plugin` or even looking at the default e2e/systemd-logs plugins they still show

```
  resources: {}
  volumeMounts:
  - mountPath: /tmp/sonobuoy/results
    name: results
```

Although the volume mount is correct here (the default value) sonobuoy later makes sure that all the plugins agree and that its what the sonobuoy worker sidecar is expecting. So the value here doesn't really matter; what the sidecar expects is what matters.

It would make things less verbose/confusing if we just removed those sections.

### LINKED ISSUE #1691: Gen Plugin includes unnecessary info
The results volumeMount gets added on the server and the user doesn't have to worry about it at all. It is explicitly ignored if it conflicts with what the aggregator was configured to use.

In addition, no need to print an empty `resources` value.

i.e. Remove these lines from `gen plugin [e2e|systemd-logs]`
```
  resources: {}
  volumeMounts:
  - mountPath: /tmp/sonobuoy/results
    name: results
```

matheusfm:

Remove unnecessary volume from `gen plugin` command output.

- Fixes #1691
- Fixes #1649

An empty resource is always marshalled because `Resources` is not a pointer: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2368

**Release note**:
```
release-note
```


johnSchnake:
Thanks, this is great but even though the empty resources is going to get serialized I'd like to remove it from the output, even if it is equivalent YAML.

Its not hard to do but is orthogonal to this. I'll do that in a followup 👍

PR #1718, opened 2022-05-25
Include Image name in image pull error message

### LINKED ISSUE #1707: Improve error message from errImagePull
See #1706 as an example; the message is so close to being extremely helpful. The error state was identified and reported as errImagePull and specifies which container is the problem. But it doesn't show which image it is trying to pull.

If we just tweak that error message it would be much more helpful.

DiptoChakrabarty:
This includes the image name which causes the errImagepull error to make the error message more meaningful
- Fixes #1707 

**Release note**:
```
release-note
```


johnSchnake (approved):
Thanks for the contribution 👍

PR #1723, opened 2022-05-31
User provided ConfigMap validation

### LINKED ISSUE #1663: Run basic validation on configmaps provided by users
When adding a configmap via the `config-map` field in a plugin, if we see they are adding a yaml file, why not parse it and try and see if its valid yaml? Especially with the whitespace of nesting it this can cause issues and I've hit the problem more than once.

I assume we could do a basic json/yaml validation, more if others request it.

If we don't have this, we don't get signal until the configmap contents are used by the plugin which means the entire run has to start, the error occurs, and someone has to try and dig into what the problem is.

Instead, if we validate in the CLI we can just shot this back on the clientside before anything else has to happen and give clear error messages.

andrewyunt:
Adds basic validation for the value of the ConfigMap field in plugin manifest. Parses YAML using the Unmarshal function provided by the Go yaml package. 

- Fixes #1663

**Release note**:
```
release-note
```


pkg/client/interfaces_test.go
```diff
@@ -32,6 +34,72 @@ func TestConfigValidation(t *testing.T) {
...
+		{
+			desc: "gen config with invalid yml value is not valid",
+			config: &GenConfig{
+				StaticPlugins: []*manifest.Manifest{
+					{
+						SonobuoyConfig: manifest.SonobuoyConfig{
+							PluginName: "test",
+						},
+						ConfigMap: map[string]string{
+							"test.yml": "test: \"in\"valid\"",
```
johnSchnake, line 95: Totally fine to leave like this but a tip that I find helpful is that if you have a string with quotes, use the backtick instead of quotes. It will prevent `"` from being a special character and it doesnt have to be escaped.

pkg/client/interfaces_test.go
```diff
@@ -32,6 +34,72 @@ func TestConfigValidation(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			valid:         false,
+			expectedError: "failed to parse value of key test.yml in ConfigMap for plugin test: error converting YAML to JSON: yaml: did not find expected key",
+		},
```
johnSchnake, line 102: Current test cases are good; lets add just a few more to cover a few different edge cases. 
 - A plugin with 2 configmaps, one valid one invalid to ensure we never introduce an early exit from the checks
 - A case where its not valid yaml but its not named "*.yaml" so we ensure we only test the intended files

pkg/client/interfaces.go
```diff
@@ -107,6 +108,15 @@ func (gc *GenConfig) Validate() error {
 		if !nameRegexp.MatchString(m.SonobuoyConfig.PluginName) {
 			return fmt.Errorf("invalid plugin name %q; name must only include lowercase alphanumeric values '.' or '-'. This is due to Kubernetes requiring various values to follow RFC 1123 for valid subdomain names", m.SonobuoyConfig.PluginName)
 		}
+
+		for key, value := range m.ConfigMap {
+			if strings.HasSuffix(key, "yml") || strings.HasSuffix(key, "yaml") {
```
johnSchnake, line 113: nit: I didnt catch this when we looked at it, but I'd add the dot as well to look for the suffix: `.yml` and `.yaml`

johnSchnake (changes requested):
Just a few changes, quick to fixup.

Also, to pass the DCO check you need to do:

`git amend --signoff` so that it signs the commit with your email. We'll have to make sure your local git is configured with the right info so that works though.

Lastly, when possible we try and squash everything down to a single commit and do fast-forward merges onto main rather than merges. I can help you with getting the flow down.

johnSchnake (approved):

Issue #1727, opened 2022-06-01
Add documentation for commands to sonobuoy.io


andrewyunt:
**Describe the solution you'd like**
Add documentation for cobra cli commands to sonobuoy.io website. It would be useful to have markup generated from all of the commands in the Sonobuoy to help with onboarding of new developers as well as users.

PR #1728, opened 2022-06-02
Add CLI docs to the site

johnSchnake:
Uses cobra to generate markdown docs for the command tree.

Known issue to resolve is that the default image used is based on
the sonobuoy version running the command. Right now that is just
a locally build version which is marked with a sha and as dirty.

Need a little think about the best way to resolve that since the
docs wont be changed on every commit, and the 'main' branch of sonobuoy
would be targeting different images. We just need to make sure that
tagged releases have the right version when those docs are published.

Fixes https://github.com/vmware-tanzu/sonobuoy/issues/1727


cmd/sonobuoy/app/gen_cli_docs.go
```diff
@@ -0,0 +1,24 @@
+package app
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/cobra/doc"
+	"github.com/vmware-tanzu/sonobuoy/pkg/errlog"
+)
+
+func NewCmdGenCLIDocs() *cobra.Command {
+	var docsCmd = &cobra.Command{
```
andrewyunt, line 10: May be better to use := to maintain consistency, but this is just personal preference

andrewyunt (approved):

cmd/sonobuoy/app/gen_cli_docs.go
johnSchnake: 👍 I tend to prefer `:=` over `var...` as well; this was actually copied/modified from another command and I just didnt mind enough to change it. Its good that you're thinking of consistency/preference though. Just wasn't worth a new push/test/cycle.

PR #1737, opened 2022-06-08
Add readable mode to results command

### LINKED ISSUE #1628: Newlines in results shown as `\n`
From slack:

```
tanzu conformance results --mode detailed --plugin e2e $(tanzu conformance retrieve) | jq '.  | select(.status == "failed") | .details
```

For this one, I’m failing at getting jq to not just write \n for new lines. I tried -r, but that didn’t do anything for system-out.

---

The issue is that when the junit stores newlines as `\n` so when we copy the data to the yaml we keep them as `\n` strings. But then when you try and print the data (or view it in the file) its hard to read and you have to do a manual replace on those.

Possible solutions:
 - When we read the xml we should be doing that translation automatically
 - when using the `results` command we should alter the data stream on the fly to make that replacement

The first solution is nice since it helps when you just open up the tarball. But it also may cause issues if we are trying to print it in a format that doesn't like actual newlines.

I think we need to just try it out and see if it works well.

andrewyunt:
Adds "readable" mode to the results command. Replaces \n with a new line and \t with a tab to make the output to stdio from the sonobuoy_results.yaml file more readable (dump mode does not do this).

- Fixes #1628

Need to test the output of the command using readable mode. I was thinking we could have printSinglePlugin and printHealthSummary also return strings of what is printed to stdout to be used for testing, but that approach may be overkill for the sake of testing.

### DIFF
--- a/cmd/sonobuoy/app/results.go
+++ b/cmd/sonobuoy/app/results.go
@@ -17,6 +17,7 @@ limitations under the License.
 package app
 
 import (
+	"bytes"
 	"compress/gzip"
 	"encoding/json"
 	"fmt"
@@ -46,6 +47,9 @@ const (
 
 	// resultModeDump will just copy the post-processed yaml file to stdout.
 	resultModeDump = "dump"
+
+	//resultModeReadable will copy the post-processed yaml file to stdout and replace \n and \t with new lines and tabs respectively.
+	resultModeReadable = "readable"
 
 	windowsSeperator = `\`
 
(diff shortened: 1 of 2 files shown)


cmd/sonobuoy/app/results.go
```diff
@@ -235,6 +247,17 @@ func printSinglePlugin(input resultsInput, r *results.Reader) error {
 		}
 		_, err = io.Copy(os.Stdout, fReader)
 		return err
+	} else if input.mode == resultModeReadable {
+		fReader, err := r.PluginResultsReader(input.plugin)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get results reader for plugin %v", input.plugin)
+		}
+		data, err := io.ReadAll(fReader)
```
johnSchnake, line 255: So this is a reasonable first approach but it won't work well when the files are large.

What we need to do is, just like the `dump` mode, is to copy from the reader (which uses a smaller buffer and series of reads rather than one big chunk) but to replace the characters as we copy.

I can give you some tips on this too.

johnSchnake (changes requested):
Again, we can talk on our sync later today about the approach and some tests for this.

PR #1738, opened 2022-06-08
Add clusterRead to aggregator permissions help

### LINKED ISSUE #1704: CLI help doesnt list proper permissions options
`sonobuoy run -h`

shows 
```
      --aggregator-permissions string            Type of aggregator permission to use in the cluster. Allowable values are [namespaced, clusterAdmin] (default "clusterAdmin")
```

Double check the values, make sure they come from code if possible so they can't get out of sync. I think there is clusterAdmin, clusterRead, and namespaceAdmin but I could be wrong.

Trying to set a random value fails too without a better message. Improve that to list the valid options too.

andrewyunt:
Updates the help for the aggregator permissions flag to add the clusterRead value. Should have all of the correct permissions listed now.

- Fixes #1704

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -165,7 +165,7 @@ func AddSecurityContextMode(mode *string, flags *pflag.FlagSet) {
 func AddAggregatorPermissionsFlag(mode *string, flags *pflag.FlagSet) {
 	flags.StringVar(
 		mode, aggregatorPermissionsFlag, "clusterAdmin",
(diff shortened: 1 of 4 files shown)


johnSchnake:
The tests that fail are just because some golden files need updated. When we pair later today I'll show you how to update those (its just a certain script to run)

johnSchnake:
Still missing the golden files; you have to source the build_funcs and then call the function to update the local golden files. I think it might be `update_local` but just peek at the funcs.

johnSchnake (approved):

PR #1744, opened 2022-06-21
Add flags to allow customizing the service account sonobuoy uses

### LINKED ISSUE #1653: namespaceAdmin: Allow running with a custom service account

We would like to be able run sonobuoy in cases where we are restricted to a namespace, don't have permissions to create RBAC resources, and are so required to use a pre-created serviceaccount.

That would mean sonobuoy would need to run with a serviceaccount whose name is provided by the user, and sonobuoy would not itself attempt to create any RBAC resources.

For our internal usage, we already have a [fork which contains this change](https://github.com/vmware-tanzu/sonobuoy/compare/main...nokia:customServiceAccount), so if this proposal sounds reasonable, we could open a new Pull request to discuss specifics.

The version built from the fork is triggered in the mentioned restricted mode with:
```
--skip-preflight --aggregator-permissions namespaceAdmin --rbac Disable --service-account <existing_serviceaccount_name>
```

**Environment:**

- Sonobuoy version: v0.56.2
- Kubernetes version: (use `kubectl version`): v1.21.5

tanmay-g:
This PR adds two flags to `run/gen`:
```
--existing-service-account                 If true, use an existing service account, else try to create one.
--service-account-name string              Name of the service account to be used by sonobuoy. (default "sonobuoy-serviceaccount")
```

- `--service-account-name` takes in a service account name that the sonobuoy pod will use.
  - If `--existing-service-account` is also set, the user is responsible for the existence and roles of the service account; if not, sonobuoy will simply change the name of the service account it creates for itself.
- If `--existing-service-account` alone is passed, the default service account ("sonobuoy-serviceaccount") should be created by the user with appropriate roles.

This allows sonobuoy to run in cases where it doesn't have permissions to create RBAC resources, or is otherwise required to use a pre-created serviceaccount, whose roles are managed by the user.

While not necessary, this is more likely to occur in cases where sonobuoy is restricted to a namespace (and uses the existing `--aggregator-permissions namespaceAdmin` flag). The following combination of flags would support such a case:
``` 
--skip-preflight --aggregator-permissions namespaceAdmin --rbac Disable --service-account-name <SERVICEACCOUNT_NAME> --existing-service-account
```

- Fixes #1653

**Release note**:
```
Added flags to allow customizing the service account sonobuoy uses
```


cmd/sonobuoy/app/args.go
```diff
@@ -1,6 +1,8 @@
 /*
 Copyright 2018 Heptio Inc.
 
+© 2022 Nokia
```
johnSchnake, line 4: So the historic heptio copyrights remain but in general we now ask that changes be tagged with `Copyright <yr> the Sonobuoy Project contributors`

pkg/config/config.go
```diff
@@ -1,6 +1,8 @@
 /*
 Copyright 2018 Heptio Inc.
 
+© 2022 Nokia
```
johnSchnake, line 4: Same comment as other copyright.

johnSchnake:
You're going to be getting a lot of "golden" test failures since you added a new field into the config.

Run:
```
source ./scripts/build_funcs.sh; update_local
```

from the root of the directory and it'll make the necessary changes. Use `git diff` to check the changes look right and then you can add them into the commit.

johnSchnake:
Once the copyright/golden files are updated and tests green, lgtm 👍 Thanks for the contribution.

cmd/sonobuoy/app/args.go
tanmay-g: The copyright message has been updated as suggested

pkg/config/config.go
tanmay-g: The copyright message has been updated as suggested, in all relevant files (I left out the `.golden` files)

tanmay-g:
> Once the copyright/golden files are updated and tests green, lgtm 👍 Thanks for the contribution.

I've updated the copyright statements, and also the golden files (the changes there are basically the addition of a `ServiceAccountName` key in the config, and the usage help for the new params in the CLI).

johnSchnake (approved):

PR #1745, opened 2022-06-21
Add readable mode to results command

andrewyunt:
*What this PR does / why we need it:*
Adds "readable" mode to the results command. Replaces \n with a new line and \t with a tab to make the output to stdio from the sonobuoy_results.yaml file more readable (dump mode does not do this).

Which issue(s) this PR fixes

Fixes https://github.com/vmware-tanzu/sonobuoy/issues/1628

*Special notes for your reviewer:*
Need to test the output of the command using readable mode. I was thinking we could have printSinglePlugin and printHealthSummary also return strings of what is printed to stdout to be used for testing, but that approach may be overkill for the sake of testing.

### DIFF
--- a/cmd/sonobuoy/app/results.go
+++ b/cmd/sonobuoy/app/results.go
@@ -17,6 +17,7 @@ limitations under the License.
 package app
 
 import (
+	"bytes"
 	"compress/gzip"
 	"encoding/json"
 	"fmt"
@@ -46,6 +47,9 @@ const (
 
 	// resultModeDump will just copy the post-processed yaml file to stdout.
 	resultModeDump = "dump"
+
+	//resultModeReadable will copy the post-processed yaml file to stdout and replace \n and \t with new lines and tabs respectively.
+	resultModeReadable = "readable"
 
 	windowsSeperator = `\`
 
@@ -83,7 +87,7 @@ func NewCmdResults() *cobra.Command {
 	)
 	cmd.Flags().StringVarP(
 		&data.mode, "mode", "m", resultModeReport,
-		`Modifies the format of the output. Valid options are report, detailed, or dump.`,
+		`Modifies the format of the output. Valid options are report, detailed, readable, or dump.`,
 	)
 	cmd.Flags().StringVarP(
 		&data.node, "node", "n", "",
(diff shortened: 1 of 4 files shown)


cmd/sonobuoy/app/results.go
```diff
@@ -235,6 +259,17 @@ func printSinglePlugin(input resultsInput, r *results.Reader) error {
...
 		return err
+	} else if input.mode == resultModeReadable {
+		fReader, err := r.PluginResultsReader(input.plugin)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get results reader for plugin %v", input.plugin)
+		}
+		writer := &humanReadableWriter{os.Stdout}
+		_, err = io.Copy(writer, fReader)
+		if err != nil {
+			panic(err.Error())
```
johnSchnake, line 270: Panic is good for quick POC and iteration, but you'll want to return the error wrapped with some context.

```
return errors.Wrapf(err, "failed to copy data for plugin %v", input.plugin")
```

cmd/sonobuoy/app/results_test.go
```diff
@@ -16,10 +16,97 @@ limitations under the License.
...
+
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			buffer := bytes.Buffer{}
+			writer := humanReadableWriter{&buffer}
+			fmt.Fprintf(&writer, tc.input)
+			for _, contains := range tc.contains {
+				out := strings.Contains(buffer.String(), contains)
+				if out != tc.expected {
+					t.Errorf("Test case %s failed. Expected output: %v", tc.desc, tc.expected)
```
johnSchnake, line 103: nit: You dont really need to state the test case name since they are running as subtests (so each subtest name is based on the tc.desc field)

johnSchnake (changes requested):
One small change to fix up then this is ready to merge. Fixing the `nit` is optional.

johnSchnake (approved):

johnSchnake:
So 2 things left to do on this:
 - security vulnerability found; it requires bumping the version of the text library. Give this a go on your own and I can help this evening if needed. See this resource on how to update a dependency.
 - squash/merge

Issue #1720, opened 2022-05-25
Document build steps/flows for new developers


johnSchnake:
The build system changed a lot over the past few years and so we don't have great docs for how to actually start contributing to Sonobuoy.

We need docs to outline these main flows:
 - How to build sonobuoy, the CLI, locally
 - How to build the sonobuoy image locally
 - How to run unit tests
 - How to edit the website
 - How to run system-tests (optional for now)

Some context historically:

We used to have an ever expanding makefile that was hard to grok. _Multiple_ times we had bugs in the build/dev flows due to some odd Makefile behavior diverging from what someone would expect being used to bash.

In addition, our makefile was really just a list of `phony` targets being composed. As a result we just changed everything to be in the scripts/ directory as bash scripts/functions.

One file is for defining the functions on their own; the flow typically used is that the functions file gets sourced and then you call the functions you need. That is not as easy as the original `make ...`.

I hate adding more layers, but it may be the case that we want to add a Makefile for just the most basic things (leaving the scripts in place as they are for composability/predictability) but making an entry point in the makefile so its easy to document how to do the most common tasks.

johnSchnake:
This has come up a few times again; I will make a basic makefile as suggested above with a few targets. I'm thinking for `install, test, update_golden` may be enough. Maybe one for linting too. Once the PR is up we can debate that though, maybe something else will become obvious.

PR #1747, opened 2022-06-23
Adds a simple Makefile with the most commonly used options

johnSchnake:
This will make it easier for people new to the project to figure
out how to contribute or test changes locally.

We should update some of the documentation accordingly as well.

Fixes: #1720

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,19 @@
+.PHONY: build test update_golden help
+.DEFAULT_GOAL := help
+
+build: ## Builds the Sonobuoy binary for your local OS
+	# This is for local use; when building for other os/arch, use functions in ./scripts/build_funcs.sh directly.
+	@echo "Building the sonobuoy binary..."
+	go build
+	@echo "Complete. Use sonobuoy in the current working directory (e.g. ./sonobuoy)"
+
+test: ## Runs go tests
+	@echo "Running unit tests..."
+	source ./scripts/build_funcs.sh; unit_local
+
+golden: ## Runs go generate to produce version info
+	source ./scripts/build_funcs.sh; update_local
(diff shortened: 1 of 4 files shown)


johnSchnake:
The failure is unrelated to this change and will be fixed immediately.

johnSchnake:
Not changing logic so not going to be blocked by tests, but creating an issue for the failed test.

johnSchnake:
#1750 created regarding the test failure. Unrelated to this PR.

Makefile
```diff
@@ -0,0 +1,18 @@
+.PHONY: build test update_golden help
+.DEFAULT_GOAL := help
+
+build: ## Builds the Sonobuoy binary for your local OS
+	@echo "Building the sonobuoy binary..."
+	source ./scripts/build_funcs.sh; native
+	@echo "Complete. Use sonobuoy in the pwd (e.g. ./sonobuoy)"
```
andrewyunt, line 7: nit: Perhaps a statement that makes more sense would be "Use sonobuoy in the current directory" (pwd stands for print working directory)

johnSchnake: ha, I've always used it interchangeably with the current working directory (cwd) and intuitively assumed it was the 'present working directory'. TIL. 👍

johnSchnake: Looking at this build target I am going to actually swap it out for a simple `go build`. The reasoning is that I realized that using this build_func, just like the others, will reset the sonobuoy versions as if we are cutting a release.

The result will be what you experienced the other day: `sonobuoy run` will lead to setting the image to a non-existant image and you'll get ErrImagePull.

andrewyunt (approved):

PR #1752, opened 2022-06-27
Add command to get worker pod names

### LINKED ISSUE #1713: Flow to get the pod names
A command to print the name of the pod for a plugin.

I often want to inspect a pod in some capacity but the random names that the plugins get lead to an awkward flow where I have to use `kubectl get pods -n <namespace>` then copy and paste the name into the other command I want.

I'd like something like:

`kubectl get pod $(sonobuoy find-pod -n <ns> -p myplugin)`

Potentially it could even return a json object with all the resources of interest. Maybe a human readable format and a `--json` flag.

andrewyunt:
Adds a "get pods" command to fetch Sonobuoy worker pod names. Contains flags to specify namespace and plugin.

- Fixes #1713

### DIFF
--- a/cmd/sonobuoy/app/get.go
+++ b/cmd/sonobuoy/app/get.go
@@ -1,0 +1,84 @@
+package app
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/vmware-tanzu/sonobuoy/pkg/errlog"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type getFlags struct {
+	namespace string
+	plugin    string
+	kubecfg   Kubeconfig
+}
+
+func NewCmdGet() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "get",
+		Short: "Fetches Sonobuoy resources of a specified type",
+	}
+
+	// Add pod subcommand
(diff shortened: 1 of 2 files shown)


cmd/sonobuoy/app/get.go
```diff
@@ -0,0 +1,90 @@
...
+	pods, err := client.CoreV1().Pods(flags.namespace).List(context.TODO(), listOptions)
+
+	if err != nil {
+		return errors.Wrap(err, "could not retrieve list of pods")
+	}
+
+	if len(flags.plugin) > 0 {
+		fmt.Printf("Dispalying pods for plugin %s in namespace %s:\n", flags.plugin, flags.namespace)
+	} else {
+		fmt.Printf("Displaying pods in namespace %s:\n", flags.namespace)
```
johnSchnake, line 82: If the ultimate goal of this command was just to print the data for user consumption then these would be ok. However, the goal is to make this easily composable with other tools and so these lines get in the way since we cant pipe the data through other commands expecting pod names anymore.

johnSchnake (changes requested):
Just removing the headers will get this to mergable I think.

I had an instinct that things should be split out into more functions but the logic is so small that it doesn't matter much at this point. If we add more printing flags or something then we can do that so we can have tests to confirm the formatting. But right now you're just printing them so there isn't a real benefit.

johnSchnake (approved):

PR #1753, opened 2022-06-27
Update plugin ordering to latest Sonobuoy version

### LINKED ISSUE #1684: Allow plugin ordering

> References #631 #432 #39

Plugins should be able to run in a predetermined order.
 
As well described on #631 , the plugins could be able to run in a certain order configured on their manifests. The default behavior could stay the current execution (all in parallel), but it would be nice to run, for example in serial mode or a mix: A->B->C->(D&E)->F.

Also could be interesting if there's one flag if one plugin has failed, the following ones will not be executed (maybe also marked as failed?). It can be useful in preflights plugins which usually run on the begging to make some prechecks (like minimal resources, minimal performance checks, etc).

andrewyunt:
Plugins should be able to run in a predetermined order. Adds order key to sonobuoy-config (lower number has a higher priority)

- Fixes #1684

### DIFF
--- a/pkg/plugin/aggregation/run.go
+++ b/pkg/plugin/aggregation/run.go
@@ -23,6 +23,8 @@ import (
 	"net"
 	"net/http"
 	"path/filepath"
+	"sort"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -190,9 +192,27 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 		return errors.Wrapf(err, "couldn't get aggregator pod")
 	}
 
-	for _, p := range plugins {
-		logrus.WithField("plugin", p.GetName()).Info("Running plugin")
-		go aggr.RunAndMonitorPlugin(context.Background(), time.Duration(cfg.TimeoutSeconds)*time.Second, p, client, nodes.Items, cfg.AdvertiseAddress, certs[p.GetName()], aggregatorPod, progressPort, pluginResultsDir)
+	orderedPlugins := getOrderedPlugins(plugins)
+
+	logrus.Info("Received plugin launch order:")
+	for _, pluginGroup := range orderedPlugins {
+		for _, plugin := range pluginGroup {
+			logrus.Infof("%v: %v", plugin.GetName(), plugin.GetOrder())
+		}
+	}
+
+	for _, pluginGroup := range orderedPlugins {
+		var waitGroup sync.WaitGroup
+		waitGroup.Add(len(pluginGroup))
+		for _, plug := range pluginGroup {
+			logrus.WithField("plugin", plug.GetName()).Info("Running plugin")
+			go func(plugin plugin.Interface) {
+				logrus.Infof("Launching plugin %v with order %v", plugin.GetName(), plugin.GetOrder())
+				aggr.RunAndMonitorPlugin(context.Background(), time.Duration(cfg.TimeoutSeconds)*time.Second, plugin, client, nodes.Items, cfg.AdvertiseAddress, certs[plugin.GetName()], aggregatorPod, progressPort, pluginResultsDir)
+				waitGroup.Done()
+			}(plug)
+		}
+		waitGroup.Wait()
 	}
 
 	// 6. Wait for aggr to show that all results are accounted for
@@ -210,6 +230,30 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 			return nil
 		}
 	}
+}
+
+func getOrderedPlugins(plugins []plugin.Interface) [][]plugin.Interface {
+	pluginGroups := map[int][]plugin.Interface{}
+	for _, p := range plugins {
+		pluginGroups[p.GetOrder()] = append(pluginGroups[p.GetOrder()], p)
+	}
+
+	orderedPlugins := [][]plugin.Interface{}
+	for _, pluginGroup := range pluginGroups {
+		orderedPlugins = append(orderedPlugins, pluginGroup)
+	}
+
+	sort.Slice(orderedPlugins, func(i, j int) bool {
+		return orderedPlugins[i][0].GetOrder() < orderedPlugins[j][0].GetOrder()
+	})
+
+	for _, plugins := range orderedPlugins {
+		sort.Slice(plugins, func(i, j int) bool {
+			return plugins[i].GetName() < plugins[j].GetName()
+		})
+	}
+
+	return orderedPlugins
 }
 
 // Cleanup calls cleanup on all plugins
--- a/pkg/plugin/aggregation/run_test.go
+++ b/pkg/plugin/aggregation/run_test.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"math/big"
 	"os"
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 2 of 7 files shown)


pkg/plugin/manifest/manifest.go
```diff
@@ -50,6 +50,7 @@ type SonobuoyConfig struct {
 	// to the plugin source would be kept.
 	SourceURL string `json:"source-url,omitempty"`
 
+	Order int `json:"order"`
```
johnSchnake, line 53: Definitely add `,omitempty` on this. That will prevent all the 0 values from having to be explicitly written. Not a problem if they do, but it ensures that the tests work as-is (e.g. wont break existing users) and also cuts down on LOC changes for the goldenfiles.

pkg/plugin/aggregation/run.go
```diff
@@ -190,9 +192,14 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 		return errors.Wrapf(err, "couldn't get aggregator pod")
 	}
 
-	for _, p := range plugins {
-		logrus.WithField("plugin", p.GetName()).Info("Running plugin")
-		go aggr.RunAndMonitorPlugin(context.Background(), time.Duration(cfg.TimeoutSeconds)*time.Second, p, client, nodes.Items, cfg.AdvertiseAddress, certs[p.GetName()], aggregatorPod, progressPort, pluginResultsDir)
+	for _, pluginGroup := range getOrderedPlugins(plugins) {
+		var waitGroup sync.WaitGroup
```
johnSchnake, line 196: 👍 to using a WaitGroup. As someone new to Go I'm really happy to see you explored the options and found this. It really simplifies the logic.

pkg/plugin/aggregation/run.go
```diff
@@ -212,6 +219,27 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 	}
 }
 
+func getOrderedPlugins(plugins []plugin.Interface) [][]plugin.Interface {
```
johnSchnake, line 222: Since I really want to emphasize readability and maintainability and optimizations are not that important we should really consider simply using the `sort` package to sort these objects in terms of `order` then `name`; once they are ordered you can loop through and group them trivially.

That may be the most succinct way to execute this and the number of plugins will always (unless trying to prove a point) be < 100.

Open to other approaches but just in case you thought I wouldn't like using that simple approach I wanted to let you know I think its an easy, clear, maintainable option.

pkg/plugin/aggregation/run_test.go
```diff
@@ -152,7 +153,9 @@ func TestRunAndMonitorPlugin(t *testing.T) {
 			}
 
 			go func() {
-				a.RunAndMonitorPlugin(ctx, testTimeout, tc.plugin, fclient, nil, "testname", testCert, &corev1.Pod{}, "", "/tmp/sonobuoy/results")
+				var waitGroup sync.WaitGroup
```
johnSchnake, line 156: You typically try to keep the waitgroup logic all within the same function rather than passing it around. Is there a reason why you want to pass it into that function rather than just calling the wg.Done() after something gets sent to the done channel?

johnSchnake: Hm I think I see why you did it. Maybe we can talk about the options on a zoom call. Since RunAndMonitor would normally be synchronous, but we are calling it with `go RunAndMonitor...` we could do something slightly different to avoid passing in the WaitGroup (which requires a dev to follow the code path to figure out what happens to it). Instead we could do:

```
go func() {
  a.RunAndMonitorPlugin(...)
  waitGroup.Done()
}()
```
so the method signature doesnt have to change and in the same functions we see when the waitGroup gets marked as done.

I could be mistaken, but that seems a slightly better option. Its a small improvement though and wont seriously change the approach or tests.

pkg/plugin/interface.go
```diff
@@ -75,13 +76,26 @@ type Interface interface {
...
 	// GetSourceURL returns the URL where the plugin came from and where updates to it will be located.
 	GetSourceURL() string
+
+	// GetOrder returns the order in which the plugin is run. Plugins will wait until all other plugins with lower order are done before being launched.
+	GetOrder() int
+}
+
+// Definition defines a plugin's features, method of launch, and other
+// metadata about it.
+type Definition struct {
```
johnSchnake, line 86: I'm confused about this diff. I see you adding a `Definition` type but I get the impression this was already there but an invalid diff. The issue is that I just can't see the corresponding removal of this type from the previous commit (e.g. normally with invalid diffs you see it added in one place and removed from the original; here we just see this new type)

Do you know why it may be showing like this?

pkg/plugin/aggregation/run.go
```diff
@@ -190,9 +192,17 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 		return errors.Wrapf(err, "couldn't get aggregator pod")
 	}
 
-	for _, p := range plugins {
-		logrus.WithField("plugin", p.GetName()).Info("Running plugin")
-		go aggr.RunAndMonitorPlugin(context.Background(), time.Duration(cfg.TimeoutSeconds)*time.Second, p, client, nodes.Items, cfg.AdvertiseAddress, certs[p.GetName()], aggregatorPod, progressPort, pluginResultsDir)
+	for _, pluginGroup := range getOrderedPlugins(plugins) {
+		var waitGroup sync.WaitGroup
+		waitGroup.Add(len(pluginGroup))
```
johnSchnake, line 197: I think it would be really beneficial as this feature gets adoption that we add some extra logging here. We need to somehow provide some info about the grouping before it happens. That way we can see clearly in the logs if there is an issue.

Some ideas:
 - `Two plugins with order %v, %v still waiting to be launched.`
 - `Launching plugin %v with order %v` (just adding the order info to the existing log line)
 - Before the loop at all we could just log the entire ordered plugins for maximum clarity: `Got plugin ordering: %v`

johnSchnake (changes requested):
I'm testing this manually on my side too with some tweaks/nits/questions already made on the PR

pkg/plugin/aggregation/run_test.go
```diff
@@ -182,95 +183,194 @@ func TestRunAndMonitorPlugin(t *testing.T) {
...
+}
+
+func TestGetOrderedPlugins(t *testing.T) {
+	testCases := []struct {
+		desc     string
+		input    []plugin.Interface
+		expected [][]plugin.Interface
+	}{
+		{
+			desc: "Ensure lists of plugin.Interface are ordered by plugin order",
```
johnSchnake, line 248: nit: Very simple to add but be sure to include empty slice and null case, just to make sure we pin that behavior down and a change with that logic wont break something else.

pkg/plugin/aggregation/run.go
```diff
@@ -190,9 +192,27 @@ func Run(client kubernetes.Interface, plugins []plugin.Interface, cfg plugin.Agg
 		return errors.Wrapf(err, "couldn't get aggregator pod")
 	}
 
-	for _, p := range plugins {
-		logrus.WithField("plugin", p.GetName()).Info("Running plugin")
-		go aggr.RunAndMonitorPlugin(context.Background(), time.Duration(cfg.TimeoutSeconds)*time.Second, p, client, nodes.Items, cfg.AdvertiseAddress, certs[p.GetName()], aggregatorPod, progressPort, pluginResultsDir)
+	orderedPlugins := getOrderedPlugins(plugins)
+
+	fmt.Printf("Received plugin launch order:")
```
johnSchnake, line 197: Sorry if I wasn't clear. In this situation (on the aggregator) be sure to be consistent with the existing code and use `logrus.Info("...")` rather than `fmt.Print` statements.

Either extract this into its own method and change to `logrus...` or have that new method return the string you want logrus to log for you.

johnSchnake (changes requested):

johnSchnake:
I tried this out manually with a few cases and it worked great. Normal runs without order, worked. Edited one plugin to have order 99, launched one then the other. Modified both to have order 99, launched in parallel again.

We may tweak this logic over time, but I think having this basic `order` is great.

I already see us being able to use this and make query logic its own plugin and set its order to be high so that it gets run last. This will further simplify the core sonobuoy logic and maybe make it easier to customize too.

johnSchnake (approved):

Issue #1754, opened 2022-06-28
Resolve and/or document differences in status aggregation


johnSchnake:
When doing junit or gojson tests, we rollup results by looking at all leaf nodes and treating it as pass/fail/unknown with clear rules: 
 - fail + * = fail
 - unknown + [pass|unknown] = unknown

For manual results we defined another method to roll these up with the intent of supporting 'any' result value, not just pass/fail/unknown. However I think it was executed poorly. A few issues:
 - as noted in #1750 there was an issue and it had to do with a daemonset node reporting no results being ignored completely
 - even if they use manual results, we do not ever apply the same pass/fail rules above. It seems if we just see pass/fail/unknown we use the same logic

I think we need to get rid of manualResultsAggregation method completely and absorb the logic in to the other method. We need to avoid special casing DS entirely if possible. They should roll up in the same way unless we can come up with a reason otherwise.

This will allow us to delete some code, centralize logic/testing, and make things more understandable/consistent.

Issue #1750, opened 2022-06-27
Manual results missing but not reported as error


johnSchnake:
CI failure unexpected:

```
    sonobuoy_integration_test.go:602: Expected plugin to have status: manual-results-1: 3, manual-results-2: 3, got manual-results-1: 2, manual-results-2: 2
    sonobuoy_integration_test.go:84: Running "../../build/linux/amd64/sonobuoy delete -n sonobuoy-testmanualresultsdaemonset" with env: [SONOBUOY_ALL_FEATURES=true KUBECONFIG=/root/.kube/kubeconfig HOME=/tmp/sonobuoy_int_test_home_4043145929]
--- FAIL: TestManualResultsDaemonSet (20.37s)
```

So it didnt run on a node for some reason. Looking in the tarball the control plane node returned:
```
{"error":"open : no such file or directory"}
```

Nothing else of relevance seems to be in the logs. Thats one issue (the test plugin should maybe have more verbose logging to help diagnose this).

Secondly, the daemonset of manual plugins should report a failure if it doesnt successfully write the results as expected. In this case, the control plane didnt generate that file but the post-processor seemed to ignore that problem. Instead, if missing it should be reported as a failure.

This may be a simple change in the code (ensure the proper code path returns the failure).

I'm not certain if this bug happens all the time or only if its a daemonset with some nodes returning results and others not.

PR #1755, opened 2022-06-28
Simplify logic for status aggregation

johnSchnake:
- manual plugins now use the same code paths
 - daemonsets use the same code paths
 - if only known status values (pass/fail/etc) we use the same rules to
aggregate the values
 - in all other cases we roll them up using values like "custom: 1, foo: 2"

This will be easier to debug, document, and utilize and it fixes 2 separate issues.

Fixes https://github.com/vmware-tanzu/sonobuoy/issues/1750
Fixes https://github.com/vmware-tanzu/sonobuoy/issues/1754

### DIFF
--- a/pkg/client/results/processing.go
+++ b/pkg/client/results/processing.go
@@ -24,6 +24,7 @@ import (
 	"path"
 	"path/filepath"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/vmware-tanzu/sonobuoy/pkg/plugin"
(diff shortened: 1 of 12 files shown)


pkg/client/results/processing.go
```diff
@@ -52,84 +53,90 @@ type postProcessor func(string, string) (Item, error)
...
-	// but we report it as passed.
-	if len(items) == 0 {
-		return StatusUnknown
-	}
-
-	results := map[string]int{}
-	var keys []string
-
+// hasCustomValues returns true if there is a leaf node with a custom status value.
+func hasCustomValues(items ...Item) bool {
```
johnSchnake, line 57: FYI This method simplified the resulting logic. Without it you had to modify the Item structure (or create another and pass it around) to track whether or not custom values were seen. Otherwise two separate branches of the tree (think the first/second node) may differ. E.g. the LHS may only have pass/fail and therefore aggregate to only "pass" or "fail" (a single value) but the RHS would have custom ("pass" and "foo") and therefore count each (e.g. "foo: 1, pass: 2"). Then we've lost the count data for the LHS when we realize we need it.

johnSchnake:
@andrewyunt Please review

andrewyunt (approved):

PR #1780, opened 2022-07-28
Allow worker to upload directory instead of file

### LINKED ISSUE #1776: Allow worker to send a directory
I think the issue is that the done file being written is a directory and not a file. Typically what happens when you want the whole directory is that you tar up the directory and then write the tarfile name as the results.

I think this is a pretty simple enhancement to make though and support listing the directory itself. It would make things more simple for users too.

_Originally posted by @johnSchnake in https://github.com/vmware-tanzu/sonobuoy/issues/1774#issuecomment-1197084998_

johnSchnake:
Current flow requires the plugin to tar up the results then
point to them. It would simplify things if we allowed pointing
to a directory and we do the tar process ourselves.

Fixes #1776

### DIFF
--- a/pkg/worker/worker.go
+++ b/pkg/worker/worker.go
@@ -227,7 +227,20 @@ func allOtherContainersCompleted(myNS, myPod, myContainer string) bool {
 
 func handleWaitFile(resultFile, url string, client *http.Client) error {
 	var outfile *os.File
-	var err error
+
+	fileInfo, err := os.Stat(resultFile)
+	if err != nil {
+		return errors.Wrapf(err, "failed to stat result file %v", resultFile)
+	}
+
+	if fileInfo.IsDir() {
+		tarlocation := filepath.Join(resultFile, "results.tar.gz")
+		logrus.Tracef("Results to return to aggregator are a directory. Archiving and compressing %v into new result file %v", resultFile, tarlocation)
+		if err := tarball.DirToTarball(resultFile, tarlocation, true); err != nil {
+			return errors.Wrapf(err, "attempted to send a directory as results but the tar process failed")
+		}
+		resultFile = tarlocation
+	}
 
 	// Set content type
 	extension := filepath.Ext(resultFile)
--- a/test/integration/sonobuoy_integration_test.go
+++ b/test/integration/sonobuoy_integration_test.go
@@ -224,6 +224,35 @@ func TestNoDoneFile(t *testing.T) {
 	tb = saveToArtifacts(t, tb)
 }
 
+// TestDoneFileDirectory runs a plugin which reports the done file as the results directory itself and not the
+// files or a tarball to ensure that processing works and all results are considered.
+func TestDoneFileDirectory(t *testing.T) {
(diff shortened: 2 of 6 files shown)


pkg/worker/worker.go
```diff
@@ -227,7 +227,20 @@ func allOtherContainersCompleted(myNS, myPod, myContainer string) bool {
 
 func handleWaitFile(resultFile, url string, client *http.Client) error {
 	var outfile *os.File
-	var err error
+
+	fi, err := os.Stat(resultFile)
```
andrewyunt, line 231: nit: It's more readable to not abbreviate variable names in most cases. I think "file" is more readable than "fi"

test/integration/sonobuoy_integration_test.go
```diff
@@ -224,6 +224,32 @@ func TestNoDoneFile(t *testing.T) {
...
+	t.Parallel()
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
+	defer cancel()
+
+	ns, cleanup := getNamespace(t)
+	defer cleanup(false)
+
+	args := fmt.Sprintf("run --image-pull-policy IfNotPresent --wait -p testImage/yaml/job-manual-directory.yaml -n %v", ns)
+	mustRunSonobuoyCommandWithContext(ctx, t, ns, args)
+	tb := mustDownloadTarball(ctx, t, ns)
```
andrewyunt, line 239: Naming this variable "tarball" would also be more readable

test/integration/sonobuoy_integration_test.go
```diff
@@ -224,6 +224,32 @@ func TestNoDoneFile(t *testing.T) {
...
+	tb = saveToArtifacts(t, tb)
+
+	// Retrieve the sonobuoy results file from the tarball
+	resultsArgs := fmt.Sprintf("results %v --plugin %v --mode dump", tb, "job-manual-directory")
+	resultsYaml := mustRunSonobuoyCommandWithContext(ctx, t, ns, resultsArgs)
+	var resultItem results.Item
+	yaml.Unmarshal(resultsYaml.Bytes(), &resultItem)
```
andrewyunt, line 246: May be good to handle an error here - yaml.Unmarshal returns err

pkg/worker/worker.go
johnSchnake: In general I agree, but in this case `fi` is not an abbreviation for `file`, its an acronym for `FileInfo` which is the type and I wouldn't want to make it confusing since its actually not a `file` (e.g. you cant read it, it only carries some info about it).  Still reading through other comments but if I fix other stuff up I can rename for more clarity if you like.

test/integration/sonobuoy_integration_test.go
johnSchnake: Excellent catch, yes we definitely want to catch the error. If there was an error the result would be wrong too but for something like this we want to be explicit about it. 👍

johnSchnake:
Addressed review comments. Merging.

PR #1796, opened 2022-07-29
Allow users to skip specific preflight checks

andrewyunt:
Allow users to skip specific preflight checks using a new flag **skip-preflight-check**

The new flag was added since changing the flag type to string slice from bool requires an input value which would break any automation calling Sonobuoy from the CLI with --skip-preflight and no value.

### DIFF
--- a/cmd/sonobuoy/app/args.go
+++ b/cmd/sonobuoy/app/args.go
@@ -301,11 +301,12 @@ func AddRBACModeFlags(mode *RBACMode, flags *pflag.FlagSet, defaultMode RBACMode
 }
 
 // AddSkipPreflightFlag adds a boolean flag to skip preflight checks.
-func AddSkipPreflightFlag(flag *bool, flags *pflag.FlagSet) {
-	flags.BoolVar(
-		flag, "skip-preflight", false,
-		"If true, skip all checks before starting the sonobuoy run.",
+func AddSkipPreflightFlag(flag *[]string, flags *pflag.FlagSet) {
+	flags.StringSliceVar(
+		flag, "skip-preflight", []string{},
+		"Skips the specified preflight checks. Valid values are [dnscheck, versioncheck, existingnamespace] or true to skip all of the checks.",
 	)
+	flags.Lookup("skip-preflight").NoOptDefVal = "true"
 }
 
 // AddDeleteAllFlag adds a boolean flag for deleting everything (including E2E tests).
--- a/cmd/sonobuoy/app/run.go
+++ b/cmd/sonobuoy/app/run.go
@@ -18,10 +18,11 @@ package app
 
 import (
 	"fmt"
+	"os"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"os"
 
 	"github.com/vmware-tanzu/sonobuoy/pkg/client"
 	"github.com/vmware-tanzu/sonobuoy/pkg/errlog"
@@ -82,11 +83,12 @@ func submitSonobuoyRun(f *genFlags) func(cmd *cobra.Command, args []string) {
 			os.Exit(1)
 		}
 
-		if !f.skipPreflight {
+		if !contains(f.skipPreflight, "true") && !contains(f.skipPreflight, "*") {
 			pcfg := &client.PreflightConfig{
-				Namespace:    f.sonobuoyConfig.Namespace,
-				DNSNamespace: f.dnsNamespace,
-				DNSPodLabels: f.dnsPodLabels,
+				Namespace:           f.sonobuoyConfig.Namespace,
+				DNSNamespace:        f.dnsNamespace,
+				DNSPodLabels:        f.dnsPodLabels,
+				PreflightChecksSkip: f.skipPreflight,
 			}
 			if errs := sbc.PreflightChecks(pcfg); len(errs) > 0 {
 				errlog.LogError(errors.New("Preflight checks failed"))
--- a/pkg/client/interfaces.go
+++ b/pkg/client/interfaces.go
@@ -230,9 +230,10 @@ func (sc *StatusConfig) Validate() error {
 
 // PreflightConfig are the options passed to PreflightChecks.
 type PreflightConfig struct {
-	Namespace    string
-	DNSNamespace string
-	DNSPodLabels []string
+	Namespace           string
+	DNSNamespace        string
+	DNSPodLabels        []string
+	PreflightChecksSkip []string
 }
 
 // Validate checks the config to determine if it is valid.
--- a/pkg/client/preflight.go
+++ b/pkg/client/preflight.go
@@ -37,10 +37,10 @@ var (
 	minimumKubeVersion = version.Must(version.NewVersion(buildinfo.MinimumKubeVersion))
 	maximumKubeVersion = version.Must(version.NewVersion(buildinfo.MaximumKubeVersion))
 
-	preflightChecks = []func(kubernetes.Interface, *PreflightConfig) error{
-		preflightDNSCheck,
-		preflightVersionCheck,
-		preflightExistingNamespace,
+	validPreflightChecks = map[string]func(kubernetes.Interface, *PreflightConfig) error{
+		"dnscheck":          preflightDNSCheck,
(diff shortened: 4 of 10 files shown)


pkg/client/interfaces.go
```diff
@@ -230,9 +230,10 @@ func (sc *StatusConfig) Validate() error {
 
 // PreflightConfig are the options passed to PreflightChecks.
 type PreflightConfig struct {
-	Namespace    string
-	DNSNamespace string
-	DNSPodLabels []string
+	Namespace       string
+	DNSNamespace    string
+	DNSPodLabels    []string
+	PreflightChecks map[string]func(kubernetes.Interface, *PreflightConfig) error
```
johnSchnake, line 236: I think it may be easier to push the processing of the `skip` logic into `preflight.go`.

So instead of getting the map ahead of time, why not just put the list of preflight skips into this struct (rather than the map).

Then your caller only needs to pass along that string slice (easy) and then the sonobuoy client, when launching the checks can do the bulk of the logic.

I like the way you're doing it though; getting a copy of the valid func map, then deleting the values listed in the slice.

cmd/sonobuoy/app/run.go
```diff
@@ -83,10 +85,23 @@ func submitSonobuoyRun(f *genFlags) func(cmd *cobra.Command, args []string) {
...
 
 		if !f.skipPreflight {
+			preflightChecks := make(map[string]func(kubernetes.Interface, *client.PreflightConfig) error)
+
+			for key, value := range sbc.GetValidPreflightChecks() {
+				preflightChecks[key] = value
+			}
+
+			for _, name := range f.skipPreflightChecks {
+				if _, ok := sbc.GetValidPreflightChecks()[name]; ok {
```
johnSchnake, line 95: My other comment below talks about thinking this would be cleaner if you push this kind of logic down closer to where the preflight checks actually get launched, and in that situation you'll have the original `validPreflightChecks` funcMap to check against instead of calling to he method to get a reference to it.

Handle the `!ok` case as well so users are informed if they provided a nonsense value. You may want to normalize the input a bit too, just make everything `strings.toLower` to avoid tiny issues like capitalization from mattering/causing a user to error.

johnSchnake (changes requested):
I think this is great; left some comments but they really are pretty small to implement, mainly just moving code.

I'd prefer we try to put less info in the config object so the CLI client can be thinner. This makes the actual `pkg/client` object a bit easier to reuse if someone is using Sonobuoy as a library or we want to call into that method from a different location.

Otherwise I think the approach of using the string slice value is good for this.

Lets talk about the reason you added a new flag though; I think it may be solvable without doing that. `pflag` has the concept of a `NoOptDefValue` which is a bit weird but helps here. It would allow us to use the old value:

```
// Flag with no value, use NoOptDevValue=* (e.g. all)
--skip-preflight

// Flag with a value behaves like the string slice
--skip-preflight=foo
```

This would allow us to add the new functionality in a backwards compatible way without another flag.

cmd/sonobuoy/app/run.go
andrewyunt: Moved logic to preflight.go

johnSchnake:
There is some issue; the tests do seem to have legitimate failures:
```
    sonobuoy_integration_test.go:119: Expected "run --wait -f /tmp/gen.2299213011.yaml" to not error but got error: "exit status 1" with stdout: "" and stderr: "time=\"2022-08-02T23:05:02Z\" level=error msg=\"Preflight checks failed\"\ntime=\"2022-08-02T23:05:02Z\" level=error msg=\"invalid preflight check \"\n"
```

Continuing with review though.

cmd/sonobuoy/app/run.go
```diff
@@ -82,11 +83,12 @@ func submitSonobuoyRun(f *genFlags) func(cmd *cobra.Command, args []string) {
 			os.Exit(1)
 		}
 
-		if !f.skipPreflight {
+		if !contains(f.skipPreflight, "true") {
```
johnSchnake, line 86: I'd allow the `*` case as well here; will be more intuitive as we move towards a list of values rather than true/false.

cmd/sonobuoy/app/args.go
```diff
@@ -301,11 +301,12 @@ func AddRBACModeFlags(mode *RBACMode, flags *pflag.FlagSet, defaultMode RBACMode
 }
 
 // AddSkipPreflightFlag adds a boolean flag to skip preflight checks.
-func AddSkipPreflightFlag(flag *bool, flags *pflag.FlagSet) {
-	flags.BoolVar(
-		flag, "skip-preflight", false,
+func AddSkipPreflightFlag(flag *[]string, flags *pflag.FlagSet) {
+	flags.StringSliceVar(
+		flag, "skip-preflight", []string{""},
```
johnSchnake, line 306: This is currently the cause of the test failure. The default value here is a single value (the empty string).

What you're meaning to do here is use `[]string{}` (a slice of strings with no values at all).

Otherwise what was happening was that you're iterating over this list later and seeing the empty string given and erroring out.

pkg/client/preflight.go
```diff
@@ -62,6 +62,22 @@ func (c *SonobuoyClient) PreflightChecks(cfg *PreflightConfig) []error {
 		return []error{err}
 	}
 
+	preflightChecks := make(map[string]func(kubernetes.Interface, *PreflightConfig) error)
+	for key, value := range validPreflightChecks {
+		preflightChecks[key] = value
+	}
+
+	for _, name := range cfg.PreflightChecksSkip {
+		if strings.ToLower(name) == "false" {
```
johnSchnake, line 71: nit: You're using `strings.ToLower` 3 times here; I'd just create a new variable `lName` for the lower name to use in this narrow context.

I'd also say that if they provide `false` we should just run all the preflight checks. Its a weird situation since a user shouldn't ever do that `--skip-preflight=foo --skip-preflight=false,bar`. They're saying to skip foo and bar but not skip any?

Regardless, I'd rather that if they said `false` at all, we just allow all valid preflight checks.

You can add a `logrus.Trace(...)` statement there to expose that logic in the logging if you want. Its a good idea to surface little branches in logic like that.

johnSchnake (changes requested):

pkg/client/preflight.go
andrewyunt: Added a loop that lowercases all preflight check names, and checks to see if a "false" check was passed in the slice which will cause all checks to execute.

cmd/sonobuoy/app/args.go
```diff
@@ -301,11 +301,12 @@ func AddRBACModeFlags(mode *RBACMode, flags *pflag.FlagSet, defaultMode RBACMode
 }
 
 // AddSkipPreflightFlag adds a boolean flag to skip preflight checks.
-func AddSkipPreflightFlag(flag *bool, flags *pflag.FlagSet) {
-	flags.BoolVar(
-		flag, "skip-preflight", false,
+func AddSkipPreflightFlag(flag *[]string, flags *pflag.FlagSet) {
+	flags.StringSliceVar(
+		flag, "skip-preflight", []string{},
 		"If true, skip all checks before starting the sonobuoy run.",
```
johnSchnake, line 307: Lets update this value to reflect the new functionality:
```
Skips the specified preflight checks. Valid values are [...] or true to skip all of the checks.
```

cmd/sonobuoy/app/run.go
```diff
@@ -82,11 +83,12 @@ func submitSonobuoyRun(f *genFlags) func(cmd *cobra.Command, args []string) {
 			os.Exit(1)
 		}
 
-		if !f.skipPreflight {
+		if !contains(f.skipPreflight, "true") && !contains(f.skipPreflight, "*") {
```
johnSchnake, line 86: I'm fine leaving `*` as a special case just FYI. In my recommended docs I didnt list it but if you think it reads ok to add another special case, feel free. `...true or * will skip all...`. In typical shells you have to escape the `*` so it can be a bit less intuitive.

johnSchnake (changes requested):
Looks good
 - tweak the help text
 - regen the golden files (`make golden`)
 - Remove `kubectl` from the diff; your binary got added by accident.
Then good to merge.

johnSchnake (approved):

PR #1805, opened 2022-08-02
Add K8s release check workflow

### LINKED ISSUE #1793: Can we automate issues to be created on k8s releases?
We currently have a bad system for updating the e2e data; we simply check periodically (manually) if there are new releases or not.

We should automate this better so that, like dependabot, when k8s pushes new releases we create issues here to update our data.

johnSchnake:
Alerts us to k8s releases in order to avoid long delays in
updating the e2e functionality.

Fixes #1793

### DIFF
--- a/.github/workflows/k8sRelease.yaml
+++ b/.github/workflows/k8sRelease.yaml
@@ -1,0 +1,23 @@
+name: K8s Release Check
+on:
+  schedule:
+  # Every day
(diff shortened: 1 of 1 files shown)


johnSchnake:
Low risk workflow separate from everything else. Merging and will adjust if it doesn't work as expected/intended.

PR #1958, opened 2024-02-05
driver/daemonset: Align node selection behavior with Kubernetes scheduler

### LINKED ISSUE #1957: Node selector / Node affinity behavior is inconsistent with Kubernetes Scheduler

1. Set up a node A and set two labels to A: `a: b` and `c: d`
2. Set up another node B and set one label to B: `a: b`
3. Set up a sonobuoy plugin with `DaemonSet` driver
  - We want to avoid the plugin running on the node A
4. Run a sonobuoy daemonset plugin with either PodSpec configuration below:
  - Case X: Put a node selector to run Pods only on node A and B and a node affinity to avoid running Pods on node A
  - Case Y: Put a node affinity that contains two match expressions in a node SelectorTerms: the one forces Pods to run only on node A and B and the other one prevents a Pod running on node A

Case X

```yaml
podSpec:
  nodeSelector:
    a: b
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: c
            operator: NotIn
            values:
            - d
```

Case Y

```yaml
podSpec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: a
            operator: In
            values:
            - b
          - key: c
            operator: NotIn
            values:
            - d
```

5. Run a sonobuoy daemonset plugin with above plugin congiguration
6. Sonobuoy counts the both node A and B as the available nodes

On our environment, node A is the [Fargate node](https://docs.aws.amazon.com/eks/latest/userguide/fargate.html) and node B is the normal node. DaemonSets cannot run on Fargate nodes, resulting the plugin always fails with `No pod aws scheduleded on node A` error.

Sonobuoy should not count the node A as an available node.

This issue is caused by inconsistency in handling `nodeSelector` and `nodeAffinity` between Kubernetes Scheduler and Sonobuoy.

Case X: Kubernetes schedules a Pod on a node that satisfies both `nodeSelector` and `nodeAffinity`.

> If you specify both nodeSelector and nodeAffinity, both must be satisfied for the Pod to be scheduled onto a node.
>
> https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/

Currently Sonobuoy performs like `OR` of `nodeSelector` and `nodeAffinity` [code](https://github.com/vmware-tanzu/sonobuoy/blob/6f9e27f1795f10475c9f6f5decdff692e1e228da/pkg/plugin/driver/daemonset/daemonset.go#L131-L144).

Case Y: In Kubernetes, match expressions in a single `matchExpressions` field are ANDed.

> If you specify multiple expressions in a single matchExpressions field associated with a term in nodeSelectorTerms, then the Pod can be scheduled onto a node only if all the expressions are satisfied (expressions are ANDed).
>
> https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/

Currently Sonobuoy counts a node as available if at least one expression is matched [code](https://github.com/vmware-tanzu/sonobuoy/blob/6f9e27f1795f10475c9f6f5decdff692e1e228da/pkg/plugin/driver/daemonset/daemonset.go#L433-L458).

**Environment:**

- Sonobuoy version: 0.57.1 (go1.21.4)
- Kubernetes version: (use `kubectl version`): Confirmed with multiple versions, 1.25~1.27
- Kubernetes installer & version: AWS EKS
- Cloud provider or hardware configuration: AWS EKS
- OS (e.g. from `/etc/os-release`): n/a
- Sonobuoy tarball (which contains * below): Please request if needed

nonylene:

Align node selection behavior with Kubernetes scheduler in DaemonSet driver.

- Fixes #1957 

**Release note**:
```
Align node selection behavior with Kubernetes scheduler in DaemonSet driver.
```


pkg/plugin/driver/daemonset/daemonset.go
```diff
@@ -119,28 +119,30 @@ func (p *Plugin) filterByNodeSelector(nodes []v1.Node) []v1.Node {
 	}
 	ls = ls.Add(reqs...)
 
-	if ls.Empty() && nodeSelector == nil {
```
nonylene, line 122: This condition is now covered within the for loop below.

pkg/plugin/driver/daemonset/daemonset_test.go
```diff
@@ -608,60 +616,89 @@ func TestExpectedResults(t *testing.T) {
...
-			desc: "Can combine filters as union",
+			desc: "Affinity: Can combine selctor terms as union",
 			expect: []plugin.ExpectedResult{
 				{NodeName: "node1", ResultType: "myPlugin"},
 				{NodeName: "node2", ResultType: "myPlugin"},
 				{NodeName: "node4", ResultType: "myPlugin"},
 			},
-			p: pluginWithAffinity([]corev1.NodeSelectorRequirement{
-				{Key: "foo", Operator: corev1.NodeSelectorOpNotIn, Values: []string{"bar", "baz"}},
-				{Key: "foo", Operator: corev1.NodeSelectorOpIn, Values: []string{"bar"}},
```
nonylene, line 657: This test case was inconsistent with the behavior of Kubernetes scheduler (fixed)

franknstyle (approved):

